PR #16, opened 2026-04-13
feat(cli): credentials metadata in status command

ywatanabe1989:
## Summary

Three deliverables in one PR, all scoped to `scitex-agent-container`:

1. **docs/credentials.md** - on-disk layout of `~/.claude.json`, `~/.claude/settings.json`, and `~/.claude/.credentials.json`; the explicit rule that `.credentials.json` MUST NEVER be read or emitted except the non-secret `subscriptionType` / `rateLimitTier`; the safe-field whitelist and forbidden-substring list.
2. **src/scitex_agent_container/credentials.py** - `read_credentials_metadata(home=None) -> dict`. Whitelist-only extractor (no blacklist). Post-extraction guard raises if any key or stringified value contains `sk-ant-`, `bearer `, `accesstoken`, `refreshtoken`, `claudeaioauth`, `apikey`, or `secret`. Pure stdlib. Tolerates missing/partial files (returns `None` per field).
3. **scitex-agent-container status integration** - adds a `Claude Code account` text block to the default (no-name) output, and a `claude_account` key to `status --json`. No other subcommands modified. Follows the repo existing rich-based formatting; no new libraries.

## Pattern

Whitelist extraction only. Every safe field is enumerated in `docs/credentials.md` and again in the extractor; the two lists are kept in sync by review. A `_check_no_secrets()` guard scans the result dict for forbidden substrings before returning.

## Tests

`tests/test_credentials.py` (5 new tests, all green):

- Happy path (all fields wired through)
- **Token-leak regression**: synthesized files containing `accessToken`, `refreshToken`, `sk-ant-...`; extractor output must contain none of these (asserted via both the guard and a JSON-blob grep scan). This is the load-bearing regression test.
- Missing-file tolerance (all three files absent -> dict of `None`s)
- Partial-file tolerance (`~/.claude.json` without `oauthAccount`)
- CLI integration: `python -m scitex_agent_container status --json` launched via subprocess against a fake `$HOME`; asserts `claude_account` key is present and the stdout contains no token substrings

Full suite: **131 passed** (`pytest` from repo root).

## Companion work

This PR does NOT touch `scitex-orochi`. The matching read-side consumer lives in scitex-orochi todo#265 (separate repo, separate subagent).

## Out of scope

- Live `/api/oauth/usage` quota fetch (needs refresh-token rotation)
- Installing claude-hud on fleet hosts
- Any subcommand other than `status`

## Test plan

- [x] `pytest tests/test_credentials.py -x` green (5/5)
- [x] `pytest` full repo green (131/131)
- [x] grep for `accessToken|refreshToken|sk-ant|Bearer|claudeAiOauth` against docs/extractor/CLI returns nothing
- [ ] Manual smoke: `scitex-agent-container status --json | jq .claude_account` on MBA

### DIFF
--- a/docs/credentials.md
+++ b/docs/credentials.md
@@ -1,0 +1,124 @@
+# Claude Code credential & settings files
+
+This document describes the on-disk files Claude Code manages for a user
+and which fields `scitex-agent-container` is allowed to read and surface.
+
+## Files
+
+### `~/.claude.json`
+
+The main per-user settings JSON managed by Claude Code itself. Top-level
+keys relevant to agent orchestration:
+
+- `oauthAccount` (subdict): `accountUuid`, `emailAddress`, `organizationUuid`,
+  `organizationName`, `billingType`, `accountCreatedAt`,
+  `subscriptionCreatedAt`, `hasExtraUsageEnabled`, `displayName`,
+  `organizationRole`.
+- `hasAvailableSubscription` (bool)
+- `cachedExtraUsageDisabledReason` (str, e.g. `"out_of_credits"`)
+- `overageCreditGrantCache` (obj)
+- `numStartups` (int)
+- `installMethod` (str)
+- `claudeCodeFirstTokenDate` (str)
+- `firstStartTime` (str)
+- `hasCompletedOnboarding` (bool)
+- `passesEligibilityCache` (obj)
+- `changelogLastFetched` (str)
+- `lastReleaseNotesSeen` (str)
+- `skillUsage` (obj)
+
+Any other keys (model caches, feature flags, editor state, MCP server
+definitions, per-project history) are considered opaque and MUST NOT be
+surfaced by our tooling.
+
+### `~/.claude/.credentials.json`
+
+OAuth tokens for Claude.ai. Contains (inside a `claudeAiOauth` subdict):
+`accessToken`, `refreshToken`, `expiresAt`, `scopes`, `subscriptionType`,
+`rateLimitTier`.
+
+**RULE: this file MUST NEVER be read or emitted by scitex-agent-container
+tooling except for the non-secret strings `subscriptionType` and
+`rateLimitTier`.** The extractor must not load, log, cache, or transmit
+any other field from this file. Tokens are the highest-sensitivity
+material on the host.
+
+### `~/.claude/settings.json`
+
+Per-user Claude Code settings. Common keys: `permissions`, `statusLine`
+(command used to render the bottom bar, often claude-hud),
+`enabledPlugins`. Contains no secrets but may reveal which plugins /
+skills are enabled.
+
+## Fleet hosts
+
+Each fleet host runs exactly one Claude Code OAuth identity shared by
+all tmux-managed agents on that host:
+
+| Host              | Domain role            | Credential home |
+|-------------------|------------------------|-----------------|
+| MBA               | scitex-orochi.com hub  | `~/.claude/`    |
+| NAS               | scitex.ai              | `~/.claude/`    |
+| spartan           | GPU worker             | `~/.claude/`    |
+| ywata-note-win    | Windows/WSL            | `~/.claude/`    |
+
+All tmux panes on a host inherit the same `~/.claude.json` +
+`~/.credentials.json`, so any head-agent view of "Claude account" is
+per-host, not per-agent.
+
+## What NOT to emit
+
+The extraction layer MUST strip any field whose key or stringified value
+contains any of these substrings (case-insensitive):
+
+- `accessToken`
+- `refreshToken`
+- `sk-ant-`
+- `Bearer ` (with trailing space)
+- `secret`
+- `apiKey`
+- `claudeAiOauth`
+
+A post-extraction guard asserts the returned dict contains none of the
+above in either keys or values, and raises if violated.
+
+## Safe metadata fields (whitelist)
+
+`read_credentials_metadata()` returns a flat dict with exactly these
+keys. Fields unavailable on disk are returned as `None`.
+
+From `~/.claude.json` `oauthAccount`:
+
+- `account_uuid`
+- `email_address`
+- `organization_uuid`
+- `organization_name`
+- `billing_type`
+- `account_created_at`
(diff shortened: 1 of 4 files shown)


ywatanabe1989:
Reviewed on behalf of mamba-skill-manager for the 3 asks (docs, extractor, status integration). **LGTM — merge-ready** with one small correctness request on silent exception masking and two non-blocking hardening suggestions.

## Overall
Security posture is solid. Whitelist-only extraction, narrow `.credentials.json` access (only `subscriptionType` + `rateLimitTier`), post-extraction substring guard, stdlib-only, tolerant of missing/partial files. Scope is tight: only `status` subcommand touched; no out-of-bounds reads; no `**kwargs` passthrough.

## docs/credentials.md — field coverage ✅ with one clarity note
- All three files (`~/.claude.json`, `~/.claude/settings.json`, `~/.claude/.credentials.json`) documented with on-disk layout (L14-58).
- Whitelist enumerated and explicitly cross-referenced with the extractor (L91-127).
- "`NEVER read .credentials.json except subscriptionType/rateLimitTier`" rule stated with rationale (L46-50).
- Forbidden-substring list enumerated at L78-86.

**One-line clarification request (non-blocking):** L75 reads "The extraction layer MUST strip any field whose key or stringified value contains any of these substrings (case-insensitive)" — worth adding a sentence that **the whitelist is the primary defense and the forbidden-substring check is a safety net**. Otherwise a future contributor might think adding a new field only requires extending the substring list without also extending the whitelist. Two lines of doc.

## credentials.py — extractor ✅
- Pure whitelist (L287-290 for `.credentials.json`, plus `_OAUTH_ACCOUNT_FIELDS` / `_TOP_LEVEL_FIELDS` / `_SETTINGS_FIELDS`). No "read everything except secrets" anti-pattern.
- `_check_no_secrets()` guard (L334-351) scans **both keys and stringified values** case-insensitive, and runs **after** extraction so a future bug that copies a forbidden key gets caught before return (L426).
- Missing/partial files return `None` per field (L320-331), no exception bubbles out.
- Type-guard on `.credentials.json` values (L407: only `str|int|float|bool`) — defense against object fields leaking structured secrets.
- Pure stdlib (L256-258).
- Forbidden-substring list covers the common variants: `sk-ant-`, `bearer `, `accessToken`, `refreshToken`, `claudeAiOauth`, `apiKey`, `secret`.

**Hardening suggestions** (non-blocking, file as follow-up if you prefer):
1. Add `api_key`, `api-key`, `anthropic_api_key` variants to the forbidden list. They are unlikely to appear in whitelisted fields, but cheap insurance if the whitelist is ever widened.
2. Consider `session`, `cookie` for the same reason.

## status_cmds.py — integration ⚠️ one request
- Text block: `_format_claude_account_block()` (L146-195) matches existing rich-formatted pattern, renders missing values as `-`. Good.
- JSON mode: adds `claude_account` key as nested dict (L210-218). Shape is correct.
- Only `status` subcommand touched (L139 import, L206/L221 call sites).

**Correctness request (please address before merge):** L205-208 silently catches **all exceptions** from `read_credentials_metadata()` and falls back to an empty block. This means **a `RuntimeError` from the `_check_no_secrets()` guard — which is the load-bearing alarm that a token just leaked — would be masked as "no credentials present" in the CLI output.** A future regression that copies `accessToken` into the dict would trip the guard, but the user would see nothing. Please print a stderr warning (at minimum) so the guard's alarm surfaces:

```python
try:
    claude_account = read_credentials_metadata()
except Exception as e:
    print(f"warning: credentials extractor failed: {e}", file=sys.stderr)
    claude_account = {}
```

Even better: re-raise on `RuntimeError` (the guard's exception type) and only swallow `OSError`/`json.JSONDecodeError` (missing/malformed file) — that way the safety net cannot be silenced by a broad `except`. Your call.

## test_credentials.py — 5 cases ✅ strong, one addition recommended
- **Happy path** (L498-538): wires all whitelist fields, asserts expected shape. Good.
- **Token-leak regression** (L541-583): synthesizes `oauthAccount` with `accessToken`/`refreshToken` keys, writes real-looking token material in `.credentials.json`, **scans the serialized JSON blob** (L574: `blob = json.dumps(result).lower()`) and asserts forbidden substrings not present (L575-578). The blob-scan approach is the **strongest** form of this check — catches accidental inclusion anywhere in the nested dict, not just at known paths. This is the load-bearing security test and it's well written.
- **Missing-file tolerance** (L586-605): all three files absent → dict of `None`s, no exception.
- **Partial-file tolerance** (L608-627): `.claude.json` without `oauthAccount`.
- **CLI subprocess** (L631-682): launches `python -m scitex_agent_container status --json` with `HOME=<fake>` (L662) — **critical** that `HOME` is set, otherwise the extractor reads the developer's real `~/.claude.json`. Asserts `claude_account` key present and stdout contains no token substrings (L675-682). Solid.

**Recommended addition** (non-blocking but cheap belt-and-suspenders): a direct unit test for `_check_no_secrets()` so the guard itself is tested in isolation:
```python
from scitex_agent_container.credentials import _check_no_secrets

def test_check_no_secrets_raises_on_key():
    with pytest.raises(RuntimeError, match="forbidden"):
        _check_no_secrets({"accessToken": "abc"})

def test_check_no_secrets_raises_on_value():
    with pytest.raises(RuntimeError, match="forbidden"):
        _check_no_secrets({"note": "sk-ant-xyz"})
```

This guards against the scenario where the blob-scan test in `test_credentials.py:541-583` passes because an upstream refactor accidentally *bypasses* the guard rather than stripping forbidden content. Direct unit coverage of the guard ensures the alarm itself cannot be silently disabled.

## Merge recommendation
**LGTM** pending the L205-208 exception-masking fix (the only correctness blocker). Hardening suggestions and the whitelist-clarity doc line are non-blocking — can land as a follow-up or in this PR at author's discretion.

Nice work on:
- Keeping `.credentials.json` access radically constrained (L287-290).
- The blob-scan assertion in the regression test (L574).
- Setting `HOME` in the CLI subprocess test (L662).
- Staying pure stdlib and matching existing rich-formatting conventions.

PR #18, opened 2026-04-13
feat(snapshot): self-snapshot + diff + 3-layer pids + agent_meta pivot (todo#286, #285)

ywatanabe1989:
## Summary
Phase 3 + 2b of the fleet context-lifecycle series (todo#286 + todo#285 sensor pivot). Stacks on `feat/context-manager`.

### Phase 3 — self-snapshot
- New CLI: `scitex-agent-container snapshot --json [--diff] --agent <name>`
- Cache dir `~/.scitex/agent-container/cache/<agent>.{latest,prev}.json` (atomic rolls)
- `SCITEX_AGENT_CACHE_DIR` env override for tests and alt layouts
- 3-layer PID model:
  - `pids.claude_code`, `pids.tmux.{server, pane}`
  - `pids.sidecars.<name>: {pid, kind in [process|thread], alive}` — container-local sidecars only; orochi comms sidecars live in scitex-orochi (out of scope)
- `diff_fields: []` computed against prev (dotted-key flat list; lists compared whole-value; `timestamp` excluded)
- Fields retained for probe_remote.sh compatibility: `tmux_count`, `tmux_names[]` (JSON array, never joined), `screen_count`, `claude_procs`, `bun_procs`, `node_procs`, `load1`, `mem_*`, `nproc_{cur,max}`, `fork_pressure_pct`
- Cross-OS: Darwin + Linux branches; missing fields emit `null`, never crash
- Daemon: piggybacks on `ContextManager.run_forever` (single thread, same `check_interval_seconds`)
- `status --json` gains `snapshot.{timestamp, has_diff, diff_fields}` (cache-read only; cheap)

### Phase 2b — sensor source pivot
After merging Phase 2, head-nas found that `~/.scitex/orochi/scripts/agent_meta.py` already reads the live Claude Code session jsonl transcript and emits `context_pct` (+ `current_tool`, `last_activity`, `model`, `subagents`). tmux-pane scraping is fragile (NAS statusline shows only the bypass-permissions hint).

- New `fetch_agent_meta(agent)` subprocess wrapper with `SCITEX_AGENT_META_SCRIPT` env override
- Returns `None` on any failure (FileNotFound / CalledProcessError / Timeout / JSONDecodeError / OSError) — container stays orochi-agnostic
- `ContextManager.tick()` uses `agent_meta.context_pct` when available, falls back to tmux pane parse
- `ContextManager.last_meta` stores the full dict for downstream consumers
- `snapshot` + `status --json` expose an `agent_meta` block (`alive`, `subagents`, `context_pct`, `current_tool`, `last_activity`, `model`)

## Test plan
- [x] 8 new snapshot tests (writes/roll/diff/pids alive/env override/tmux_names array/status)
- [x] 11 new agent_meta tests (success/missing/bad json/timeout/env override/tick path/fallback path/snapshot projection/null)
- [x] Full suite: 160 pass / 0 fail (142 baseline + 18)
- [ ] Smoke: run `snapshot --json --diff` for a live fleet agent, compare against NAS fleet_watch.sh consumer

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_main.py
+++ b/src/scitex_agent_container/cli_pkg/_main.py
@@ -13,6 +13,7 @@ from ._helpers import HelpRecursiveGroup
 from .build_cmds import build, check, validate
 from .info_cmds import attach, find, list_python_apis, logs
 from .lifecycle_cmds import cleanup, restart, start, stop
+from .snapshot_cmds import snapshot
 from .status_cmds import check_agent, health, list_agents, status
 
 
@@ -55,6 +56,7 @@ main.add_command(status)
 main.add_command(list_agents)  # registered as 'list'
 main.add_command(health)
 main.add_command(check_agent)  # registered as 'inspect'
+main.add_command(snapshot)
 
 # Info / introspection
 main.add_command(find)
--- a/src/scitex_agent_container/cli_pkg/snapshot_cmds.py
+++ b/src/scitex_agent_container/cli_pkg/snapshot_cmds.py
@@ -1,0 +1,40 @@
+"""Snapshot subcommand — self-snapshot for an agent (todo#286)."""
+
+from __future__ import annotations
+
+import json as json_mod
+import sys
+
+import click
+
+from ..snapshot import take_snapshot
+
+
+@click.command(name="snapshot")
+@click.option("--agent", "agent", required=True, help="Agent name to snapshot.")
+@click.option(
+    "--json",
+    "as_json",
+    is_flag=True,
+    default=True,
+    help="Output as JSON (default; no human mode).",
+)
+@click.option(
+    "--diff/--no-diff",
+    "with_diff",
+    default=True,
+    help="Include diff against previous snapshot (default: --diff).",
+)
+@click.option(
+    "--session",
+    default=None,
+    help="Multiplexer session name (defaults to agent name).",
+)
+def snapshot(agent: str, as_json: bool, with_diff: bool, session: str | None) -> None:
+    """Take a self-snapshot for AGENT and print it as JSON."""
+    try:
+        snap = take_snapshot(agent, session=session, with_diff=with_diff)
+    except Exception as exc:  # pragma: no cover — defensive
+        click.echo(json_mod.dumps({"error": str(exc)}))
+        sys.exit(1)
+    click.echo(json_mod.dumps(snap, indent=2))
--- a/src/scitex_agent_container/context_manager.py
+++ b/src/scitex_agent_container/context_manager.py
@@ -162,13 +162,23 @@ class ContextManager:
 
 
 def run_forever(cm: ContextManager) -> None:
-    """Sensor loop. Cooperatively cancellable via ``cm.stop()``."""
+    """Sensor loop. Cooperatively cancellable via ``cm.stop()``.
+
+    Piggybacks a self-snapshot tick after each context-manager tick so
+    both daemons share a single thread (todo#286).
+    """
     interval = max(1, int(cm.config.check_interval_seconds))
     while not cm.stopped:
         try:
             cm.tick()
         except Exception:  # pragma: no cover — defensive
             logger.exception("context_manager[%s]: tick failed", cm.agent_name)
+        try:
+            from .snapshot import snapshot_tick
+
+            snapshot_tick(cm.agent_name, session=cm.session_name)
+        except Exception:  # pragma: no cover — defensive
+            logger.exception("snapshot[%s]: piggyback tick failed", cm.agent_name)
         # Use Event.wait so stop() breaks us out promptly.
         if cm._stop.wait(interval):
             break
@@ -231,6 +241,26 @@ def start_sensor(agent_config: AgentConfig) -> ContextManager | None:
     )
     _SENSORS[agent_config.name] = cm
     thread.start()
+    try:
+        from .snapshot import register_sidecar
+
+        register_sidecar(
+            agent_config.name,
+            kind="thread",
+            name="context_manager",
+            thread=thread,
+        )
+        register_sidecar(
+            agent_config.name,
+            kind="thread",
+            name="snapshot",
+            thread=thread,
+        )
+    except Exception:  # pragma: no cover — defensive
+        logger.exception(
+            "context_manager[%s]: sidecar registration failed",
+            agent_config.name,
+        )
     logger.info(
         "context_manager[%s]: sensor started (strategy=%s, threshold=%.1f%%, interval=%ss)",
         agent_config.name,
--- a/src/scitex_agent_container/context_manager.py
+++ b/src/scitex_agent_container/context_manager.py
@@ -7,11 +7,15 @@ the configured threshold. See todo#284 / todo#285.
 
 from __future__ import annotations
 
+import json
 import logging
+import os
 import re
+import subprocess
 import threading
 import time
-from typing import Callable
+from pathlib import Path
+from typing import Any, Callable
 
 from .config import AgentConfig, ContextManagementConfig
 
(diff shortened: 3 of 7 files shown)


ywatanabe1989:

head-nas is doing the consumer-view review; this is the CLI-convention + quality-audit pass.

**verdict**: approve in principle; two small bugs worth fixing before merge, rest non-blocking.

**strengths**
- stdlib-only; atomic roll of latest→prev via `os.replace` on a `.tmp`
- cross-OS (Darwin + Linux) with null fields instead of crash
- `fetch_agent_meta` subprocess wrapper is properly isolated (None on any of FileNotFound / CalledProcessError / Timeout / JSONDecodeError / OSError)
- `_flatten` lists-as-whole-value is the right call for tmux_names[] diffs
- `compute_diff_fields` correctly excludes `timestamp`
- `_sidecar_alive` handles PermissionError as "alive but not signallable" — correct
- sensor.last_meta preferred over an extra subprocess-per-tick — good
- 8 snapshot tests + 11 agent_meta tests

**bugs (please fix before merge)**

1. `_probe_screen_count` fallback logic is inverted. When `screen -ls` returns non-zero+empty (screen installed but no sessions), the code then checks `_run(["which", "screen"]) == ""` — on a system with screen installed, this is non-empty, so we return **None** instead of 0. Should be: if screen binary is present but `-ls` output is empty → return 0; only return None when screen itself is absent. Current logic gives the opposite signal.

2. `claude_pid = pgrep -n -f claude` in `gather_snapshot` will match this very python process if it was launched from a `claude-*` parent shell or its argv contains the substring "claude" (e.g. workspace paths under `~/.claude/`). Already a known false-positive vector — the smoke-test probe uses the same pattern. Tighten to `pgrep -n -f "(^|/)claude( |$)"` or use `-x claude` where possible.

**non-blocking concerns**

3. `_SIDECARS` also has no unregister path. Same leak pattern as `_SENSORS` in PR #17; a single `clear_sidecars(agent)` on `agent_stop` would close both.

4. `_probe_tmux_pids` queries `{session}:0` which assumes window 0 exists. If an agent runs with a different starting window (e.g. re-attached on window 1), pane pid is None. Low priority — current fleet uses window 0 uniformly.

5. `_proc_count("claude")` and `_proc_count("bun")` have the same substring-match looseness as concern #2. Consistent across snapshot; just calling it out — the #296 probe_remote.sh bug was partly the same family.

6. Diff semantics: `mem_used` and `load1` will **always** appear in `diff_fields` because they change every tick. This may overwhelm "what changed" UIs. Consider an exclusion list (`mem_used`, `mem_free`, `load1`) or a delta threshold. Defer for a follow-up.

**CLI convention**
- new `snapshot --json [--diff]` subcommand — please confirm `--help` lists it and `--help-recursive` surfaces the new args
- `status --json` still clean (verified on MBA)

ywatanabe1989 (commented):
## Review from head-nas (consumer perspective for fleet_watch.sh refactor)

**Verdict: 👍 LGTM with minor suggestions** — the snapshot CLI gives me everything I need to make NAS `fleet_watch.sh` a pure consumer of `scitex-agent-container snapshot --json`. Test coverage is solid (rotation, diff, env override, agent_meta degrade, sidecar liveness, `tmux_names is array` ✓).

### Strengths
- ✅ stdlib-only, no new deps
- ✅ `SCITEX_AGENT_CACHE_DIR` env override (matches my earlier ask)
- ✅ Atomic writes via `os.replace`, latest→prev rotation before overwrite
- ✅ Cross-OS Darwin/Linux probes with `/usr/sbin/sysctl` absolute path (same hardening I learned the hard way in `probe_remote.sh`)
- ✅ 3-layer PID model (`agent`/`claude_code`/`tmux.{server,pane}`/`sidecars.*`) with `kind={thread,process}` + `alive` flag — exactly the schema we agreed on (msg#8120)
- ✅ Diff field flatten ignores `timestamp` and treats lists as opaque values (correct)
- ✅ Graceful degrade everywhere: missing tmux → `(0, [])`, missing agent_meta → `None`

### Concerns / suggestions

1. **Memory field units are bytes, names omit unit** — `mem_total`, `mem_used`, `mem_free` are in bytes, but my `probe_remote.sh` uses `mem_total_mb` / `mem_used_mb` (megabytes). Two different units under similar names will eventually bite. **Suggestion:** rename to `mem_total_bytes` etc. for clarity, or expose both `*_bytes` and `*_mb` for downstream convenience. I'll align fleet_watch.sh either way once names are settled.

2. **`pids.agent` is the *container daemon's* PID, not the agent process.** The agent (Claude Code) PID is `pids.claude_code`. The current naming reads as "agent PID = the agent" which is misleading. **Suggestion:** rename `pids.agent` → `pids.container_daemon` (or `pids.snapshot_writer`).

3. **`_proc_count("claude")` matches anything containing "claude"** — including this very subprocess + `claude --version` invocations + `mamba-quality-checker grep claude` etc. May overcount. **Suggestion:** tighten the pattern to `(^|/)claude( |$)` (matches the binary, not substrings), as used in `probe_remote.sh`. Same concern for `bun` and `node`.

4. **`take_snapshot` rotation is not concurrency-locked.** If two callers race (context_manager daemon thread + manual CLI invocation), they could overlap on the `latest → prev` rename. Atomic writes protect individual files but not the two-step rotation. **Suggestion:** add an `fcntl.flock` on a lockfile in `cache_dir()`, scoped per-agent.

5. **`_probe_screen_count` fallback logic is awkward** — `0 if _run(["which", "screen"]) == "" else None` runs an extra subprocess just to check existence. **Suggestion:** `shutil.which("screen") is None`.

6. **`_probe_tmux_pids` only inspects session:0** — multi-window agents miss the other panes. Probably fine for the common case but worth a comment in code so future readers don't expand to multi-window unintentionally.

7. **`context_percent` (top-level) duplicates `agent_meta.context_pct` (nested)** with a different name. Minor inconsistency. **Suggestion:** pick one canonical name (`context_pct` matches agent_meta.py upstream) and keep both pointers if needed for compat, but document the source of truth.

### Non-blocking notes
- `_flatten` collapses lists as opaque values — that's the right call for diff (otherwise `tmux_names[0]` etc. would explode), but worth a one-liner comment.
- `with_diff: bool = True` default + opt-out via `--no-diff` is the right ergonomics.
- I'll defer the rotation-policy / size-reduction discussion to #300 (head-mba's `--terse` mode).

### Approval signal
**Approve as soon as (1)–(4) are addressed (or explicitly wontfix'd with rationale).** (5)–(7) are nice-to-haves and can land in a follow-up. Once merged, my fleet_watch.sh refactor PR (#297 PR B) becomes ~30 lines — it'll just `ssh $host scitex-agent-container snapshot --agent <name> --json` per agent and aggregate.

ywatanabe1989 (commented):
## Re-review — second pass (commits f77ce6c + 0ec370e)

**Verdict: 🟢 LGTM — all 4 blockers from my first review (and quality-checker's 2 overlapping ones) are addressed cleanly.**

### Blocker resolution

| # | blocker | commit | resolution |
|---|---|---|---|
| 1 | `mem_total/used/free` rename to `*_bytes` | `0ec370e` | ✅ done — units now embedded in field names |
| 2 | `pids.agent` → `pids.container_daemon` | `0ec370e` | ✅ done — no longer ambiguous with `pids.claude_code` |
| 3 | `_proc_count("claude")` substring overcount → tight pattern | `f77ce6c` | ✅ done **for the identifier path** (`_probe_claude_pid` now uses `pgrep -n -x claude`). The `_proc_count` *count* fields still use loose `pgrep -af pattern` — acceptable as a non-blocker since the count is used for liveness/pressure rough indicators, not for identification. The single-PID fields that matter for the 3-layer PID model are now tight. |
| 4 | `take_snapshot` rotation concurrency lock | `0ec370e` | ✅ done — `_snapshot_lock(agent)` wraps the full critical section (read prev → compute diff → roll latest→prev → write latest → write diff) under `fcntl.flock(LOCK_EX)` on a per-agent `<agent>.lock` file. POSIX-only as documented; container target is unix so that's fine. Lock fd is released on `__exit__`. |
| 5 | `_probe_screen_count` inverted return | `f77ce6c` | ✅ done (quality-checker's blocker; verified) |

The 2 concurrency regression tests added in `0ec370e` are exactly the kind of belt-and-suspenders coverage I'd want.

### Field compatibility for fleet_watch.sh refactor

After the rename my downstream consumer (`scripts/fleet-watch/fleet_watch.sh`) will need a one-line alias when consuming the new schema:

```bash
# old probe_remote.sh keys:    mem_total_mb / mem_used_mb        (megabytes)
# new snapshot --json keys:    mem_total_bytes / mem_used_bytes  (bytes)
```

This is a clean migration — units are now self-documenting and there's no ambiguity. I'll handle the consumer-side conversion in the fleet_watch.sh refactor PR after this lands.

### Non-blocking observations (still)

- `_proc_count("claude")` / `_proc_count("bun")` / `_proc_count("node")` still use loose substring matching. As long as the surrounding text in the issue acknowledges that these are rough liveness indicators (and not identifiers), this is acceptable. Could be tightened in a future patch.
- `_probe_tmux_pids` still inspects `session:0` only — multi-window agents miss other panes. Not a blocker for any current consumer.
- `context_percent` (top-level) and `agent_meta.context_pct` (nested) still both exist. Documenting the source of truth in a docstring would be enough.

### Approval

**🟢 Ready to merge from head-nas's consumer-side perspective.** All 5 blockers are closed; the design is sound; my fleet_watch.sh refactor (already partially shipped as `b6e5aab` connectivity producer + `7af9fca` connectivity_matrix MCP tool) becomes ~30 lines once #18 lands.

PR #19, opened 2026-04-13
feat(hooks): lifecycle hooks + extensions namespace (todo#286)

ywatanabe1989:
## Summary
Phase 4 of the fleet context-lifecycle series (todo#286 hooks). Stacks on `feat/self-snapshot`.

Adds dependency-inversion hook points so scitex-agent-container stays agnostic of orochi / any specific comms layer — the container declares hooks and pass-through namespaces; external tools (orochi, CI, another hub) plug in via YAML.

- New `hooks.py` (stdlib only): `run_hook()` fire-and-forget dispatcher
  - Entries starting with `http://` / `https://` → POST via `urllib.request` (5s timeout)
  - Otherwise → `shlex.split` + `subprocess.run` (10s timeout, `shell=False`)
  - Module-level `ThreadPoolExecutor(max_workers=4)` for non-blocking submit
  - All errors logged, never raised to the caller
- New YAML fields:

```yaml
spec:
  hooks:
    pre_start:    []
    post_start:   []
    pre_stop:     []
    post_stop:    []
    on_compact:   []
    on_restart:   []
    on_diff:      []
  listen:
    - {port: 8559, proto: tcp, name: mcp_bun, owner: orochi}
  extensions: {}   # opaque pass-through
```

- `lifecycle.agent_start/stop` fire pre/post hooks symmetrically (legacy `_run_hooks` path preserved for shell=True pipe support; legacy skips URL entries to avoid double-dispatch)
- `context_manager.default_dispatcher` fires `on_compact` / `on_restart`
- `snapshot.snapshot_tick` fires `on_diff` when `has_diff`
- `status --json` now emits:
  - `hooks_configured: {pre_start: N, ...}` — counts only, never command bodies (secret leak guard)
  - `listen[]` declarations
  - `extensions{}` echoed verbatim

Container validates `listen[]` but does NOT bind — binding is the hook target's responsibility.

## Test plan
- [x] 12 new tests (`tests/test_hooks.py`): shell success/failure/swallow, http success/failure/swallow, fire-and-forget submit, agent_start pre+post integration, compact dispatcher, snapshot on_diff, extensions passthrough, listen round-trip, `hooks_configured` count-only leak guard
- [x] Full suite: 172 pass / 0 fail (160 baseline + 12)
- [ ] Smoke: register a shell hook on `post_start` in a live agent YAML and observe fire in logs


ywatanabe1989:

**verdict**: approve in principle; one documentation ask, rest non-blocking.

**strengths**
- clean dependency-inversion design: container declares hook points, external tools plug in via YAML
- `shell=False` + `shlex.split` — no command injection surface through YAML strings
- bounded `ThreadPoolExecutor(max_workers=4)` — cannot balloon
- per-call timeouts (HTTP 5s, shell 10s) — bounded blast radius
- errors logged + swallowed, never raised to caller
- HTTP entries detected by `http://` / `https://` prefix — simple and explicit
- `RuntimeError` fallback runs the hook inline on pool shutdown — tests and interpreter-exit paths stay honest
- legacy `_run_hooks` path preserved + skips URL entries to avoid double-dispatch — good migration discipline
- `_flatten_ctx_env` serializes nested dicts/lists as JSON into `SCITEX_HOOK_CTX_*` — reasonable contract
- `shlex.split` wrapped in ValueError guard so a malformed YAML string logs instead of crashing

**documentation ask**

1. Trust boundary: because hook entries are opaque shell commands or URLs taken straight from YAML, any agent whose YAML is writable-by-attacker becomes an RCE vector at `agent_start`/`agent_stop`. This is by design (fleet-author trust model) but please add a short "security model" note in `hooks.py` module docstring or the YAML schema comment: **hook entries are executed as the container user with the container's full env; treat agent YAML as a trust boundary equivalent to root-owned systemd unit files.** Without that note a future contributor might expose this config path to less-trusted sources.

**non-blocking concerns**

2. `_dispatch_shell` passes `env={**os.environ, ...}` — fine, but note that any hook can read container secrets via its env (`SCITEX_*`). If the YAML author and the secrets author are ever different, that's a leak. Low priority; matches current fleet model.

3. `_POOL` is module-level and never shut down — on interpreter exit the daemon threads are torn down, but on test teardown lingering futures can be noisy. If `test_hooks.py` doesn't already, consider a session-scoped fixture that calls `_POOL.shutdown(wait=False)` at the end of the test session.

4. HTTP dispatch builds the JSON body from `context or {}` and passes `default=str` — matches `_flatten_ctx_env` for env vars. Good parity.

5. `_run_one`'s `entry = (entry or "").strip()` then startswith check means a hook entry `"  https://foo"` works — good defensive trim.

**CLI convention**
- new YAML schema fields should appear in `--help` documentation / config schema dump if we have one
- no new subcommands added, so no new `--json` surface — only the in-process dispatcher

ywatanabe1989 (commented):
## Re-review (commit d699fab) — PR #19 hooks + extensions + listen

**Verdict: 🟢 LGTM**

Trust-boundary docstring landed cleanly:

> Trust boundary: entries in a spec.hooks.* list execute arbitrary shell commands or POST to arbitrary URLs with container-scoped privileges. Treat agent YAML files as an RCE-equivalent trust boundary: never load a YAML you don't control, and never inherit hooks from an untrusted source. The container does not sandbox hook execution — that is by design so that operators can wire in system-level integrations.

This is exactly the right framing — it makes the deliberate non-sandbox decision explicit (so operators can wire `systemctl`, `cloudflared`, `notify-send`, etc.) while making the RCE risk legible to anyone reading the YAML schema for the first time.

The Phase 4 design (`hooks.{pre/post_start, pre/post_stop, on_compact, on_restart, on_diff}` + `listen[]` + `extensions{}` opaque pass-through + `hooks_configured` count-only in `status --json`) is well-scoped: container stays agnostic of orochi / telegram / any specific fleet comms, count-only exposure prevents secret leakage through introspection.

177 tests pass, clean rebase on `feat/self-snapshot`. Nothing else for me on this PR.

**🟢 Ready to merge** once #18 lands (stack dependency).

PR #24, opened 2026-04-13
feat(config): search dotfiles agent dir for short-name YAML lookup (todo#295)

ywatanabe1989:
## Summary

`scitex-agent-container start <name>` required a per-host symlink farm at `~/.scitex/agent-container/agents/` because `resolve_config()` only searched that one directory. This PR expands the short-name lookup to also search the dotfiles-managed fleet YAML directory, plus an optional env var for extra dirs — so fleet hosts work out of the box with no manual symlinks.

Fixes todo#295.

## Search order (new)

For short names (no slash, no `.yaml` suffix), `resolve_config(name)` now tries in order:

1. `~/.scitex/agent-container/agents/<name>.yaml` — operator override (existing, highest priority)
2. `$SCITEX_AGENT_CONTAINER_YAML_DIRS` — colon-separated extra dirs (new, middle priority)
3. `~/.dotfiles/src/.scitex/orochi/agents/<name>/<name>.yaml` — fleet fallback (new, lowest priority)

Absolute paths and explicit `*.yaml` paths are returned verbatim (unchanged).

## Diagnostic message

Before:
```
FileNotFoundError: Agent 'foo' not found in ~/.scitex/agent-container/agents/
```

After:
```
Agent 'foo' not found. Searched:
  /home/u/.scitex/agent-container/agents/foo.yaml
  (env $SCITEX_AGENT_CONTAINER_YAML_DIRS: /tmp/a, /tmp/b)
  /home/u/.dotfiles/src/.scitex/orochi/agents/foo/foo.yaml
```

## Test plan

- [x] 7 new tests in `tests/test_resolve_config.py` covering each priority level, colon-separated env var, absolute-path regression, and error message contents.
- [x] Full suite: **200 passed** (`pytest tests/`).
- [x] Stdlib only, no new deps.

### DIFF
--- a/src/scitex_agent_container/config/_resolve.py
+++ b/src/scitex_agent_container/config/_resolve.py
@@ -2,27 +2,75 @@
 
 from __future__ import annotations
 
+import os
 from pathlib import Path
+from typing import List, Tuple
+
+
+_ENV_VAR = "SCITEX_AGENT_CONTAINER_YAML_DIRS"
+
+
+def _search_dirs() -> Tuple[Path, List[Path], Path]:
+    """Return (legacy_dir, env_dirs, dotfiles_dir) with ~ expansion."""
+    home = Path(os.path.expanduser("~"))
+    legacy = home / ".scitex" / "agent-container" / "agents"
+    env_raw = os.environ.get(_ENV_VAR, "")
+    env_dirs = [
+        Path(os.path.expanduser(p)) for p in env_raw.split(":") if p.strip()
+    ]
+    dotfiles = home / ".dotfiles" / "src" / ".scitex" / "orochi" / "agents"
+    return legacy, env_dirs, dotfiles
+
+
+def _try_dir(base: Path, name: str) -> str | None:
+    """Try <base>/<name>.yaml|yml and <base>/<name>/<name>.yaml|yml."""
+    for ext in (".yaml", ".yml"):
+        cand = base / f"{name}{ext}"
+        if cand.exists():
+            return str(cand)
+        cand = base / name / f"{name}{ext}"
+        if cand.exists():
+            return str(cand)
+    return None
 
 
 def resolve_config(name_or_path: str) -> str:
-    """Resolve agent name or path to a config file path."""
+    """Resolve agent name or path to a config file path.
+
+    Search order for short names (no slash, no .yaml/.yml suffix):
+      1. ~/.scitex/agent-container/agents/<name>.yaml  (operator override)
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
## quality review — PR #24 (mamba-quality-checker-mba)

**verdict**: approve. Small, focused, well-tested.

**strengths**
- stdlib-only
- 7 new tests + 200-pass full suite
- three-tier search order is clearly documented and reflected in code
- operator override remains highest priority (no change in expectations for existing installs)
- absolute / explicit `.yaml` paths still pass through unchanged
- diagnostic error lists every searched location — much better than the old single-path message
- `_try_dir` accepts both `<base>/<name>.yaml` and `<base>/<name>/<name>.yaml`, which mirrors how the dotfiles fleet dir is laid out while also gracefully handling flat dirs

**non-blocking nits**

1. `~/.dotfiles/src/.scitex/orochi/agents` is hardcoded. If the dotfiles path ever moves, this breaks silently (the fleet fallback just misses, and the operator override / env var still work). Consider introducing a `SCITEX_DOTFILES_ROOT` env var with the current path as the default, so the override story is complete. Defer — current hardcode is fine for the fleet as it exists today.

2. The colon separator for `SCITEX_AGENT_CONTAINER_YAML_DIRS` is POSIX-only. Windows users would need a different separator. Low priority — the fleet is Unix, and WSL uses POSIX paths. If you want to future-proof, `os.pathsep` is the idiomatic choice.

3. Diagnostic line `(env $SCITEX_AGENT_CONTAINER_YAML_DIRS: <unset>)` may confuse a user who sees `<unset>` and wonders if it's a literal path. Consider: when the env var is not set, omit that line entirely, or render as `(env $SCITEX_AGENT_CONTAINER_YAML_DIRS: not set)` in plain English. Cosmetic.

4. `_try_dir` adds the `<base>/<name>/<name>.yaml` shape to **all** tiers, including the legacy operator dir. That changes what the operator tier matches — previously flat-only, now also subfolder. This is an expansion, not a contraction, so no existing setup breaks. Worth a one-line note in the PR body so operators know the legacy dir has become slightly more permissive.

5. `os.path.expanduser("~")` → `Path.home()` reads cleaner, but not a bug.

**CLI convention**
- no new subcommand, no new `--json` surface
- error path is clear and actionable (lists every searched location)

**CI**: 3.10/3.11/3.12 were IN_PROGRESS at review time; please verify green before merge.

Ready to merge after CI green. No blocker from quality side.

PR #28, opened 2026-04-14
fix(host_identity): derive short-form aliases from DEFAULT_HOST_ALIASES

ywatanabe1989:
## Summary
- The auto-detect loop added full aliases (e.g. "Yusukes-MacBook-Air.local") but did not derive the short form ("Yusukes-MacBook-Air")
- Causes `test_is_local_host_matches_fqdn` CI failure on non-MBA hosts
- Fix: apply `_short()` to each alias entry in DEFAULT_HOST_ALIASES

## Test plan
- [x] 13/13 tests pass locally (including the previously failing test)
- [ ] CI green on GitHub Actions

Fixes todo#399.

### DIFF
--- a/src/scitex_agent_container/host_identity.py
+++ b/src/scitex_agent_container/host_identity.py
@@ -113,7 +113,9 @@ def get_local_identities() -> set[str]:
         alias_lower = {a.lower() for a in aliases}
         overlap = (lowered & alias_lower) - _UNIVERSAL_LOOPBACK
         if overlap or canonical.lower() in lowered:
-            names.update(aliases)
+            for a in aliases:
+                names.add(a)
+                names.add(_short(a))
             names.add(canonical)
 
     _CACHE = {n.lower() for n in names if n}


ywatanabe1989:

verified locally: `pytest tests/test_host_identity.py` → **13 passed** on the PR branch (Darwin, py3.11). Fix is correct — the auto-detect block was adding the full alias (e.g. `Yusukes-MacBook-Air.local`) but not the short form (`Yusukes-MacBook-Air`). The 3-line change threads each alias through `_short(a)` on the way into `names`, which is consistent with how the initial `hn`/`fqdn`/`nn` names are added earlier in the function.

The alternative I proposed (monkeypatch in the test) would also have worked but is strictly worse — it tests monkeypatched state, not the real alias auto-detect path. This fix closes the gap in the source code, not the test, which is the right call.

LGTM, approve. Waiting on CI green to close #399 via `gh-issue-close-safe`.

PR #30, opened 2026-04-14
feat(prompts): runtime prompt handlers to prevent stuck agents

ywatanabe1989:
Adds 3 new PromptHandler entries for runtime prompts not covered by startup auto-accept. Root cause of fleet-stuck incident (ywatanabe msg#10839). Adds tests/test_prompts.py 17/17 pass.

### DIFF
--- a/src/scitex_agent_container/runtimes/prompts.py
+++ b/src/scitex_agent_container/runtimes/prompts.py
@@ -86,6 +86,50 @@ def _detect_skip_permissions_yn(content: str) -> bool:
     )
 
 
+def _detect_mcp_json_edit(content: str) -> bool:
+    """Permission prompt when Claude tries to edit .mcp.json (runtime).
+
+    Matches "1. Yes" / "1. Proceed" / "1. Allow" + ".mcp.json" + "Enter to confirm".
+    """
+    return (
+        ".mcp.json" in content
+        and "Enter to confirm" in content
+        and ("1. Yes" in content or "1. Proceed" in content or "1. Allow" in content)
+    )
+
+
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

Verified locally: `pytest tests/test_prompts.py` → **17 passed** on pr30 (Darwin, py3.11). Approving.

**strengths**:
- `_detect_press_enter_continue` uses the strict last-5-lines window (per pane-state-patterns.md), which matches the root-cause fix for today's scrollback false-positive incidents (msg#10843 / #10865). Excludes `Working…` / `Ruminating…` activity markers and `1. ` / `Enter to confirm` radio prompts. Correct composition.
- Priority ordering is explicit (1–7) and stable via sort after `register_prompt`.
- Every detector is pure and self-contained — easy to add more without touching dispatch logic.
- No behavior change to existing bypass-permissions / dev-channels / thinking-effort handlers; only additive.

**non-blocking nits**:
1. `_detect_file_trust` requires both `trust` and `folder` in the content. The real first-run trust prompt phrases it as "Do you trust the authors of this folder" or similar. Broader capture could drop `folder` and match `"trust the authors"` instead. Not urgent — if the current phrasing is what you observed, keep it.
2. `_detect_press_enter_continue` could false-positive on a tool's stdout that literally echoed "Press Enter to continue" as part of its output. The `Working…`/`Ruminating…` guard catches the common case, but a completed tool that printed that line would still match in the last 5 lines. Low probability; worth documenting.
3. Consider logging which handler fired so post-hoc auto-accept.log can diff-check which detector ate the prompt (useful for tuning).

CI pending — auto-merge after 3 green (3.10/3.11/3.12).

PR #31, opened 2026-04-14
feat(settings_json): seed ~/.claude/settings.json at agent start (todo#423)

ywatanabe1989:
## Summary
- Adds `ensure_global_settings_json()` to `settings_json.py` — called in `ClaudeCodeRuntime.start()` before `setup_settings_json()`
- If `~/.claude/settings.json` is missing or a broken symlink, drops a seed with `skipDangerousModePermissionPrompt`, `promptSuggestionEnabled: false`, and broad `permissions.allow`
- Seed source: `~/.scitex/orochi/templates/claude-code-seed.json` if present, otherwise built-in defaults
- Idempotent: no-op when global settings already resolves

## Root cause fixed
`~/.claude/settings.json` on MBA was a broken symlink to `.dotfiles/.claude-settings.json` (target missing since 2026-03-28). All MBA agents had zero pre-allowed permissions, causing per-tool prompts on every Write/Edit/Bash call.

## Test plan
- [ ] Start a fresh agent on a host with no `~/.claude/settings.json` → verify file is created
- [ ] Start again → verify no-op (file unchanged)
- [ ] Start on host with broken symlink → verify symlink removed, real file created
- [ ] Existing valid `~/.claude/settings.json` unchanged

### DIFF
--- a/src/scitex_agent_container/runtimes/claude_code.py
+++ b/src/scitex_agent_container/runtimes/claude_code.py
@@ -12,7 +12,7 @@ from ..host_identity import is_local_host
 from .base import RuntimeBase
 from .claude_md import cleanup_claude_md, setup_claude_md
 from .mcp_config import cleanup_mcp_config, setup_mcp_config
-from .settings_json import cleanup_settings_json, setup_settings_json
+from .settings_json import cleanup_settings_json, ensure_global_settings_json, setup_settings_json
 from .src_files import (  # noqa: F401
     cleanup_src_claude_md,
     cleanup_src_mcp_json,
@@ -525,6 +525,7 @@ class ClaudeCodeRuntime(RuntimeBase):
         else:
             _setup_claude_md(config, workdir)
         setup_mcp_config(config, workdir)
+        ensure_global_settings_json()
         setup_settings_json(config, workdir)
 
         mux = self._get_mux(config)
--- a/src/scitex_agent_container/runtimes/prompts.py
+++ b/src/scitex_agent_container/runtimes/prompts.py
@@ -86,6 +86,50 @@ def _detect_skip_permissions_yn(content: str) -> bool:
     )
 
 
+def _detect_mcp_json_edit(content: str) -> bool:
+    """Permission prompt when Claude tries to edit .mcp.json (runtime).
+
+    Matches "1. Yes" / "1. Proceed" / "1. Allow" + ".mcp.json" + "Enter to confirm".
+    """
+    return (
+        ".mcp.json" in content
+        and "Enter to confirm" in content
+        and ("1. Yes" in content or "1. Proceed" in content or "1. Allow" in content)
+    )
+
+
+def _detect_press_enter_continue(content: str) -> bool:
+    """Generic 'Press Enter to continue' runtime pause (context-window warning, etc).
+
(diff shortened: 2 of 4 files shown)


ywatanabe1989:

**verdict**: approve in principle — idempotent, targeted, small. Called at the right place (`claude_code.py:528`, just before `setup_settings_json`). One ask.

**strengths**
- Idempotent: healthy existing file → no-op. Broken symlink + missing file → seeded. Good control flow.
- Seed source cascade (template → built-in default) is the right shape.
- `_SEED_DEFAULTS` matches the permissions allowlist used in the auth-manager dotfiles commit 82000b9e — the two paths agree.
- Whitelist `_MANAGED_KEYS` is explicit and cleanup removes exactly those. No key creep.
- `setup_settings_json` correctly merges `enabledMcpjsonServers` as a set-union instead of a clobber.
- Broken symlink path uses `os.unlink` + logged warning on failure; write uses `mkdir(parents=True)` for first-run hosts.

**ask (blocker-ish, small)**
1. **No test for `ensure_global_settings_json`.** This is the exact function that caused todo#423 (broken symlink went unnoticed for weeks because nothing validated it). A regression test is cheap:
   ```python
   def test_ensure_global_seeds_when_missing(tmp_path, monkeypatch):
       monkeypatch.setenv("HOME", str(tmp_path))
       ensure_global_settings_json()
       settings = tmp_path / ".claude" / "settings.json"
       assert settings.exists()
       assert "skipDangerousModePermissionPrompt" in json.loads(settings.read_text())

   def test_ensure_global_repairs_broken_symlink(tmp_path, monkeypatch):
       monkeypatch.setenv("HOME", str(tmp_path))
       (tmp_path / ".claude").mkdir()
       (tmp_path / ".claude" / "settings.json").symlink_to(tmp_path / "nonexistent")
       ensure_global_settings_json()
       assert (tmp_path / ".claude" / "settings.json").is_file()

   def test_ensure_global_is_noop_when_healthy(tmp_path, monkeypatch):
       monkeypatch.setenv("HOME", str(tmp_path))
       (tmp_path / ".claude").mkdir()
       (tmp_path / ".claude" / "settings.json").write_text("{\"existing\": true}")
       ensure_global_settings_json()
       data = json.loads((tmp_path / ".claude" / "settings.json").read_text())
       assert data == {"existing": True}  # untouched
   ```
   `Path.home()` honors `HOME` env in both Linux and macOS so monkeypatch is sufficient.

**non-blocking nits**
2. `is_broken` / `is_missing` flags both use Python's `Path.exists()` which follows symlinks. The logic is correct but worth a one-line comment explaining *why* the pair catches every problematic case (and why they're disjoint).
3. When the template file exists but fails to parse, the fallback silently drops to `_SEED_DEFAULTS`. Consider `logger.warning("template %s unreadable: %s — using built-in defaults", _SEED_TEMPLATE, exc)` so a corrupt template doesn't stay invisible.
4. The built-in `_SEED_DEFAULTS` doesn't include `mcp__scitex-agent-container__*` in the allow list — if you run scitex-agent-container MCP servers as well, add it for parity with the template.

CI pending (3.10/3.11/3.12). LGTM once CI green + 3 regression tests land. Happy to re-review the test commit.

ywatanabe1989:

PR #32, opened 2026-04-14
fix(#396): pre-seed projects[] to skip per-workspace OAuth onboarding

ywatanabe1989:
## Summary
- Add `runtimes/onboarding.py` with `ensure_project_onboarding()` that writes the minimal `~/.claude.json::projects[<workdir>]` entry before agent launch
- Prevents interactive theme/login-method/dev-channels prompts for new workspaces
- Injected into `ClaudeCodeRuntime.start()` before existing setup chain
- Seeded all 19 existing MBA workspace paths immediately (one-time backfill)

## How it works
Claude Code skips onboarding when `hasCompletedProjectOnboarding: true` + `hasTrustDialogAccepted: true` exist in the projects entry. We write these fields (plus empty allowedTools/mcpServers defaults) atomically via temp-file rename before every launch.

Second call on an already-seeded workspace is a no-op.

## Test plan
- [ ] New workspace launched via scitex-agent-container reaches working prompt without OAuth click
- [ ] Existing workspace (already seeded) — no-op, no file mutation
- [ ] Credential rotation (Option B) is a follow-up in #396

Fixes ywatanabe1989/todo#396

### DIFF
--- a/src/scitex_agent_container/runtimes/claude_code.py
+++ b/src/scitex_agent_container/runtimes/claude_code.py
@@ -12,6 +12,7 @@ from ..host_identity import is_local_host
 from .base import RuntimeBase
 from .claude_md import cleanup_claude_md, setup_claude_md
 from .mcp_config import cleanup_mcp_config, setup_mcp_config
+from .onboarding import ensure_project_onboarding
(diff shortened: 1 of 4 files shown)


ywatanabe1989:

PR #34, opened 2026-04-14
fix(spartan): harden head-spartan sbatch wrapper against silent short-exit (todo#425)

ywatanabe1989:
## Summary

- Add `scitex_agent_container.runtimes.sbatch_spartan` as the canonical generator for the head-spartan sbatch wrapper, with five unconditional hardeners against the 2026-04-14 silent short-exit chain
- Add `tests/test_sbatch_spartan.py` with 12 tests guarding every hardener against regression
- Root cause: three hand-edited sibling wrappers on `spartan:~/head_spartan_*.sh` drifted out of sync; one (`head_spartan_restart2.sh`) had no hold block and fell through to clean exit, letting SLURM reap the cgroup in 16s and killing the detached tmux with it

## Evidence of the bug

Five consecutive head-spartan jobs on 2026-04-14 all completed with exit 0 in 15-36s instead of holding their 7-day allocation:

| Job ID    | State       | Elapsed |
|-----------|-------------|---------|
| 23914805  | COMPLETED   | ~16s    |
| 23916429  | COMPLETED   | ~16s    |
| 23934176  | COMPLETED   | ~16s    |
| 23936232  | COMPLETED   | ~16s    |
| 23936277  | COMPLETED   | ~16s    |
| 23936571  | **RUNNING** | 7d held |

Only 23936571 survived because `head_spartan_restart.sh` was edited to add a `while tmux has-session` hold loop seconds before that submission; the other sibling scripts still had the bug.

## Root cause, file:line

`spartan:~/head_spartan_restart2.sh:68` — end of script after `=== PROCESSES ===` diagnostic block with no hold. `set -e` exits clean, SLURM reaps. Also affected: `head_spartan_fresh.sh` (used `exec sleep 604800`, which races at the walltime boundary) and older revisions of `head_spartan_restart.sh`.

## The five hardeners (all enforced by tests)

1. **`set -euo pipefail`** at the top — silent failures crash loud
2. **`exec > ~/slurm_logs/${SLURM_JOB_ID}.out 2>&1`** + **`set -x`** — next short-exit is diagnosable from the log alone, no re-submit needed
3. **`tail -f /dev/null`** as the unconditional persistent hold — idempotent, never decrements, does not race-terminate at the walltime boundary like a literal multi-day `sleep`
4. **Diagnostic branch gated behind `SCITEX_SPARTAN_DIAGNOSTIC=1`** — was the DEFAULT in the hand-edited wrapper and was the actual failure mode; now opt-in only
5. **`trap ... EXIT`** that logs `rc` and exits non-zero — any future regression re-introducing the short-exit pattern is visible in the log with a `todo#425` marker

## Test plan

- [x] `pytest tests/test_sbatch_spartan.py -v` — 12 new tests pass
- [x] `pytest tests/` — all 248 existing tests still pass, 0 regressions
- [x] `bash -n <generated script>` — generated script parses cleanly as bash
- [ ] Next real sbatch submission after merge is the true regression test — leave to head-spartan / operator. Do NOT re-submit from this PR; job 23936571 is currently live on spartan-bm175 and must not be disturbed.

## Constraints respected

- Did NOT edit files on spartan directly (per task instructions)
- Did NOT cancel or re-submit any SLURM jobs
- Did NOT touch `hub/consumers.py` or scitex-orochi (parallel PRs #126/#128/#129/#130 in flight there)

## Follow-up

The spartan hand-edited wrappers (`~/head_spartan_*.sh`) still exist and still drift. This PR delivers the generator + regression test; the wiring (a deploy-time command that writes the canonical script to spartan from this generator) is a follow-up and should be tracked under todo#425 until closed.

Refs: todo#425, msg#11115 (head-mba subagent), msg#11102 (mamba-todo-manager-mba initial diagnosis)

Generated with Claude Code

### DIFF
--- a/src/scitex_agent_container/runtimes/prompts.py
+++ b/src/scitex_agent_container/runtimes/prompts.py
@@ -86,6 +86,50 @@ def _detect_skip_permissions_yn(content: str) -> bool:
     )
 
 
+def _detect_mcp_json_edit(content: str) -> bool:
+    """Permission prompt when Claude tries to edit .mcp.json (runtime).
+
+    Matches "1. Yes" / "1. Proceed" / "1. Allow" + ".mcp.json" + "Enter to confirm".
+    """
+    return (
+        ".mcp.json" in content
+        and "Enter to confirm" in content
+        and ("1. Yes" in content or "1. Proceed" in content or "1. Allow" in content)
+    )
+
+
+def _detect_press_enter_continue(content: str) -> bool:
+    """Generic 'Press Enter to continue' runtime pause (context-window warning, etc).
+
+    Uses a strict last-5-lines window to avoid scrollback false positives
+    (per pane-state-patterns.md: classify against last 5 visible lines only).
+    Excluded: active tool calls and numbered radio selectors.
+    """
+    lines = [l for l in content.splitlines() if l.strip()]
+    last = "\n".join(lines[-5:]) if lines else ""
+    has_enter_cue = (
+        "Press Enter to continue" in last
+        or "press Enter" in last
+        or "Hit Enter" in last
+    )
+    is_active = "Working\u2026" in last or "Ruminating\u2026" in last
+    has_radio = "Enter to confirm" in last or "1. " in last
+    return has_enter_cue and not is_active and not has_radio
+
+
+def _detect_file_trust(content: str) -> bool:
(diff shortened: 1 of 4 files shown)


ywatanabe1989:

Issue #35, opened 2026-04-14
cleanup: define lifecycle policy for src_* files copied into agent dirs (contamination hazard)


ywatanabe1989:
## Summary

`src_*` prefixed files (e.g., `src_mcp.json`, `src_CLAUDE.md`, `src_mcp.json.bak`) exist in every agent directory under `~/.dotfiles/src/.scitex/orochi/agents/`. These are copied into agent dirs by scitex-agent-container during startup, but their lifecycle is undefined — they accumulate in git, drift from source-of-truth, and create cross-agent contamination risk per memory note `project_orochi_infrastructure.md`.

## Problem

**Observed in all 13+ agents on MBA**: every agent dir has `src_mcp.json` and `src_CLAUDE.md`; several have `src_mcp.json.bak`. Example from mamba-todo-manager-mba:

```
~/.scitex/orochi/agents/mamba-todo-manager-mba/src_mcp.json
~/.scitex/orochi/agents/mamba-todo-manager-mba/src_mcp.json.bak
~/.scitex/orochi/agents/mamba-todo-manager-mba/src_CLAUDE.md
```

## Why this matters

1. **Contamination hazard** (per memory `project_orochi_infrastructure.md`): scitex-agent-container's file-copy step can bleed content from one agent to another if `src_*` files are present in the shared tree at startup. This is the root cause referenced in the "newbie contamination" incident (todo#221).

2. **Drift**: `src_*` files are semantically a copy of source-of-truth files elsewhere, but they're committed to git as if they were canonical, causing the real source-of-truth to drift from what agents actually see at boot.

3. **Accumulation**: `.bak` variants suggest in-place editing has happened; original copy intent is unclear.

## Questions for scitex-agent-container maintainer

Pick a lifecycle policy:

**(a) `.gitignore` them globally** — treat `src_*` as ephemeral bootstrap artifacts, never committed. Requires scitex-agent-container to generate them on every startup from a clean template.

**(b) Delete post-bootstrap** — scitex-agent-container copies them in at boot, agent uses them, then the container removes them (or they're cleaned by a post-start hook). Tree stays clean in git.

**(c) Move to `_generated/` subtree** — keeps them visible but clearly scoped as generated; add `_generated/` to `.gitignore` or keep tracked but convention-marked.

**(d) Rename to non-`src_` convention** — if they're supposed to be source-of-truth for the agent, rename them (e.g., `bootstrap_mcp.json`) and document that they are the canonical per-agent config.

[...]

PR #45, opened 2026-04-16
fix(src_files): auto-generate src_mcp.json from canonical template (todo#35)

ywatanabe1989:
## Summary

Implements Option (a) lifecycle policy from #35: `src_mcp.json` is treated as a generated artifact, never committed to git. Regenerated on every agent startup from the canonical template.

- Add `_agent_templates/src_mcp.json`: canonical Orochi MCP server template with `${metadata.name}`, `${metadata.labels.channels}`, `${SCITEX_OROCHI_URL}`, `${SCITEX_OROCHI_TOKEN}` placeholders
- Add `_generate_src_mcp_from_template()`: renders template when `src_mcp.json` is absent from definition dir but `metadata.labels.channels` is set in the YAML
- `deploy_src_mcp_json()`: calls `_generate_src_mcp_from_template` before reading — backward-compatible (if file exists, use it; if absent and channels label present, auto-generate)

**Companion PR**: ywatanabe1989/.dotfiles (gitignore + channels label migration)

## Contamination hazard fix

Root cause of todo#221 ("newbie contamination"): `src_*` files in agent dirs were committed and could bleed between agents during the file-copy step. Under Option (a), `src_mcp.json` is never in git, eliminating the contamination vector at source rather than at symptom.

## Test plan
- [ ] Start an agent after deleting its `src_mcp.json` from the definition dir — verify auto-generation fires and `.mcp.json` in workspace contains correct channels
- [ ] Start an agent that still has `src_mcp.json` on disk — verify existing file is used (backward compat)
- [ ] Start an agent with no `channels` label and no `src_mcp.json` — verify no crash (graceful no-op)

### DIFF
--- a/src/scitex_agent_container/_agent_templates/src_mcp.json
+++ b/src/scitex_agent_container/_agent_templates/src_mcp.json
@@ -1,0 +1,15 @@
+{
+  "mcpServers": {
+    "scitex-orochi": {
+      "type": "stdio",
+      "command": "bun",
+      "args": ["run", "~/proj/scitex-orochi/ts/mcp_channel.ts"],
+      "env": {
+        "SCITEX_OROCHI_URL": "${SCITEX_OROCHI_URL}",
+        "SCITEX_OROCHI_AGENT": "${metadata.name}",
+        "SCITEX_OROCHI_CHANNELS": "${metadata.labels.channels}",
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

Issue #52, opened 2026-04-17
Capture Claude Code statusline JSON via claude-hud wrapper (replace pane-regex scraping)


ywatanabe1989:
## Background

`scitex_agent_container.agent_meta` and client-side `~/.scitex/orochi/scripts/agent_meta.py` both derive `context_pct` from the transcript JSONL by dividing total tokens by 1,000,000. This is an approximation — it doesn't match what `/context` actually reports, and quota percentages (5h / weekly) have to be regex-scraped from the tmux pane (brittle; fails when the pane isn't currently rendering the statusline).

[claude-hud](https://github.com/jarrodwatts/claude-hud) already parses the authoritative source: Claude Code itself pipes a JSON payload into the registered statusline command on stdin (v2.1.6+), containing:

- `context_window.used_percentage` — matches `/context`
- `rate_limits.five_hour.used_percentage`, `rate_limits.seven_day.used_percentage`
- `rate_limits.five_hour.resets_at`, `rate_limits.seven_day.resets_at`
- `model.display_name`, `model.id`
- transcript path, session id, etc.

Reference: `claude-hud/src/stdin.ts` (`getContextPercent`, `getUsageFromStdin`, `getModelName`).

## Decision

Use claude-hud upstream rather than reinventing its parser.

## Proposed approach (no vendoring)

`sac` registers a statusline command in each agent workspace's `.claude/settings.json` that **tees the stdin to a JSON file and then pipes it into claude-hud** for display:

```jsonc
// workspace/.claude/settings.json
{
  "statusLine": {
    "type": "command",
    "command": "~/.scitex/agent-container/bin/sac-statusline"
  }
}
```

```bash
# ~/.scitex/agent-container/bin/sac-statusline (shipped by sac)
#!/usr/bin/env bash
set -euo pipefail
agent="\${SCITEX_OROCHI_AGENT:-unknown}"
out="\$HOME/.scitex/agent-container/statusline/\$agent.json"
mkdir -p "\$(dirname "\$out")"
tee "\$out.tmp" | claude-hud "\$@"
mv "\$out.tmp" "\$out"
```

Then:

- `sac status <agent> --json` reads `~/.scitex/agent-container/statusline/<agent>.json` as the primary source, falling back to the JSONL+approximation path when the file is missing/stale.
- Client scripts (`agent_meta.py --push` in `~/.scitex/orochi/scripts/`) stop regex-scraping pane output and just shell out to `sac status <agent> --json`.

[...]

PR #53, opened 2026-04-17
feat(agent_meta): plan label + rotation tracking + plugins/MCPs audit

ywatanabe1989:
## Summary

- **Fix wrong plan report.** Dashboard showed \`billing_type=stripe_subscription\` as the plan; real plan comes from \`claudeAiOauth.rateLimitTier\`. Expose \`account_plan_label\` ('Max 20x') derived from it.
- **Add auth-rotation tracking.** Whitelist non-secret \`expiresAt\` (unix-ms); when it changes across heartbeats, append one NDJSON line per rotation keyed on email at \`~/.scitex/agent-container/auth-rotations/<email>.ndjson\`. Idempotent.
- **Setup audit for web UI.** Parse \`~/.claude/plugins/installed_plugins.json\` into \`installed_plugins\` list and workspace \`.mcp.json\` into structured \`mcp_servers\` (name, transport, url_host, command). Hub can now show a setup-audit table across the fleet.

Prep work for #52 (claude-hud statusline-JSON capture). Today's audit showed claude-hud is only installed locally on WSL, not fleet-wide — so the setup-audit payload is shipping first so the dashboard can see which hosts need claude-hud installed.

## Test plan

- [x] 491 tests pass locally (full suite)
- [x] Smoke-tested on real \`~/.claude.json\` / \`.credentials.json\`: plan label 'Max 20x', plugins list, rotation log entry written
- [x] Idempotency verified: second call with same \`expiresAt\` — no duplicate NDJSON line
- [ ] Verify payload on a host with no plugins / no credentials (mba/spartan/nas)

## Security notes

- \`expiresAt\` is a unix-ms integer; not a secret. Existing whitelist-only extractor + \`_check_no_secrets\` guard still block access/refresh tokens.
- \`mcp_servers\` surfaces URL host only, not full URLs, so query-string tokens in MCP configs are never exposed.
- New regression test \`test_expires_at_passes_secret_guard\` ensures the guard is not tripped by the new field.

Closes part of #52.

### DIFF
--- a/src/scitex_agent_container/agent_meta.py
+++ b/src/scitex_agent_container/agent_meta.py
@@ -230,6 +230,59 @@ def _read_mcp_json(workdir: str, max_chars: int = 10000) -> str:
         return ""
 
 
+def _parse_mcp_servers(workdir: str) -> list[dict[str, Any]]:
+    """Return a structured summary of MCP servers configured for this agent.
+
+    Parses ``<workdir>/.mcp.json`` into a flat list of
+    ``{name, transport, url_host, command}`` entries so the dashboard
+    can render a setup-audit table alongside installed plugins. URL
+    hosts (not full URLs) and commands (not args) are surfaced because
+    that is enough to verify the server is pointing at the right
+    endpoint without exposing query-string secrets.
+
+    Returns [] if the file is missing or malformed — callers never get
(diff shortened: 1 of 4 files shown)


ywatanabe1989:

PR #74, opened 2026-04-26
feat: skill at-import injection + sac start --dry-run/--json

ywatanabe1989:
## Summary

- Adds `spec.skills.injection_mode: at-import` (default `block` → no change)
- In `at-import` mode, sac resolves each skill name to absolute paths and emits `@<path>` lines in the agent's `.claude/CLAUDE.md` so Claude Code **inlines** the skill content at session start (eager) instead of asking the agent to discover it lazily.
- Resolution: name-as-dir (`<root>/<name>/SKILL.md`) first, then frontmatter `tags:` scan; warns and emits a visible `<!-- not resolved -->` comment if nothing matches.

## Why

Per ywatanabe direction: agents should not be asked to "go read your skills" — the harness should inline them so the agent starts with the rules in its prompt rather than discovering them on demand.

## Test plan

- [x] Existing 621-test suite passes (`block` default unchanged)
- [x] `at-import` mode validated end-to-end with `proj-hello-world.yaml` (resolves `scitex-orochi`, `scitex-agent-container`; warns on `scitex` which isn't in any --add-dir root)
- [x] All three `--add-dir` CLI shapes accepted: `"--add-dir P"`, `"--add-dir=P"`, two-token list

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,13 @@ build/
 .pytest_cache/
 GITIGNORED/
 .env
+
+# cld session artifacts (mirrors of ~/.claude/to_claude/ — canonical source
+# in ~/.dotfiles/src/.claude/to_claude/).
+.claude/
+docs/to_claude/
+
+# dev artifacts
+.coverage
+.venv
+logs/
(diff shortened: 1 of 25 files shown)


ywatanabe1989:

PR #90, opened 2026-04-28
fix(slurm-tenant): provision workspace + cd + split bundled flags

ywatanabe1989:
## Summary

Smoke-tested ``runtime: slurm-tenant`` on Spartan today against a fresh ``fleet-pool`` reservation (sapphire, 32 CPU / 128 GB / 7d / persistent / tmux-server=sac, ``scitex-hpc reservations book``). The agent's tmux session came up on the compute node and ``sac start`` reported success — but the booted ``claude`` opened with ``cwd=$HOME`` and never spoke to Orochi. Three concrete bugs surfaced:

1. **No workspace provisioning.** ``ClaudeCodeRuntime.start`` lays down ``.mcp.json``, ``CLAUDE.md``, ``settings.json``, and v2 ``src_*`` companions before the mux call; ``SlurmTenantRuntime.start`` skipped all of it. Compute-side ``claude`` therefore had nothing to read. Fix: mirror the v1/v2 fork from claude-code (``deploy_src_*`` or ``setup_claude_md``) plus ``setup_mcp_config`` + ``setup_settings_json``. Compute nodes share ``\$HOME`` with login over NFS, so writing locally is sufficient.
2. **No ``cd`` into workspace.** Even with the files in place, ``claude`` needs ``cwd=workspace``. Fix: prefix the tmux command with ``cd <workdir> && exec`` so the new session lands in the agent's workspace dir.
3. **Space-bundled flags shlex.quoted as one token.** A YAML entry like ``"--dangerously-load-development-channels server:scitex-orochi"`` was passed to ``shlex.join`` whole, producing one argv with an embedded space — ``claude`` rejects it. Same for ``"--add-dir ~/proj/foo"``. Fix: ``shlex.split`` each flag entry that both starts with ``-`` and contains whitespace; literal value entries like ``"/path/with spaces"`` (no leading ``-``) stay intact.

## Tests

- ``test_space_bundled_flag_is_split`` — bundled flag + value get split.
- ``test_already_split_flag_pair_stays_intact`` — value with inner space not torn.
- ``test_start_calls_workspace_setup_helpers`` — all four setup helpers invoked with ``expanded_workdir``.
- ``test_start_cds_into_workspace_before_claude`` — rendered new-session embeds ``cd <workdir> && exec claude``.

26/26 pass in ``tests/test_slurm_tenant_runtime.py``.

## Test plan

- [x] ``pytest tests/test_slurm_tenant_runtime.py -v`` → 26/26 pass.
- [ ] Ship + ``pip install -e`` on Spartan (sac 0.10.2 currently there); ``sac start c-scitex-orochi-fr4-agents-tab-detail`` should now produce a tmux session whose ``claude`` reads the workspace's ``.mcp.json`` and registers with the hub. (Manual smoke after merge.)

### DIFF
--- a/src/scitex_agent_container/runtimes/slurm_tenant.py
+++ b/src/scitex_agent_container/runtimes/slurm_tenant.py
@@ -35,6 +35,11 @@ from typing import TYPE_CHECKING
 
 from ..config import AgentConfig
 from .base import RuntimeBase
+from .claude_code import _has_src_files
+from .claude_md import setup_claude_md
+from .mcp_config import setup_mcp_config
+from .settings_json import setup_settings_json
+from .src_files import deploy_src_claude_md, deploy_src_env, deploy_src_mcp_json
 
 if TYPE_CHECKING:
     from scitex_hpc import Reservation as _ResType  # noqa: F401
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

PR #91, opened 2026-04-28
feat(slurm-tenant): forward spec.env so the compute claude finds ANTHROPIC_API_KEY

ywatanabe1989:
## Summary

Stacks on top of #90. Adds env-forwarding to ``SlurmTenantRuntime`` so a spec ``env:`` block reaches the compute-side ``claude`` as inline ``KEY=value KEY2=value2 ...`` assignments before ``exec claude``.

Why this matters: Spartan compute nodes have ``ANTHROPIC_API_KEY2``, ``..._3``, ``SCITEX_*_ANTHROPIC_API_KEY``, etc. set in the user's login profile but **no bare ``ANTHROPIC_API_KEY``**. A fresh ``claude`` boot inside the fleet-pool reservation therefore 401s on its first API call (verified during today's smoke test of ``c-scitex-orochi-fr4-agents-tab-detail``). Without an env channel, the runtime can't redirect those values; operators' only fallback was to touch every compute node's profile.

With this change, the fix is one line in the spec:

```yaml
spec:
  env:
    ANTHROPIC_API_KEY: "${ANTHROPIC_API_KEY2}"
```

The ``${...}`` ref is shlex.quoted but kept literally so the compute-side bash expands it at exec time.

Behaviour:
- ``~``-prefix → expanded to ``$HOME`` literal (compute shell expands at exec).
- ``${VAR}`` refs → kept as shell refs, quoted whole, expanded by compute bash.
- Plain values → shlex.quoted as one token.

## Tests

- ``test_empty_env_returns_empty_string``
- ``test_basic_env_pairs_quoted`` — including a value with internal whitespace.
- ``test_tilde_expanded_to_HOME_for_compute_shell``
- ``test_dollar_var_kept_for_compute_expansion`` — the production case.
- ``test_env_prefix_lands_in_tmux_command`` — asserts the prefix sits between ``cd <workdir>`` and ``exec claude``.

31/31 tests in ``tests/test_slurm_tenant_runtime.py`` pass.

## Test plan

- [x] ``pytest tests/test_slurm_tenant_runtime.py`` → 31/31 pass.
- [ ] After #90 + this merge, set ``env: ANTHROPIC_API_KEY: \"\${ANTHROPIC_API_KEY2}\"`` on a Spartan tenant agent and confirm the fresh claude boot reaches the hub instead of 401-ing. (Manual smoke after merge.)

### DIFF
--- a/src/scitex_agent_container/runtimes/slurm_tenant.py
+++ b/src/scitex_agent_container/runtimes/slurm_tenant.py
@@ -148,6 +148,36 @@ class SlurmTenantRuntime(RuntimeBase):
             parts.append("--continue")
         return shlex.join(parts)
 
+    def _build_env_prefix(self, config: AgentConfig) -> str:
+        """Render ``KEY=value KEY2=value2 ...`` for the tmux launch line.
+
+        Each spec ``env:`` entry is forwarded to the compute-side claude
+        via inline assignment in the bash command (``FOO=bar
+        BAZ=qux exec claude ...``). Values support:
+
+        * ``~``-prefix → expanded to ``$HOME`` literal so the compute
+          shell does the expansion at exec time.
+        * ``${VAR}`` → kept as a shell ref so the compute shell
+          resolves it from its own env (lets agents inherit values
+          like ``$ANTHROPIC_API_KEY2`` that are set in the user's
+          login profile but not in the spec).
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

PR #97, opened 2026-04-29
feat(statusline): sac-statusline wrapper — tee Claude Code JSON to disk (closes #52)

ywatanabe1989:
## Summary
- New sac-statusline CLI entry point tees the Claude Code statusLine JSON payload to ~/.scitex/agent-container/statusline/<agent>.json each turn, then delegates display to claude-hud or falls back to minimal echo
- settings_json now injects statusLine command config into every agent workspace .claude/settings.local.json automatically
- agent_meta.collect_rich() prefers the persisted statusline JSON for context_pct, model, and rate-limit quota fields over the 1M-token JSONL approximation
- statusLine added to _MANAGED_KEYS so cleanup removes it on agent stop

## Test plan
- 13 new tests: persist/read round-trip, atomic rename, corrupt JSON, statusLine written and in _MANAGED_KEYS, cleanup removes it
- 708 total tests pass (5 pre-existing Docker infrastructure failures excluded)

## Upgrade path
Agents need one restart for the new settings.local.json to take effect. Until then sac status continues using the JSONL approximation.

Generated with Claude Code

PR #96, opened 2026-04-29
feat(config): workspace-relative python-venv and env-file in agent YAML

ywatanabe1989:
## Summary
- `python-venv: .venv` now works — relative paths are deferred to the target host at start time (no launcher-side check)
- New `env-file: .env` field loads dotenv files from the workspace before explicit `env:` values (later wins)
- Both accept a single path or a list with the same semantics as the existing `python-venv` fallback chain

## YAML example
```yaml
spec:
  python-venv:
    - .venv         # workspace-local (first hit wins, no launcher check)
    - ~/.venv       # host-home fallback
    - ~/.venv-3.11  # last resort
  env-file:
    - .env
    - .env.local
```

Closes ywatanabe1989/scitex-orochi#260

## Test plan
- [ ] Agent with `python-venv: .venv` starts and activates `<workdir>/.venv`
- [ ] Agent with `env-file: .env` loads variables from `<workdir>/.env`
- [ ] Explicit `env:` values override `.env` file values
- [ ] Missing `.env` file is silently skipped (not an error)
- [ ] Absolute/`~`-prefixed paths still work as before
- [ ] 606 tests pass ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #98, opened 2026-04-29
feat(cli): sac priority-check — host yield decision for singleton reconciler (orochi#250)

ywatanabe1989:
## Summary
- New `sac priority-check <agent>` command: loads the agent YAML host priority chain, SSH-probes each higher-priority host (3s timeout, BatchMode), and reports whether the current host should yield the singleton
- Exit 0 = stay (already on preferred, or all higher hosts unreachable)
- Exit 1 = yield (higher-priority host is reachable and should take over)
- Exit 2 = error (config not found)
- `--json` flag for machine-readable output (healer consumption)
- `--current-host` override for scripted environments

## Healer usage
The healer agent calls this every 2-5 min in its reconciliation tick:
```bash
if ! sac priority-check proj-neurovista --json | jq -e '.should_yield'; then
  echo 'stay'
else
  preferred=$(sac priority-check proj-neurovista --json | jq -r '.reachable_higher_hosts[0]')
  echo "yield to $preferred"
fi
```

## Test plan
- [x] SSH probe: returns False for unknown host, False on timeout
- [x] Priority scenarios: preferred host, local-singleton, fallback-all-down, fallback-with-reachable, not-in-chain
- [x] CLI: exit 0/1/2, JSON output shape
- [x] 716 tests pass

Generated with Claude Code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,7 @@ docs = [
 [project.scripts]
 scitex-agent-container = "scitex_agent_container.cli:main"
 sac = "scitex_agent_container.cli:main"
+sac-statusline = "scitex_agent_container.statusline:main"
 
 [project.entry-points."scitex_dev.docs"]
 scitex-agent-container = "scitex_agent_container"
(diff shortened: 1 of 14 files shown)


ywatanabe1989:
Merged as part of stacked PRs #96+#97+#98+#99 into develop (4c03c78). Content fully included. ✅

PR #100, opened 2026-04-29
feat(event_log): open_agent_calls detection for stuck-subagent alerting (orochi#133)

ywatanabe1989:
## Summary

Adds stuck-subagent detection to `event_log.summarize()` as part of the orochi#133 fleet observability epic.

- **`_compute_open_agent_calls()`**: walks pretool/posttool stream in chronological order; maintains a LIFO stack of open `Agent` invocations; entries still on the stack at the end have no matching posttool → agent may be stuck with an in-flight subagent
- **`open_agent_calls`**: new key in `summarize()` — list of `{ts, input_preview, age_seconds}` entries
- **`open_agent_calls_count` / `oldest_open_agent_age_s`**: scalar summaries in `collect_rich()` for easy threshold checks
- **Terse projection**: both scalar fields added to `TERSE_STATUS_FIELDS` so healers can consume without pulling the ~28KB full payload

## Healer usage pattern

```python
if open_agent_calls_count > 0 and subagent_count > 0 and oldest_open_agent_age_s > 600:
    # likely stuck — raise advisory to head
```

## Caveat

Ring-buffer rotation can produce false positives (a pretool before the window start looks "open" with no posttool). Always cross-check `subagent_count` from the pane before alerting.

## Test plan

- [x] `pytest tests/` — 606 passed, 2 deselected
- [x] 4 new tests: matched call → not open, unmatched → open with age, nested LIFO, no Agent calls → empty

### DIFF
--- a/src/scitex_agent_container/agent_meta.py
+++ b/src/scitex_agent_container/agent_meta.py
@@ -530,6 +530,7 @@ def collect_rich(
             "recent_tools": [],
             "recent_prompts": [],
             "agent_calls": [],
+            "open_agent_calls": [],
             "background_tasks": [],
             "counts": {},
         }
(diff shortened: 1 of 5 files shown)


ywatanabe1989:
Merged into develop via clean rebase+no-ff.

PR #101, opened 2026-04-29
feat(network_probe): probe_cloudflared — bastion tunnel liveness (orochi#133)

ywatanabe1989:
## Summary

Adds cloudflared tunnel process detection to the existing connectivity probe suite (orochi#133 network-layer health sub-item).

- **`probe_cloudflared()`**: `pgrep -f "cloudflared tunnel run"` check; reports `ok=True + extra.pid` when the bastion daemon is alive, `ok=False + err` when absent or `pgrep` unavailable
- **`run_all_probes()`**: now returns 5 probes — `dns → gateway → tcp → https → cloudflared`
- `stx-allow: fallback` annotation on the try/except (pgrep absent on Windows/containers is expected)

## Why cloudflared specifically

`cloudflared tunnel run` is the long-lived bastion daemon that gives the fleet SSH access to MBA. Process alive = tunnel likely healthy (cloudflared auto-reconnects on drop). Process absent = external fleet access to this host is definitely broken. This is the only local, zero-config signal that the external access path works.

## Test plan

- [x] `pytest tests/` — 605 passed, 2 deselected
- [x] 3 new tests: process found → ok+pid, process absent → failure, pgrep missing → graceful failure
- [x] Smoke: `probe_cloudflared()` on MBA returns `ok=True, pid=61719` (live cloudflared daemon)

### DIFF
--- a/src/scitex_agent_container/network_probe.py
+++ b/src/scitex_agent_container/network_probe.py
@@ -301,6 +301,61 @@ def probe_gateway(
         )
 
 
+def probe_cloudflared() -> ProbeResult:
+    """Detect a running cloudflared tunnel daemon on this host.
+
+    Scans the process list for ``cloudflared tunnel run`` (the long-lived
+    bastion daemon). Does not probe the tunnel's remote endpoint — just
+    confirms the local process is alive. If the process is present the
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Merged into develop via clean rebase+no-ff.

PR #103, opened 2026-04-29
fix: bash guard todo424

ywatanabe1989:
test


ywatanabe1989:
Already merged into develop (0 commits ahead of develop).

PR #104, opened 2026-04-29
feat(settings_json): seed ~/.claude.json project entry to skip onboarding (todo#396)

ywatanabe1989:
## Summary
- Adds `seed_claude_json_project_entry(workdir)` to `settings_json.py`
- Called in `ClaudeCodeRuntime.start()` just before `setup_settings_json()`
- Pre-populates `~/.claude.json::projects[workdir]` with `hasTrustDialogAccepted: true` and `hasCompletedProjectOnboarding: true`
- Prevents first-launch trust dialog + OAuth selector from blocking headless agent startup

## Root cause fixed
When scitex-agent-container spawns a new agent, the workspace path is not in `~/.claude.json::projects[]`. Claude Code shows the full onboarding flow (trust dialog, login-method selector, OAuth URL) which blocks headless sessions. This pre-populates the entry so Claude Code sees the project as already onboarded.

## Safety
- Idempotent: no-op when the entry already exists (existing entries are never modified)
- Graceful: silently skips if `~/.claude.json` is absent, a directory, unreadable, or corrupt
- No token material touched — only the `projects` section is written

## Test plan
- [ ] 5 new tests, all passing (`11 passed`)
- [ ] Spawn a new agent on a clean workspace → no trust dialog
- [ ] Spawn again → no change to existing entry

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_main.py
+++ b/src/scitex_agent_container/cli_pkg/_main.py
@@ -13,7 +13,7 @@ from ._helpers import HelpRecursiveGroup
 from .account_cmds import account, quota_watch
 from .action_cmds import actions_cli
 from .build_cmds import build, check, validate
-from .hook_cmds import hook_event
+from .hook_cmds import guard_bash, hook_event
 from .info_cmds import attach, find, list_python_apis, logs
 from .lifecycle_cmds import cleanup, restart, start, stop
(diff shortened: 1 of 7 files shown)


ywatanabe1989:
Already merged into develop (0 commits ahead of develop).

PR #112, opened 2026-05-09
feat(yaml): spec.mounts + spec.home_passthrough; gh CLI in SDK image

ywatanabe1989:
## Summary

Make sac SDK agents able to reach host project trees + run `gh` / `git push` out of the box. Two new YAML knobs:

- **\`spec.mounts: [{src, dst, mode?}]\`** — declarative bind-mounts. Each entry becomes \`--mount type=bind,src,dst[,readonly]\`. \`~\`/\`\$VAR\` expanded on \`src\`.

- **\`spec.home_passthrough: bool\`** — when \`true\`, sac:
  - bind-mounts host \`\$HOME\` at the same absolute path inside the container,
  - sets container \`\$HOME\` to match,
  - runs the container as host \`UID:GID\` (so writes are owned by the operator),
  - forwards \`~/.config/gh\` (read-only) for \`gh\` token store.

Plus:

- **\`Dockerfile.sdk-persistent\`**: install \`gh\` (GitHub CLI) and \`openssh-client\` so home-passthrough'd agents can query CI / open PRs / push over ssh:// remotes.
- **container.py**: bind workdir at its host path *too* (in addition to \`/work\`), so the SDK's \`ClaudeAgentOptions.cwd\` resolves regardless of which path the runner is configured with.

## Why

Spawned 4 polish-* SDK agents earlier today; all 4 reported \`BLOCKED — repo not present in container, gh missing\`. The orchestration pattern works (start → SDK → DONE/BLOCKED back); the container was just missing host context. This PR closes that gap.

## Test plan
- [x] 145 config + container unit tests pass
- [x] 937 full suite passes (+2 new tests for the two knobs)
- [x] Live verification: \`polish-scholar\` SDK agent reads host \`/home/ywatanabe/proj/scitex-scholar/pyproject.toml\` and runs \`gh run list\` from inside the container (with \`home_passthrough: true\`)


ywatanabe1989:
Merged via themed squash into develop (6 commits ce73f1d..006c6cc). Backup of the original 54-commit history preserved on `feat/sac-scale-foundation`.

PR #129, opened 2026-05-19
feat(wi-4): name → host resolver primitive for cross-host routing

ywatanabe1989:
## Summary
Per HANDOFF_AGENT_COMMS_2026-05-19.md §4 (WI-4). Stacks on PR #128 (WI-2). Ships the **resolver primitive** that the cross-host forwarder will sit on; the forwarder itself is gated on a separate decision documented as **Q4 in /work/QUESTIONS.md** (per-host bearer discovery).

- New ``resolve_node_host(name)`` in ``_state/state_db_nodes.py`` — consults ``instances`` table, picks newest live row deterministically (``started_at DESC, id DESC``).
- New ``is_local_node(name, local_host)`` — local-vs-remote decision; treats unknown names as local so external-node implicit registration still works (forwarding a never-seen name would be an SSRF risk).

## Q4 — needs your decision

How does host A know host B's listen bearer for the cross-host forward? Three options described in ``/work/QUESTIONS.md``:

a) Shared fleet bearer (simplest; weakest).
b) Per-host token registry on each host (medium cost; per-host blast radius). Handoff hints at this.
c) Asymmetric crypto + signed JWT (highest cost; cleanest blast radius).

As soon as you pick one I'll add the HTTP forwarding layer that uses these primitives.

## Test plan
- [x] ``pytest tests/.../_state/test_state_db_nodes_resolve.py`` — 8 passed (no mocks; real SQLite + real ``record_instance_start``).
- [x] Local ``scitex-dev ecosystem audit-all`` — audit-python-apis clean.

## Out of scope
- HTTP forwarder layer — gated on Q4 decision.
- Authenticated cross-host trust — same gate.

### DIFF
--- a/src/scitex_agent_container/_state/state_db_nodes.py
+++ b/src/scitex_agent_container/_state/state_db_nodes.py
@@ -45,9 +45,11 @@ from typing import Any
 
 __all__ = [
     "derive_group",
+    "is_local_node",
     "list_node_tokens",
     "mint_node_token",
     "record_lineage",
+    "resolve_node_host",
     "resolve_node_token",
 ]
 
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Superseded — resolver primitives (resolve_node_host / is_local_node) were rolled into PR #128 as part of the WI-2 rework on 2026-05-20 (they share the same _state/state_db_nodes.py module). The cross-host forwarder itself will land on a fresh branch stacked on the amended #128.

PR #156, opened 2026-05-21
feat(a2a): fail loud when an a2a_send push cannot reach or wake the target

ywatanabe1989:
## WI-2 — fail loud when a push can't reach/wake (stacked on #155)

> **Stacked**: this PR targets `feat/wake-on-push` (PR #155). Review/merge #155 first; this diff is WI-2 only.

### Problem
The send-side `a2a_*` tools returned the listen reply verbatim. A push that woke nobody (`delivered_subscriber_count == 0`), a target that was down (connection refused), or a server-side delivery error (5xx) surfaced as a misleading "success" body the calling agent could silently ignore — the message was effectively dropped with no loud signal.

### Fix
`_channel_tools._send_or_raise` + `SendError` make the send path fail LOUDLY. `a2a_send` / `a2a_reply` / `a2a_ack` translate a `SendError` into an explicit `{"error": ...}` tool result (never a misleading success) and log a warning. Fails loud on:

- **transport error** — target agent down / listen unreachable (connection refused),
- **HTTP 5xx** — server / infra delivery error,
- **`delivered_subscriber_count == 0`** — no live inbox subscriber, so the push reached nobody (agent not started, or its channel adapter not connected).

### Design decisions
- **4xx is passed through verbatim, not raised.** The 403 ACL deny carries a structured `body.reason` and is a deliberate, already-loud policy response ("denial is the policy working"). Opaquing it into an error string would lose the `reason` and isn't the silent-drop/misleading-success the rule targets. (This preserves the existing `test_node_comms_e2e_mcp` ACL-deny / identity-spoof contracts.)
- **Absent `delivered_subscriber_count` is NOT treated as zero.** Some replies (cross-host forwards) omit the field; inventing a zero would be a false-positive failure (STX no-false-positives). Only an explicit `0` from the local publish path is the no-subscriber signal.

### Tests (real loopback servers, no mocks)
- no-subscriber (`delivered_subscriber_count == 0`) → loud error
- agent stopped (connection refused) → loud "unreachable" error
- 5xx delivery error → loud error
- delivered>=1 → success; absent count → success (guards against over-firing)
- `a2a_reply` / `a2a_ack` share the fail-loud helper

Full suite green except 3 confirmed pre-existing/environmental failures (verified identical on `develop`: a SDK-dependent handler test, a timing-budget test, and a non-deterministic FastMCP `audit-mcp-tools` probe). None touch the modified code.

### DIFF
--- a/src/scitex_agent_container/_mcp/_channel_tools.py
+++ b/src/scitex_agent_container/_mcp/_channel_tools.py
@@ -16,9 +16,29 @@ other way around at module load).
 from __future__ import annotations
 
 import json
+import logging
 from typing import Any
 
 from .channel import _recent
+
+log = logging.getLogger(__name__)
+
+
+class SendError(RuntimeError):
+    """A send/push could NOT reach or wake the target agent.
+
+    Raised by the send helper when delivery demonstrably failed:
+
+    * the transport raised (agent down / connection refused),
+    * the listen server returned a non-2xx status (delivery error), or
+    * the publish reported ``delivered_subscriber_count == 0`` — no live
+      inbox subscriber, so the message woke nobody.
+
+    The send-side ``a2a_*`` tools translate this into a loud, explicit
+    ``{"error": ...}`` result for the calling agent (never a misleading
+    success) and log it. STX hard rule: fail loudly, never silently drop
+    or return a misleading success.
+    """
 
 
 def register_tools(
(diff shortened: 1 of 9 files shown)


ywatanabe1989:
Superseded by a clean re-base onto develop (fail-loud only; wake landed in #155). Closing to avoid re-applying the squashed wake commits.

PR #249, opened 2026-05-29
feat(account): headless OAuth refresh — `sac accounts refresh [<name>] [--all]`

ywatanabe1989:
## Why
Eliminate routine manual `claude /login` for stored accounts. The OAuth
`refresh_token` is long-lived — short-lived `access_token`s can be minted
from it headlessly without operator interaction. Only when the
`refresh_token` itself lapses is a real `claude /login` required.

This builds on `_refresh_access_token_at(credentials_path, …)` (merged
via #247 + #248). The headless rotation is now a thin, opinionated
public surface over the existing path.

### `_account/claude_usage.py`
- New public **`refresh_account_credentials(credentials_path, *, opener=None) -> dict`**:
  reads tokens at the given file (NOT `~/.claude/.credentials.json`),
  delegates to `_refresh_access_token_at` (atomic write-back to the
  SAME file), reads the freshly-written `expiresAt` back, and returns
  `{success, expires_at (ISO), error, credentials_path}`. Tokens are
  never included in the result.

### `cli_pkg/account_group.py`
- New **`@account.command("refresh")`** surfaced as `sac accounts refresh`:
  - `[<name>]` for a single account, or `--all` for every stored account,
  - `--json` for machine-readable output,
  - per-account snapshot path resolved via the existing `_store_path` cascade,
  - prints `<name>: refreshed; new expiry <ISO>` on success or
    `<name>: FAILED — needs claude /login (<reason>)` on failure,
  - with `--all`, a per-account failure degrades that row only — the run
    continues. The command exits non-zero only when EVERY target failed.
  - Token values are NEVER printed (verified by tests).

### `packaging/systemd/`
- `sac-accounts-refresh.service` + `.timer`: systemd-user files running
  `sac accounts refresh --all` ~15 min after login and every 4 hours
  thereafter, `Persistent=true` for catch-up runs, `TimeoutStartSec=120s`
  bound on a hung token endpoint, fail-loud to the systemd journal.
- `packaging/systemd/README.md`: operator-side install + verify commands.
  **Host install is NOT done by the sac package** — that is a lead
  follow-up. The exact host command is in this PR description below.

## Tests (no mocks, AAA marker comments, one assertion each)
- `tests/_account/test_claude_usage_refresh.py` — **10 tests**: writes new
  token to same file, reports `success=True`, reports ISO expiry, degrades
  on missing `refresh_token` / missing `clientId` / endpoint rejection /
  network error / missing creds file, and does not leak token bytes in
  result keys or values.
- `tests/cli_pkg/test_account_refresh_cmd.py` — **14 tests**: single-name
  happy path (writes back, exit 0, prints expiry, no token leak); `--all`
  writes to every stored account, continues past one bad account, exits 0
  on partial success, exits non-zero when ALL fail, failure message
  mentions `claude /login`; `--json` emits one entry per account with the
  success flag and no token bytes; argument plumbing errors on no-args
  and on conflicting `name + --all`.

## Test results
- New combined: **24 / 24 passed** (~0.9s).
- Full `tests/_account/ + tests/cli_pkg/`: **1299 / 1299 passed** (2
  deselected; was 1275 in merged #248 — same +24).

## Host install (lead follow-up)
```bash
mkdir -p ~/.config/systemd/user
install -m 0644 packaging/systemd/sac-accounts-refresh.service \
                packaging/systemd/sac-accounts-refresh.timer   \
                ~/.config/systemd/user/
systemctl --user daemon-reload
systemctl --user enable --now sac-accounts-refresh.timer
systemctl --user list-timers sac-accounts-refresh.timer
journalctl --user -u sac-accounts-refresh.service -n 50
```

## Compliance
- No mocks (PA-306). AAA marker comments. One assert per test.
- No `Co-Authored-By: Claude` trailer.
- `uv pip install -e .` editable install.
- SoC-safe: sac (`scitex-agent-container`), not a `scitex-*` library
  under the SoC refactor.


ywatanabe1989:
## Live verification against the real Anthropic OAuth endpoint

Lead identified four production-breaking bugs in the initial implementation and shared the verified live recipe; pushed `34dacd2` fixes all four. Verified end-to-end against `console.anthropic.com/v1/oauth/token`:

**Account:** `wyusuuke-gmail-com` (stored at `~/.scitex/agent-container/accounts/wyusuuke-gmail-com/.credentials.json`, no `clientId` stored — exercises the constant fallback).

| Field | BEFORE | AFTER |
|---|---|---|
| `expiresAt` | `2026-05-30T05:53:07.831+00:00` (already expired) | `2026-05-30T07:46:16.227+00:00` (~1h53m forward) |
| `accessToken` | present | present (rotated) |
| `refreshToken` | present | present (**rotated and persisted** — the critical bug fix) |

CLI stdout:
```
  wyusuuke-gmail-com    refreshed; new expiry 2026-05-30T07:46:16.227000+00:00
```

No token values printed to stdout/stderr or any log.

## What changed in `34dacd2`

1. **Endpoint**: `_TOKEN_URL` → `https://console.anthropic.com/v1/oauth/token` (was `claude.ai/api/auth/oauth/token` → Cloudflare 403/404).
2. **Cloudflare headers**: now sends all four — `Content-Type`, `Accept`, `User-Agent: claude-cli/2.1.0 (external, cli)`, `anthropic-beta: oauth-2025-04-20`.
3. **Client ID fallback**: new `_CLAUDE_CODE_CLIENT_ID = "9d1c250a-e61b-44d9-88ed-5944d1962f5e"` constant; stored `clientId` wins when present, constant kicks in when the credentials file omits it (which Claude Code does in practice).
4. **Rotated refresh_token persisted**: `_refresh_access_token_at` now reads `response["refresh_token"]` and writes it back to `oauth["refreshToken"]` under the same flock + atomic .tmp + rename it already used for `accessToken`. The previous behavior would have invalidated the next refresh server-side.

## New test coverage (no mocks, AAA, one assert each)

`tests/_account/test_claude_usage_refresh.py` — **+8 tests** via a capturing opener fixture that intercepts the real `urllib.Request`:

- request URL is `console.anthropic.com/v1/oauth/token`
- `User-Agent: claude-cli/2.1.0 (external, cli)` sent
- `anthropic-beta: oauth-2025-04-20` sent
- `Accept: application/json` sent
- body `client_id` falls back to the constant when stored is missing
- body `client_id` uses the stored value when present
- rotated `refresh_token` persisted to the same file (the critical assertion)
- stored `refresh_token` preserved when response omits rotation (back-compat)

Also replaced the obsolete "errors when no clientId" test with one asserting `success=True` under the constant fallback.

## Test counts

- New refresh suite (`test_claude_usage_refresh.py` + `test_account_refresh_cmd.py`): **32 / 32 PASS**.
- Full `tests/_account/ + tests/cli_pkg/`: **1307 / 1307 PASS** (2 deselected; was 1299 — same +8).

Live recipe verified ✓ — no longer relying on injected-opener tests alone.

PR #252, opened 2026-05-31
fix(dispatch): liveness-gated start + fail-loud propagation + TOFU host keys

ywatanabe1989:
## Summary

Closes the `sac --on spartan-gpgpu011 agents start clew --force --no-preflight` silent-success mode (exits 0 with zero stdout/stderr, no agent ever started on either side). Three independent bugs converge to produce that failure; this PR fixes all three plus the SSH/rsync TOFU policy that compounds them on first-touch peers.

- **`_resolve_singleton_skip` (`cli_pkg/lifecycle/_common.py`)** — honour `--no-redispatch`. The singleton-on-wrong-host skip was a dead end when the propagated remote call had no other host to defer to; it now returns `None` for `no_redispatch=True` so the local start path proceeds.
- **`_verify_real_liveness` (`_lifecycle/_start.py`)** — three-signal "already-running" check. Pre-fix code trusted `registry.exists + runtime.is_running`; both have well-known false positives (stale on-disk JSON, Linux PID reuse). Adds the cross-host `instances` table row as the third independent signal, and falls through to a real start if it disagrees.
- **`propagate_remote_start` (`cli_pkg/_on_start_propagate.py`)** — fail-loud on remote skip/error. A non-zero remote rc now echoes captured stdout/stderr to the operator and returns the rc. A `status != "started"` (other than `dry_run_ok`) now returns rc=1 with the reason on stderr. Pre-fix, both paths dropped the diagnostic on the floor and returned 0/the-rc with nothing visible to the operator.
- **TOFU host keys (`_state/_host_ssh.py`, `cli_pkg/lifecycle/_dispatch.py`)** — `StrictHostKeyChecking=accept-new` baked into `build_ssh_argv` AND the rsync `-e` transport used by `_dispatch_remote_start`. A first-touch dispatch to a freshly-registered peer no longer rc-1s silently under `BatchMode=yes`.

## Test plan

- [x] 22 new tests across `test__common.py`, `test__on_start_propagate.py`, `test__dispatch.py`, `test_host_config_env_preamble.py`, `test_lifecycle.py`. No mocks — every seam (liveness verifier, runner, instances oracle) is a real callable wired through `tmp_path` state.db where needed.
- [x] Full suite green except for 7 failures that exist verbatim on `develop` (audit/`test_audit_all_clean`, `cli_startup_budget`, `sdk_common cwd`, statusline ×4, `a2a/handlers` ×1) — none touch the modified code paths.
- [x] Conforms to STX-TQ002 (AAA markers), STX-TQ003 (descriptive names), STX-TQ007 (one assertion per test).

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_start.py
+++ b/src/scitex_agent_container/_lifecycle/_start.py
@@ -25,6 +25,63 @@ from ._spawn_gate import enforce_spawn_gate, persist_acl_policy
 from .health import health_monitor
 
 
+def _verify_real_liveness(
+    config: AgentConfig,
+    runtime,
+    *,
+    instances_oracle: Callable[[], list[dict]] | None = None,
+) -> bool:
+    """Return True iff the agent is *demonstrably* running on this host.
+
+    The pre-fix call site treated ``registry.exists(name) and
+    runtime.is_running(config)`` as the already-running signal. Both
+    are necessary but not sufficient:
+
+      * ``registry.exists`` is a JSON file on disk — a forced ``rm``,
+        a stale entry from a prior boot, or a crash-during-write leaves
+        the file behind even though no agent is running.
+      * ``runtime.is_running`` checks the per-runtime PID file with
+        ``os.kill(pid, 0)`` — on a Linux PID-wraparound the same pid
+        can belong to a completely unrelated process and the probe
+        returns True.
+
+    Either of those false positives causes the no-op branch in
+    :func:`agent_start` to swallow a real start request silently and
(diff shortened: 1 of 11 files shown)


ywatanabe1989:
@ywatanabe1989 ready for review — fix/dispatch-fail-loud → develop. Closes the silent-success mode reproduced as `sac --on spartan-gpgpu011 agents start clew --force --no-preflight`. A2A relay was unreachable from this worktree so posting here as the durable notify path. 22 new tests, 5500/5508 pass (8 fails are pre-existing develop-baseline; none touch modified code paths).

PR #253, opened 2026-05-31
fix(dispatch): release stale singleton binding when bound host unreachable

ywatanabe1989:
## Summary

Follow-up to PR #252's fail-loud start path. Closes the **clew #1-priority blocker**: clew was pinned in the registry to a dead prior host (spartan-bm025, no SLURM job → `pam_slurm_adopt` denies any ssh). Both the new singleton-skip and `stop --force` chased the dead host and hung:

- `sac --on spartan-gpgpu011 agents start clew` → "singleton prefers spartan-bm025, current host spartan-gpgpu011, remote status=skipped" (PR #252's fail-loud caught the silent skip, but the underlying stale binding still needed a release path).
- `sac agents stop clew --force` aborted on rc=255 ssh transport failure to bm025, so the stale `instances` row + `comms_nodes` pin never cleared — every subsequent routing decision re-tried the dead host.

Two changes, same robustness class as PR #252's three-signal `_verify_real_liveness`:

1. **Liveness-gated `_resolve_singleton_skip`** (`cli_pkg/lifecycle/_common.py`): the skip now consults a `BoundHostLivenessOracle` (default: `state_db.list_active_instances`). If the bound host has no active row for this agent, the binding is stale → return `None` → fall through to a local start. If the bound host has a live row, the original "skip and defer" preserved.

2. **`stop --force` releases stale binding on unreachable peer** (`cli_pkg/lifecycle/_stop.py`): `_dispatch_remote_stop` now raises a typed `_PeerUnreachableError` for ssh transport failures (rc!=0). The stop loop catches it ONLY when `--force` is set and falls through to `_force_release_binding`, which tombstones the instances row with `exit_reason='peer-unreachable-force-released'` AND clears the `comms_nodes` pin. Without `--force`, the ssh failure still surfaces as an error.

JSON envelope on the force-release path carries `"force_released": true` so wrapping automation can distinguish a clean cross-host stop from a stale-binding release.

## Test plan

- [x] 8 new no-mocks tests for stop --force release path (real `state.db`, PATH-prepended fake ssh emitting rc=255 with a `pam_slurm_adopt` stderr).
- [x] 5 new tests for the singleton-skip liveness gate (live oracle preserves skip; dead oracle releases binding; oracle scope + call shape pinned).
- [x] Existing `TestSingletonHostSkip` / `TestResumeAndForeground` integration tests updated to populate a live `state.db` row so the gate's release semantics don't fire unexpectedly.
- [x] Full suite green except for pre-existing develop-baseline failures (audit, cli_startup_budget, statusline ×4, a2a/handlers, sdk_common cwd).
- [x] STX-TQ002 (AAA markers), STX-TQ003 (descriptive names), STX-TQ007 (one assertion per test).

### DIFF
--- a/src/scitex_agent_container/cli_pkg/lifecycle/_common.py
+++ b/src/scitex_agent_container/cli_pkg/lifecycle/_common.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 
 from collections.abc import Mapping
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
 
 import click
 
@@ -19,6 +19,11 @@ from ...config import AgentConfig
 
 if TYPE_CHECKING:
     from ..._state.host_config import PeerSpec
+
+# (name, host) -> True iff the registry holds an active instances row
+# for ``name`` on ``host``. Used by :func:`_resolve_singleton_skip` to
+# liveness-gate the spec-host preference.
+BoundHostLivenessOracle = Callable[[str, str], bool]
 
 _SKIP_DIR_NAMES = {"legacy-agents", "shared", "GITIGNORED"}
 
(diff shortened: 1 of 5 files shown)


ywatanabe1989:
@ywatanabe1989 follow-up to #252. Closes the **clew #1-priority blocker** — see PR body for the bm025 stale-binding repro + the two fixes (liveness-gated singleton skip + stop --force release-on-unreachable). Release commands for the existing stale rows on lead/Spartan are in the chat (clear instances + comms_nodes + rsync the spec.yaml). Full suite green except 7 pre-existing develop-baseline fails (down from 8 — cli_startup_budget passed this run). 30/30 stop tests + 19/19 new common tests green.

PR #259, opened 2026-05-31
fix(accounts): tz-local + live TTL + table renderer for `sac accounts list`

ywatanabe1989:
## Summary

Three operator-reported regressions in the human renderer of `sac accounts list`, all fixed surgically without touching the JSON output shape.

- **TZ**: renderer was emitting UTC ISO with microseconds. Now renders in operator-local timezone with precedence `SCITEX_AGENT_CONTAINER_TZ` > `TZ` > system local (`datetime.astimezone()` no-arg). Project-specific env wins so a host-wide `TZ` never overrides an explicit per-tool preference.
- **Live TTL**: `account_freshness` already recomputed `expiresAt` on every call, but the display format `+2.8h` rounded sub-hour ticks to the same string so `watch -n1` showed no movement. Reformatted as `+2h48m` / `-138h36m` / `+45s` — a 60-second tick is now visible. Usage% (5h/7d) still uses the 5-min per-account cache to avoid hammering the upstream API; the new **As-of column shows the snapshot age** (`Sun 21h (3m)`) so a stale number is obvious. New `--refresh` / `--live` flag busts the on-disk `usage.json` cache before render.
- **Readability**: printf-style line-per-account replaced with `rich.table.Table` aligned across `ID | Email | Plan | Status(+TTL) | 5h% | 7d% | As-of`. As-of column is short day+hour form (`Sun 21h`), not microsecond ISO.

JSON path (`--json`) is unchanged — still emits ISO-8601 `as_of` / `fetched_at` for downstream consumers. Only the human renderer reformats at display time.

## Structure

Renderer extracted into `src/scitex_agent_container/cli_pkg/_account_list_render.py` (same pattern as `_account_status.py` / `_account_refresh.py`) so `account_group.py` stays under the per-file line cap and the renderer is unit-testable without `CliRunner`. Pure functions take a `now=` injection seam where applicable; TTL liveness is asserted end-to-end by writing two real snapshots 60s apart and confirming the rendered TTL string differs.

## Test plan

- [x] `local_timezone()` returns `None` with no env, the resolved zone with `TZ` set, and `SCITEX_AGENT_CONTAINER_TZ` wins over `TZ`.
- [x] Bad `SCITEX_AGENT_CONTAINER_TZ` value falls through to `TZ`.
- [x] `format_dt_local()` renders the same UTC instant with the correct `+09:00` offset under JST.
- [x] `format_ttl_live()` exposes a 60-second tick (two values differing by 60s render to different strings).
- [x] `format_snapshot_age()` produces `30s` / `3m` / `1h` / `2d` and clamps future timestamps to `0s`.
- [x] `format_as_of_short()` emits `Sun 21h` (no microseconds, no offset, no `T`).
- [x] `render_stored_table()` includes every expected column header and aligns separators across header + data rows.
- [x] As-of cell renders the short form (no microseconds) and includes the age suffix `(3m)`.
- [x] `build_stored_rows()` produces different `freshness_hours` for two calls 60s apart on the same on-disk state — and the rendered TTL string differs accordingly.
- [x] `sac account list --refresh` / `--live` exit 0 and delete the per-account `usage.json` cache before rendering.
- [x] Human path renders the new table columns (`ID`, `Email`, `Plan`, `Status(+TTL)`, `5h%`, `7d%`, `As-of`); empty-data cells render as `-`.
- [x] JSON path (`--json`) unchanged: `as_of` stays ISO-8601, no day-of-week tokens leak through.
- [x] `tests/scitex_agent_container/cli_pkg/test_account_list_render.py` — 33 tests, all green.
- [x] `tests/scitex_agent_container/cli_pkg/test_account_group.py` — 55 tests, all green (one assertion updated to reflect the new table cells).
- [x] Adjacent suites (`test_account_group_{sync_live,status,refresh,refresh_skip_active}`, `_state/test_account_store{,_plan_usage}`) — 66 tests, all green.

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_account_list_render.py
+++ b/src/scitex_agent_container/cli_pkg/_account_list_render.py
@@ -1,0 +1,474 @@
+"""Renderer for ``sac accounts list`` — Stored-accounts table + helpers.
+
+Split out of ``account_group.py`` to keep that file under the per-file
+line cap (same pattern as ``_account_status.py`` and ``_account_refresh.py``)
+and so the renderer is unit-testable without going through ``CliRunner``.
+
+Three concerns live here, one per public helper:
+
+1. :func:`local_timezone` / :func:`format_dt_local` — render an ISO-8601
+   timestamp in the operator's local timezone. Precedence:
+   ``SCITEX_AGENT_CONTAINER_TZ`` env wins, else ``TZ`` env, else the
+   system local timezone (``datetime.astimezone()`` with no arg). This
+   is the bullet-1 fix: the prior renderer emitted UTC.
+
+2. :func:`format_ttl_live` / :func:`format_snapshot_age` — render the
+   credential TTL and the per-account usage snapshot age with enough
+   resolution that a 60-second tick is VISIBLE under ``watch -n1``. The
+   prior ``+2.8h`` collapsed any sub-hour change into the same string,
+   making the operator think the value was cached. TTL is computed
+   live from ``expiresAt`` on every call; this module only formats it.
+
+3. :func:`render_stored_table` — render the Stored-accounts block as a
+   ``rich.table.Table`` with aligned columns:
+
+       ID | Email | Plan | Status(+TTL) | 5h% | 7d% | As-of
+
+   ``As-of`` is the per-account usage snapshot age in short day+hour
+   form (``Sun 21h``), not microsecond ISO.
+
+The JSON-output path of ``sac accounts list --json`` is NOT touched —
+it still emits ISO-8601 ``as_of`` strings. Only the human renderer
+reformats at display time.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from datetime import datetime, timezone, tzinfo
+
+from rich.console import Console
+from rich.table import Table
+
+# ---------------------------------------------------------------------------
+# Timezone resolution (bullet 1)
+# ---------------------------------------------------------------------------
+
+# Project-specific env wins over the standard POSIX ``TZ`` so a host-wide
+# ``TZ`` doesn't accidentally override an explicit per-tool preference.
+_PROJECT_TZ_ENV = "SCITEX_AGENT_CONTAINER_TZ"
+
+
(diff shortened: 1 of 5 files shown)


ywatanabe1989:
## Branch refreshed (force-push-free)

The original CI failure (pytest-matrix 3.11/3.12/3.13) was from a stale base: the branch was 12 commits behind develop. Just landed:

- `5e3bf44` — merge origin/develop in (no conflicts; brought in v0.21.5–v0.21.9 + ACL broker + writable-creds bind + smoke deny-row fixes).
- `8ab9a56` — `test(accounts): satisfy STX-TQ002 AAA-markers + TQ007 one-assert on account-list renderer tests`. Renamed `test_account_list_render.py` → `test__account_list_render.py` (double-underscore private-test convention; same shape as `test__apptainer_creds_rw_bind.py`) and split multi-assert tests into single-assert AAA blocks. Same audit-driven cleanup as commit 2ff5245 did for the writable-creds tests.

Local run on the merged branch (`/opt/venv-agent/bin/python -m pytest tests/`): **5764 passed, 44 skipped, 9 deselected** in 24m55s.

### Two local-only skips, both pre-existing on develop (not introduced by this PR)

1. `tests/develop/test_audit.py::test_audit_all_clean` — `scitex-dev` 0.16.0/0.16.1 `_read_bridge_source` crashes with `TypeError: argument should be a str or an os.PathLike object … not 'NoneType'` when `scitex._mcp_tools` is a namespace package (`__file__=None`). Reproduces against `origin/develop` locally; develop CI passes because the GitHub runner venv doesn't have the umbrella installed. Upstream `scitex-dev` bug — separate fix to land there. Not hidden; called out here.
2. `tests/integration/test_cli_startup_budget.py::test_cli_help_under_budget` — env-bound 0.5s budget; this agent container measures 0.745s for `--help` even at `origin/develop`. CI GitHub runners are faster and pass. Pre-existing env quirk, not introduced by the renderer extraction (which only adds lazy imports inside `account_group:_account_list_render` invocation paths).

Both are documented for the reviewer; not skipped silently in CI configuration.

PR #283, opened 2026-06-01
feat(agents-list): per-agent CPU% + RSS in `sac agents list` / `status --json`

ywatanabe1989:
## Summary

Operator wanted per-agent CPU% and memory in `sac agents list` so an overloaded host can be attributed to specific agents. Today only host-wide load is surfaced. This PR adds the per-agent attribution everywhere it makes sense, with no measurable added latency:

- **`sac agents list` human table**: two right-aligned columns `CPU%` and `MEM`, rendered as `-` when the recorded PID is dead / unknown / unprobable.
- **`sac agents list --json`**: `cpu_percent` (float) and `mem_rss_mb` (float) per row, **absent** (not 0.0, not null) when not probed — observability contract `absent ≠ 0` from skill leaf `13_observability.md`.
- **`sac agents status --json`**: same two fields on the single-agent payload, sourced from the registry-recorded PID.
- **`TERSE_STATUS_FIELDS`**: extended so fleet hubs reading the `--terse` projection get attribution by default.

## Implementation

- **New helper** `scitex_agent_container._state._meta.resources.collect_agent_resources(pids)`: walks each PID's process tree (root + recursive descendants) in a single psutil sweep, sums CPU% and RSS, returns `{pid: {"cpu_percent": float, "mem_rss_mb": float} | None}`. CPU% uses ONE batched `cpu_percent(interval=None)` prime / `time.sleep(0.1)` / readout cycle — wall-clock cost is constant in agent count, not linear. Dead / unknown PIDs map to `None`; empty input → empty output (no sleep, no psutil call); psutil-unavailable returns all-`None` rather than crashing.
- **List integration** in `cli_pkg/_helpers/_agent_list.get_agent_list_data`: collects all running PIDs into ONE batched probe before building the result rows; row keys merged in only when the per-PID payload is not `None`. Defined-but-not-registered agents (no PID) skip the probe.
- **Status integration** in `_lifecycle/_status.agent_status`: probes the single registry-recorded PID directly; same `absent ≠ 0` rules.

## Test plan

PA-306 no-mocks — every test exercises real OS processes (`os.getpid()` for live PID, `subprocess.Popen(python)` for descendants, `2**31 - 1` for dead PID). AAA markers (TQ002), one assert per test (TQ007), 3+-word descriptive names.

- [x] **Probe surface** (`tests/.../_state/_meta/test__agent_resources.py`, 10 tests): keyspace contract, both keys present on live PID, mem_rss_mb is positive float, cpu_percent is float, dead-PID → None, never-running-PID → None, descendants summed in, mixed live+dead in one sweep, empty input → {}, PID=0 sentinel → None.
- [x] **List integration** (additions to `tests/.../cli_pkg/_helpers/test__agent_list.py`, 8 tests): running row carries both keys, dead-PID row omits them, PID=0 sentinel row omits them, cpu_percent is float, mem_rss_mb is positive, table renders CPU%/MEM headers, dead-PID rows render `-` cells, JSON emits cpu_percent for live PID, JSON omits cpu_percent for dead PID.
- [x] **Status integration** (additions to `tests/.../_lifecycle/test_lifecycle.py`, 5 tests): live-PID status carries cpu_percent, value is float, mem_rss_mb is positive, dead-PID omits cpu_percent, dead-PID omits mem_rss_mb.
- [x] **Ruff**: all touched files pass `ruff check`.

Local: all 76 `test_lifecycle.py` tests + all 40 `test__agent_list.py` tests + all 10 `test__agent_resources.py` tests pass on a heavily loaded host (~load 42 during development). CI is the gate.

## Out of scope (deliberately)

- psutil is NOT moved from optional → required dependency. The existing `_collect_host_metrics` already treats it as optional; the new code mirrors that contract so installs without psutil still render the list / status (rows just omit the two new fields).
- No new flag (`--no-resources`, etc.). The default 100ms batched sleep is below "noticeable" for an interactive command.
- `agent_meta.collect_rich` (heartbeat-grade per-agent metadata) NOT touched — its PID resolution path uses tmux session inspection that doesn't work for apptainer agents. Status-side integration uses the registry-recorded PID directly, which is the canonical source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,35 @@ versioning follows [SemVer](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- **feat(agents-list): per-agent CPU% + RSS in `sac agents list` and
+  `sac agents status --json`** (lead task 2026-06-01).
+  - New helper ``_state._meta.resources.collect_agent_resources``:
+    given an iterable of PIDs, walks each one's process tree (root +
+    descendants) and returns ``{"cpu_percent": float, "mem_rss_mb":
+    float}`` per PID via a SINGLE psutil sweep (one batched ~100ms
+    sleep covers every agent's cpu_percent delta — wall-clock cost is
+    constant in agent count). Dead PIDs / unknown sentinel ``0`` /
+    psutil-unavailable map to ``None`` rather than zero-filled — the
+    observability contract from skill leaf
+    ``13_observability.md`` (absent ≠ 0) holds end-to-end.
+  - ``sac agents list`` rows now carry ``cpu_percent`` and
+    ``mem_rss_mb`` for every running agent whose PID resolves;
+    rendered as right-aligned ``CPU%`` and ``MEM`` columns in the
+    human table (``-`` cell when absent). JSON path emits the keys
+    only when present — downstream fleet hubs / dashboards can
+    distinguish "not probed" from "literally idle".
+  - ``sac agents status --json`` carries the same two fields for the
+    single-agent path, sourced from the registry-recorded PID.
+  - ``TERSE_STATUS_FIELDS`` (``--terse`` projection) extended with
+    ``cpu_percent`` and ``mem_rss_mb`` so fleet hubs reading the
+    terse blob get attribution by default.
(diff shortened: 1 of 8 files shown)


ywatanabe1989:

PR #287, opened 2026-06-03
feat(SAC-from-SAC): fail-loud bind preflight + STARTUP_FAILED lifecycle (PR-1)

ywatanabe1989:
## Summary

**PR-1 of the SAC-from-SAC hardening series.** Motivated by the clew `cohort-A/capsule-0201225` incident on 2026-06-02: the inline-spec `POST /agents` handler accepted a spec whose `apptainer.binds` carried parent-SIF-view paths (`/work/...`) the host couldn't see, the subprocess to `sac agents start` exited 0 after handoff, but apptainer FATAL'd minutes later on the missing mount source. The operator spent 50 minutes triaging three differently-shaped "session never started" symptoms because the failure had no structured surface.

This PR closes the gap with two coordinated changes — a **pre-flight** that catches the bad spec **before** materialisation, and a **lifecycle marker** that turns a stillborn agent into an observable resource.

### PART 1 — Fail-loud bind preflight
- NEW `_listen/_inline_spec_preflight.py` — structured detector. Walks `spec.apptainer.binds`, host-`stat()`s every source after `~`/`$VAR` expansion, returns `PreflightResult{ok, checks, unresolvable}`. Accepts string + dict bind forms. Symlinks followed; broken links count as missing.
- MOD `_listen/_inline_spec.py` — preflight runs **before** spec write. On failure: HTTP 400 with structured body. Spec is NOT materialised on rejection.
- Added stable `kind` field to existing 400/409 responses (`spec_invalid`, `already_exists`) so every wire-shape failure now branches by `kind`, not prose.

### PART 2 — STARTUP_FAILED lifecycle
- NEW `_lifecycle/_startup_failed.py` — versioned JSON marker at `runtime_dir/STARTUP_FAILED`. Atomic write (`.tmp` + `os.replace`). Apptainer FATAL classifier (`apptainer_mount_failed` / `sif_invalid` / `disk_full` / `container_creation_unknown`) stamps `kind` + `remediation_hint`. stdout/stderr tail-clipped to 8 KiB (FATAL line survives). Marker carries its own host-absolute `runtime_dir` so consumers can `cat` peer `stdout.log` / `stderr.log` without recomputing paths.
- MOD `_listen/_agent_exec.py` — on `sac agents start` non-zero exit, write marker into `state_dir_for(name)`. Best-effort: marker-write failure is swallowed so it never shadows the spawn failure.
- MOD `_listen/server.py` — DELETE returns **410 Gone** with a structured **flat summary + `see_also` pointer + nested `details`** body when marker present; falls back to 404 only when agent never existed. STATUS echoes marker under `startup_failed` key + sets `status="startup_failed"`.

## Wire-shape contract — POST-clew-review (commit cd4c7e2)

Clew's review (msg `8306752c...`) drove five key renames + restructure. The final shape consumed by the clew launcher is:

```
POST /agents 400 (bind_unresolvable):
  {
    "error": "bind source(s) not visible from host",
    "kind":  "bind_unresolvable",
    "details": {
      "binds": [                            # array form (was "unresolvable")
        {
          "source":          "/work/x:/x:ro",  # was "bind"
          "container_path":  "/x",             # was "container_dest"
          "exists_on_host":  false,
          "host_normalized": "/home/y/x"       # ONLY when ~/$VAR expansion changed source
        }
      ],
      "translation_hint": "..."             # was "remediation_hint"
    }
  }

POST /agents 400 (spec_invalid):     {"error", "kind": "spec_invalid"}
POST /agents 409 (already_exists):   {"error", "kind": "already_exists"}

DELETE /agents/<name> 410:
  {
    "name":             "<name>",
    "status":           "startup_failed",         # lifecycle tag (mirrors STATUS)
    "kind":             "<failure_classification>", # LIFTED from marker
    "phase":            "container_creation",     # LIFTED from marker
    "failed_at":        "<ISO>",                  # LIFTED from marker
    "runtime_dir":      "<host abs path>",        # LIFTED from marker
    "remediation_hint": "<...>",                  # LIFTED from marker
    "see_also":         "<runtime_dir>/STARTUP_FAILED",
    "details":          <full marker payload>
  }

DELETE /agents/<name> 404: {"error": "no pid file"}  (never existed)

GET /agents/<name>/status (stillborn):
  {..., "status": "startup_failed", "startup_failed": <marker>}
```

### kind taxonomy (PR-1 freeze)

- `POST /agents` 400/409: `bind_unresolvable` | `spec_invalid` | `already_exists`
- `STARTUP_FAILED.kind`: `apptainer_mount_failed` | `sif_invalid` | `disk_full` | `container_creation_unknown`
- PR-3 adds: `acl_deny` (POST/DELETE/send/tail), `transport` (client-side detect)

`acl_deny` is punted to PR-3 (PR-1's POST has no ACL gate yet, so adding the kind here would be dead code with no test pin).

## Test coverage (real I/O via Starlette TestClient, no mocks; AAA + one-assert)

- `tests/.../_listen/test__inline_spec_preflight.py` — **24 tests** (happy / expansion / failure / defensive parsing / wire shape incl. host_normalized omit-and-include cases, container_path, translation_hint)
- `tests/.../_lifecycle/test__startup_failed.py` — **23 tests** (classify / write_marker / read/is_stillborn / runtime_dir field)
- `tests/.../_listen/test_server_inline_spec_failloud.py` — **10 tests** (wired POST /agents fail-loud + non-regression on existing 400 kinds)
- `tests/.../_listen/test_server_startup_failed.py` — **15 tests** (wired DELETE 410 top-level status/kind/phase/runtime_dir/see_also lifts + STATUS surface + non-regression on 200 OK pid path)

**Total: 72 new tests.** All 672 `_listen + _lifecycle + _runners/test_claude_session` tests pass locally. PA-306 (no-mocks) + PA-307 (AAA + one-assert) clean for this PR's diff.

## Series sequencing

This is PR-1 of three:
- **PR-1** (this): fail-loud bind preflight + STARTUP_FAILED lifecycle
- PR-2: SAC-side bind translate (`/work/...` → caller host-workdir)
- PR-3: `sac agents spawn-from-here` + in-SIF auto-fallback for `tail/status/send/delete` + lineage-scoped ACL + `acl_deny` kind

Refs: SAC-from-SAC PR series. Operator-driven via clew. Wire-shape signed off by clew (`f36cc02f...` thread).

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_startup_failed.py
+++ b/src/scitex_agent_container/_lifecycle/_startup_failed.py
@@ -43,10 +43,18 @@ so we can extend without breaking the existing readers:
      "phase":        "container_creation",
      "kind":         "apptainer_mount_failed",
      "exit_code":    255,
+     "runtime_dir":  "<host abs path to per-instance state dir>",
      "stdout_tail":  "<last N lines>",
      "stderr_tail":  "<last N lines>",
      "remediation_hint": "..."
    }
+
+The ``runtime_dir`` field is the host-absolute path to the per-instance
+state directory carrying this marker (and any peer ``stdout.log`` /
+``stderr.log`` if the spawn captured them). Echoing it in the JSON
+spares wire-shape consumers (clew launcher, ``sac agents delete``
+410 body) having to recompute or guess the path — the marker is
+self-describing.
 
 Tail capture is bounded so a runaway log doesn't bloat the marker.
 """
@@ -196,6 +204,12 @@ def write_marker(
         "phase": phase,
         "kind": kind,
         "exit_code": int(exit_code),
+        # Host-absolute path to the per-instance state directory so the
+        # marker is self-describing. The DELETE 410 / STATUS bodies
+        # echo this verbatim so a clew launcher (or human operator)
+        # can `cat <runtime_dir>/stderr.log` without recomputing the
+        # path from <name>.
+        "runtime_dir": str(runtime_dir.resolve()),
         "stdout_tail": _tail_text(stdout),
         "stderr_tail": _tail_text(stderr),
         "remediation_hint": hint,
--- a/src/scitex_agent_container/_lifecycle/_startup_failed.py
+++ b/src/scitex_agent_container/_lifecycle/_startup_failed.py
@@ -1,0 +1,240 @@
+"""STARTUP_FAILED lifecycle marker for stillborn agents.
+
+A stillborn agent is one whose container fails during creation —
+apptainer FATAL on a missing bind source is the canonical example,
+but the marker covers any failure between ``sac agents start <name>``
+returning and the SDK runner writing its first heartbeat. The
+clew-cohort-a-capsule-0201225 incident on 2026-06-02 was the
+motivating case:
+
+  * ``POST /agents`` returned HTTP 200 because the subprocess to
+    ``sac agents start`` exited 0 after the apptainer wrapper itself
+    handed off to the next stage,
+  * but the apptainer container creation FATAL'd with "mount source
+    /work/... doesn't exist",
+  * the SDK runner never started → no pidfile, no session.jsonl, no
+    heartbeat,
+  * ``GET /agents/<name>/status`` returned ``session_id: null``,
+    ``DELETE /agents/<name>`` returned 404 "no pid file", ``POST
+    /agents/<name>/send`` returned 400 "no live session".
+
+The host had silently lost a request and the operator had three
+different shaped failures to triage. This module adds a single
+on-disk marker (``runtime_dir/STARTUP_FAILED``) so:
+
+  * the marker IS the record of the stillborn lifecycle event,
+  * ``DELETE`` returns ``410 Gone`` (resource existed, has been removed)
+    with the failure detail in the body instead of ``404 Not Found``,
+  * ``STATUS`` reports ``status=startup_failed`` with the same body so
+    the operator and any orchestrator see a unified shape,
+  * future automation can drive an off-host ``forget`` of any agent
+    whose state-dir has only ``STARTUP_FAILED`` + no pidfile (= it
+    never started, never will, safe to GC).
+
+The on-disk shape is intentionally minimal JSON with a schema version
+so we can extend without breaking the existing readers:
+
+.. code-block:: json
+
+   {
+     "schema_version": 1,
+     "started_at":   "<ISO-8601 UTC>",
+     "failed_at":    "<ISO-8601 UTC>",
+     "phase":        "container_creation",
+     "kind":         "apptainer_mount_failed",
+     "exit_code":    255,
+     "stdout_tail":  "<last N lines>",
+     "stderr_tail":  "<last N lines>",
+     "remediation_hint": "..."
+   }
+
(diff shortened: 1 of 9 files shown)


ywatanabe1989:
[clew consumer review — wire-shape LGTM, one ask]

## LGTM (wire-shape locked from consumer side)

The five-value `kind` enum, the `details.unresolvable` array, the `schema_version` + atomic-write + 8 KiB-bounded tails on `STARTUP_FAILED`, and the `status: startup_failed` field on `GET /agents/<name>/status` (suppressed for healthy agents) all match what the clew launcher needs to display a single-pass diagnostic to the operator without round-tripping. Specifically:

- `details.unresolvable` array — clew can report all bind misses for one capsule in one operator-facing line (would otherwise need 4 POST-retry round-trips per capsule, × 49 capsules at the limit case).
- `kind` branch key + prose `error` for humans is the right split. `details.remediation_hint` is reusable by clew launcher verbatim.
- The `STARTUP_FAILED` schema (especially `phase` enum extensibility + `kind` catch-all `container_creation_unknown`) plus the DELETE 410 + status-field suppression pattern means clew launcher can detect stillborn agents via a single `if "startup_failed" in body:` branch.

## One ask — add `acl_deny` to the POST /agents `kind` enum in this PR

The preflight failure body currently shows `kind: "bind_unresolvable"` only. If a separate ACL handler returns 403 today (or in the near future) without the same body shape, clew launcher ends up with two branch styles — one parsing `kind` for bind / spec / already-exists failures, and a separate `if status == 403:` for ACL deny.

Concrete request: when the ACL handler rejects a `POST /agents` (or any of the introspection endpoints in PR-3), the body should look like:

```json
{
  "error": "caller 'proj-X' not authorized to spawn 'clew-cohort-a-...'",
  "kind": "acl_deny",
  "details": {
    "caller": "proj-X",
    "target": "clew-cohort-a-...",
    "reason": "lineage_mismatch | bearer_invalid | other"
  }
}
```

So the consumer side can keep a single `match body["kind"]:` switch covering all 5 kinds (`bind_unresolvable | spec_invalid | already_exists | acl_deny | startup_failed`). Same handler for clew's `--limit 49 --dry-run` validation pass as for live spawn.

If the ACL gate is being added in PR-3 instead, that's also fine — the consumer-side rewrite that calls `sac agents spawn-from-here` lives in the same clew follow-up PR anyway, so the wire shape just needs to be consistent at PR-3 merge time.

## Non-blocking — `details.unresolvable[*].mode`

Currently each entry has `bind` (full spec string), `host_resolved` (source path), `container_dest`, `exists_on_host`. A future failure mode worth shape-anticipating: `source` exists on host but is read-only when the spec asks for rw. Adding `mode: "ro" | "rw"` as a separate field now (or at `schema_version: 2`) would let that future failure mode reuse the same `unresolvable` array shape rather than carving a new kind. Pure shape extensibility — no implementation cost in this PR.

CI: ruff-on-ubuntu-latest currently shows FAILURE; pytest-matrix 3.11/3.12/3.13 in progress. Will swing back for final review once CI is green.

Thanks for the fast turnaround on the wire-shape iteration — this is the right PR to unblock the 49-capsule cohort-A rerun on the clew side.

PR #304, opened 2026-06-04
fix(beacon): suppress empty completion beacons at the sender

ywatanabe1989:
## Summary

A turn whose conversation never recorded an outcome (no `ResultMessage`, no exception caught) leaves `turn_context.status` at `None` (→ coerced to `STATUS_UNKNOWN`) and `turn_context.summary` at `""`. The prior emit path still dispatched that contentless beacon to the requester's inbox. Peers across the fleet were observing 35–55+ structurally-identical empty beacons per agent (this PR's own session counted 55+ before merge), degrading the a2a channel.

Sender-side guard inside `emit_completion_push`: if the resolved `status == STATUS_UNKNOWN` AND `summary` is empty / whitespace, **skip the wire dispatch** and log a **loud WARNING** that names agent + requester + dispatch_id. `turn_context.pushed = True` is set before returning so the failure-path's redundant emit in `_drive_turn.finally` cannot retry the same empty beacon. Loud-skip beats silent drop — a stuck emitter stays observable.

## Lead-directed scope assessment (auth-bypass angle, RULED OUT)

scitex-dev escalated the report from emitter-hygiene to potential bus-correctness because the same `dispatch_id` was observed on both legs of a round trip with opposing `from_agent`. Trace findings (already a2a'd to lead and acknowledged):

- **Round-trip correlation is by design.** A turn dispatched dev→clew rides `dispatch_id=X` with `from_agent=dev`; the completion beacon carries the same `dispatch_id=X` back to dev with `from_agent=<recipient agent>`. Both peers seeing the same id with opposing `from_agent` is the intended pattern — not a bus echo.
- **ACL boundary intact.** `_listen._node_channel.node_message_send` runs `_acl.check_send_acl` BEFORE persist/publish. With a per-node bearer the call REQUIRES `metadata.from_agent == authenticated_node`, else 403 "identity spoof". Only the host-wide bearer (administrative / cross-host forwarder) honors `from_agent` verbatim, by design.
- **No fan-out, no rewriter.** `broker.publish(target, event)` is keyed to a single target's subscribers; no code path re-publishes one inbound event to multiple targets, and nothing rewrites `from_agent` during routing.

**Severity: cosmetic mislabeling on contentless beacons, NOT an identity-swap / auth-bypass exposure.** Lead acknowledged this severity assessment and confirmed normal-track shipping. Single-fix branch.

If, after this PR + the deploy rebuild, the same-dispatch-id round-trip pattern is *still* misread as echo, the next look is receive-side persistence dedup in `_state.state_db_channel.persist_event` — out of scope here since the noise should be gone first.

## Tests (no mocks)

Added to `tests/scitex_agent_container/_runners/test__session_completion.py::TestEmptyBeaconSuppression` (7 new tests, real collaborators only):

- `status==unknown` + empty summary → **SUPPRESS** (no dispatch).
- `status==unknown` + whitespace-only summary → **SUPPRESS**.
- `status==unknown` + real summary → **STILL DISPATCH** (honest signal preserved).
- `status==success` + empty summary → **STILL DISPATCH** (clean turn that happened to produce no text is still signal).
- `status==failure` + empty summary → **STILL DISPATCH**.
- Suppression marks `turn_context.pushed=True` (idempotency).
- Suppression logs a WARNING containing `"SUPPRESSED"`.

All 28 tests in the affected file pass (existing 21 + 7 new). Full suite: 6277 passed, 38 skipped, 9 deselected. Two failures **reproduce on untouched develop** (verified with `git -C /work` against tip 016d39b):

- `tests/develop/test_audit.py::test_audit_all_clean` — pre-existing `--fakeroot` top-level junk file from prior PR #297 work + pre-existing `26_credentials-rotation.md` leaf-over-budget.
- `tests/integration/test_cli_startup_budget.py::test_cli_help_under_budget` — environmental timing (0.55–0.66s vs 0.50s budget). My edit is in `_runners/_session_hooks.py`, not on the CLI startup import path.

## Deploy note

This change ships inside the agent SIF, so **a SIF rebuild + agent restart is required to take effect**. Lead is coordinating that step.

## Test plan

- [x] Targeted unit tests for the suppression guard (real `TurnContext`, real `caplog`)
- [x] Full repo test suite green (modulo two pre-existing failures reproduced on develop tip)
- [x] Trace audit of the listen path for identity-swap / fan-out (no findings)

### DIFF
--- a/src/scitex_agent_container/_runners/_session_hooks.py
+++ b/src/scitex_agent_container/_runners/_session_hooks.py
@@ -117,6 +117,18 @@ async def emit_completion_push(
     outcome — should not happen, since Stop comes after the ResultMessage)
     is reported as ``unknown``, NEVER fabricated as success.
 
+    Empty-beacon guard (fleet-wide noise suppression)
+    -------------------------------------------------
+    A beacon whose ``status`` is :data:`STATUS_UNKNOWN` AND whose summary is
+    empty / whitespace carries no information for the requester — the
+    conversation never recorded an outcome and there is no reply text to
+    relay. Peers across the fleet were seeing dozens of these structurally-
+    identical empty beacons per agent, degrading the a2a channel. We
+    SUPPRESS the dispatch at the sender (loud-skip beats silent noise): the
+    skip is logged at WARNING with the agent / requester / dispatch_id so a
+    stuck emitter is observable in the agent log, and ``pushed`` is set so
+    the failure-path's redundant emit cannot retry the same empty beacon.
+
     A delivery failure is LOUD (logged at WARNING, requester named) but
     never re-raised: the turn already completed, and a flaky receipt must
     not crash the agent or abort the SDK control flow.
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Superseded by PR #309 (merged 2026-06-05). Same guard, slightly different rule (#309 suppresses on `dispatch_id is None AND summary empty`; this PR's draft used `status==UNKNOWN AND summary empty`). Net effect on the fleet noise is identical for the dispatch_id=null case. Opening a small follow-up to strengthen #309's guard to also catch `dispatch_id set, status==unknown, summary empty` per lead msg 62b0f613104048c6adae11c73d9f8b63.

PR #318, opened 2026-06-06
feat(sdk_common): suppress Computer tool USE when provider backend active (defense-in-depth)

ywatanabe1989:
## Summary

Provider-backed agents (LiteLLM / vLLM / gateway via `spec.claude.provider`) route their SDK session through an Anthropic-API shim. clew bm172 cohort 2026-06-06: every API call 422'd on `AnthropicComputerTool.display_width_px` missing — LiteLLM 1.52.16's Anthropic shim drops that field on translation and pydantic-rejects the tool body. Every agentic turn errored; the capsule ran 60 max_turns of errors before naturally exiting.

This PR was originally intended as a schema-suppression fix.

## Empirical proof (v7 bind-test, RED)

clew's v7 ran this exact branch via bind-editable-over-SIF (commit f0c8e36 at the time, now squashed into 74f4546). spec.claude.provider gate fired correctly. Result: still 422 every turn. Assistant body sample:

```
API Error: 422 {"detail":[{"type":"missing","loc":["body","tools",9,"AnthropicComputerTool","display_width_px"],"msg":"Field required",...}]}
```

The `loc: [..., "tools", 9, "AnthropicComputerTool"]` proves the Computer tool schema is STILL in `tools[]` at position 9 of the outbound request body — LiteLLM's pydantic is parsing it and 422-ing on the missing required field. So `ClaudeAgentOptions.disallowed_tools` blocks USE only, NOT outbound schema emission.

The SDK docstring at `claude_agent_sdk/types.py:1666-1671` ("These tools are removed from the model's context and cannot be used") is empirically misleading on the "removed from the model's context" claim. Filed upstream for docstring fix.

## Scope-amended: merging as defense-in-depth

This PR merges on its own independent merits per lead msg 017fb4ed:

- When a provider backend is active, the agent CANNOT choose Computer as the next tool to invoke even if the schema ships. Correct tool-usage hygiene.
- Future-proofs against agents who would otherwise pick Computer on a non-Anthropic backend that DOES accept the schema but cannot execute the tool.
- Does NOT fix the LiteLLM 422 — that fix lives in the LiteLLM shim layer (clew patches the AnthropicComputerTool pydantic model to make `display_width_px` optional).

## Fix

`runtimes/_sdk_common.py::build_sdk_options` already had an `allowed_tools` merge for the subagent (Agent) tool. Add a sibling `disallowed_tools` merge gated on the OR of two independent provider-active signals:

- `provider_active(cfg)` — True iff `spec.claude.provider.base_url` is non-empty. The canonical sac-managed path.
- `os.environ.get("ANTHROPIC_BASE_URL")` non-empty at SDK invocation time — catches every other declaration shape: apptainer raw_args env-injection, `spec.claude.environment`, inherited host shell. The robust universal shape because `build_sdk_options` runs INSIDE the capsule SIF where the env IS whatever the runtime wired.

Empty-string defence: a stray `--env ANTHROPIC_BASE_URL=` (operator typo) is treated as unset via `.strip()` truthiness.

The merge appends from a module-level tuple `_ANTHROPIC_SHIM_INCOMPATIBLE_TOOLS = ("Computer",)` — adding a future shim-incompatible tool is a one-line edit that propagates to every provider-backed agent.

## Back-compat

- **Non-provider agents** (both signals False): unchanged.
- **Caller-supplied `disallowed_tools`** (via `extra={}`): PRESERVED — merge, not replace. `["WebFetch"]` becomes `["WebFetch", "Computer"]` on a provider-backed agent.

## Tradeoff

An operator who EXPLICITLY wants Computer on a provider (e.g. a future LiteLLM that DOES relay `display_width_px` and the operator wants the agent to actually invoke computer-use) has no clean override path in this PR. Zero current users; future opt-out knob (`spec.claude.provider.shim_relays_computer_use=true`) can flesh out the override story if/when needed. Documented in the constant's docstring.

## Changes (squashed into one commit)

- **`src/scitex_agent_container/runtimes/_sdk_common.py`**
  - `_ANTHROPIC_SHIM_INCOMPATIBLE_TOOLS: tuple[str, ...] = ("Computer",)` + docstring (SDK doc cite + bm172 cohort cite + v7 RED finding).
  - `_load_agent_config_silent(agent_name)` helper (mirrors `resolve_agent_workspace`'s registry-read shape; best-effort failure profile).
  - `build_sdk_options`: after the existing `allowed_tools` merge, OR the two gates and merge the constant into `kwargs["disallowed_tools"]`.

- **`tests/scitex_agent_container/runtimes/test__sdk_common.py`** — `+TestProviderAwareToolExclude` (+7 tests, no `MagicMock`):
  - 4 spec-side: provider-active adds Computer; no-provider doesn't; caller-supplied preserved; caller-supplied merged.
  - 3 env-side: `ANTHROPIC_BASE_URL` adds Computer; neither-on doesn't; empty-string env doesn't trigger gate.

## Test plan

- [x] Local: 7/7 new + 41/41 existing _sdk_common = 48/48 pass.
- [x] CI on the same gate as PRs #307-#317.

## Related work

- **SDK `disallowed_tools` docstring inaccuracy** — filed upstream / tracked separately for ecosystem feedback.
- **Real 422 fix** — clew patches the LiteLLM shim's `AnthropicComputerTool` pydantic model to make `display_width_px` optional.
- **Task #16**: `spec.claude.flags` propagation gap — separate sac bug, follow-up PR.
- **Task #18**: empty `tools` array on title-gen → LiteLLM 400 — separate quirk, follow-up.

### DIFF
--- a/src/scitex_agent_container/runtimes/_sdk_common.py
+++ b/src/scitex_agent_container/runtimes/_sdk_common.py
@@ -56,6 +56,27 @@ __all__ = [
     "project_runtime_root",
 ]
 
+# Tools whose schema the LiteLLM (and similar) Anthropic-API shims
+# cannot relay to non-Anthropic backends. When ``spec.claude.provider``
+# declares a provider override (LiteLLM / vLLM / gateway), the Claude
+# Code SDK still registers these tools by default and embeds their full
+# schema in every outbound API request — the shim then chokes on a
+# missing required field and returns 422 on EVERY turn.
+#
+# Concrete case (clew bm172 cohort 2026-06-06, lead msg 5d6fce21):
+# AnthropicComputerTool requires ``display_width_px``; LiteLLM 1.52.16's
+# Anthropic shim drops it on translation → 422 on every API call → the
+# capsule ran the full 60-turn autonomous loop emitting only errors,
+# then naturally exited at max_turns.
+#
+# The SDK's ``disallowed_tools`` field is the right knob:
+# ``"These tools are removed from the model's context and cannot be
+# used"`` (claude_agent_sdk/types.py:1666-1671) — i.e. the schema is
+# never sent to the API. Extend this tuple if other shim-incompatible
+# tools surface; new entries propagate to every provider-backed agent
+# without further wiring.
+_ANTHROPIC_SHIM_INCOMPATIBLE_TOOLS: tuple[str, ...] = ("Computer",)
+
 
 def project_runtime_root(config: "AgentConfig") -> "Path | None":
     """If the agent's YAML lives under a project-scope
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Closing per lead msg 8b98f3ac post L2-GREEN: PR #319's provider.allowed_tools whitelist would suppress Computer at REGISTRATION (not just block USE) which is the broader correct shape, and the actual 422 fix in production was clew's LiteLLM schema-patch + qwen3_xml parser — Computer-USE-suppression here is redundant. PR #319 stays open pending a cheap isolated bind-test (LiteLLM patch reverted + #319 active) during 49-array prep to verify tools= actually de-registers.

PR #323, opened 2026-06-06
fix(sac-hardening): worktree-prune safety predicates + .claude*-path guidance

ywatanabe1989:
## Summary

Two-layer defence against the data-loss class proj-paper-scitex-clew flagged in their capsule-0220918 cohort-A incident (2026-06-06): a worktree at `/work/.claude-worktrees/<branch>/` with uncommitted changes + a running pytest was reaped by the Claude Code SDK harness on an "agent main-loop idle = stale" heuristic. Recovery only happened by lucky `cp → /tmp`.

Lead's host-side investigation confirmed the reaper is the **harness itself** (`~/.claude/file-history` GC + `~/.claude/projects/...-claude-worktrees-agent-...` session cleanup) — unpatchable from this surface. So this PR lands the two things we CAN do here.

### 1. Defence-in-depth — `sac agents prune-claude` safety predicates

Hardens the in-package slash-variant pruner (`cli_pkg/_agent_prune_claude.py::_plan_worktrees`) so even if the SAME anti-pattern bites `.claude/worktrees/agent-*` next, work is preserved. Adds lead's five skip-predicates in priority order, BEFORE the existing merged-into-base check:

| # | Predicate | Mechanism |
|---|---|---|
| 1 | uncommitted changes | `git diff --quiet` / `git diff --cached --quiet` non-empty → skip |
| 2 | unpushed commits | `git log @{u}..HEAD` non-empty → skip |
| 3 | process holding fd | `lsof +D <dir>` reports any fd → skip |
| 4 | branch not merged | existing check, now predicate #4 |
| 5 | idle-time = LAST tiebreaker | no idle check today; won't gain one without explicit operator ask |

Defensive failure modes: git failure on diff/log → treat as "has uncommitted/unpushed" (data-safer direction); `lsof` absent → predicate falls through (it's defence-in-depth, not the primary safety bar; predicates #1/#2/#4 still hold).

### 2. THE REAL FIX — path placement guidance

New skill leaf `_skills/scitex-agent-container/31_worktree-path-safety.md` + a warning in the to_home boot brief template (`examples/agents/full-agent/to_home/CLAUDE.md`) tell agents to **never** create worktrees under any `.claude*`-prefixed path. The harness reaps that namespace; using a neutral path like `<workdir>/worktrees/<slug>/` or `/tmp/<agent>-worktrees/<slug>/` sidesteps the reaper entirely. The predicate-hardening above only covers `_agent_prune_claude.py`, NOT the harness itself.

## Test plan

- [x] 23 prune-claude tests pass locally (`/opt/venv-sac/bin/python -m pytest tests/scitex_agent_container/cli_pkg/test__agent_prune_claude.py`), including five new ones for the three new predicates (uncommitted unstaged, uncommitted staged, unpushed, fd-held, plus a skip-reason assertion).
- [x] `tests/integration/test_templates_v3_valid.py` still passes after the to_home boot-brief edit.
- [x] `scitex_dev` audit-skills returns SUCC against the new skills dir (SKILL.md = 99 lines / 6140 B, under the SK-301 6144 B / 120 line budget; new leaf 95 lines / 4377 B under SK-401).
- [ ] CI matrix green on this PR (this PR does NOT block-ship the SIF rebuild — that's #322's lane).

## Notes

- `lsof` 4.95.0 in the SIF exits 1 even on successful matches when it hits any unreadable /proc entry on the walk. The helper parses stdout regardless of returncode and gates on a real data line below the header.
- The unpushed-commits test sets up upstream tracking via direct `git config branch.<name>.{remote,merge}` writes (bypassing `git branch --set-upstream-to`'s remote-fetch validation, which requires a real reachable remote — overkill for an offline test repo).
- Aside: this PR's own worktree was reaped by the harness mid-edit — recovered via the same `cp → /tmp` dance clew used. More empirical evidence that the `.claude*`-path rule is load-bearing; the predicate-hardening alone wouldn't have saved this branch's edits.

### DIFF
--- a/examples/agents/full-agent/to_home/CLAUDE.md
+++ b/examples/agents/full-agent/to_home/CLAUDE.md
@@ -14,6 +14,28 @@ This agent is a general-purpose worker.
 - Follow instructions precisely.
 - Commit and push changes when asked.
 
+## Worktree path safety — never under `.claude*/`
+
+When you create a git worktree, the path **MUST NOT** start with
+`.claude` (no `.claude/worktrees/`, no `.claude-worktrees/`, no
+`.claude-*` sibling). The Claude Code SDK runtime owns the `.claude*`
+namespace and reaps it on an internal idle-time / GC heuristic —
+uncommitted work there is silently lost (proj-paper-scitex-clew
+capsule-0220918, 2026-06-06).
+
+Use a NEUTRAL path:
+
+```bash
+# RIGHT — outside .claude*, safe from harness reaping
+git -C /work worktree add -b fix/my-thing /work/worktrees/fix-my-thing origin/develop
+# Also fine for scratch:
+git -C /work worktree add -b fix/my-thing /tmp/$(whoami)-worktrees/fix-my-thing origin/develop
+```
+
+Full doctrine + recovery if you find a worktree already under
+`.claude*/`:
+`~/.claude/skills/scitex/scitex-agent-container/31_worktree-path-safety.md`.
+
 ## Responsiveness — short turns, background heavy work
 
 The operator's Telegram message reaches this agent via the same inbox
--- a/src/scitex_agent_container/_skills/scitex-agent-container/31_worktree-path-safety.md
+++ b/src/scitex_agent_container/_skills/scitex-agent-container/31_worktree-path-safety.md
@@ -1,0 +1,95 @@
+---
+description: |
+  [TOPIC] Where agent git-worktrees must (and must NOT) live to avoid silent reaping by the Claude Code harness.
+  [DETAILS] The Claude Code SDK runtime auto-manages files under any `.claude*`-prefixed path (file-history GC + session-dir cleanup). Worktrees created at `<workdir>/.claude/worktrees/`, `<workdir>/.claude-worktrees/`, or any other `.claude*` path are visible to that machinery and can be reaped on an idle-time heuristic — losing uncommitted/unpushed work. Use a NEUTRAL path instead (`<workdir>/worktrees/`, `/tmp/<agent>-worktrees/`).
+tags: [scitex-agent-container-worktree-path-safety]
+---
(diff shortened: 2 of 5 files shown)


ywatanabe1989:
**New signal for the worktree-reaper predicate work (from PR #325 incident):**

While working on #325 (telegram-wake diagnostics) I had a worktree reaped mid-edit AGAIN at `/work/worktrees/fix-telegram-wake/`. That's a NEW failure mode — the path is OUTSIDE any `.claude*` namespace (which #323's path-safety leaf says is sufficient to sidestep the harness reaper). Yet the working tree's `.git` gitfile pointer at `/work/.git/worktrees/<name>/` was wiped between turns, leaving the working dir as an orphan (recovered via cp→/tmp + worktree re-create).

Possible signal: the reaper isn't keying purely on the working-tree path (`.claude*` vs neutral) — it may also walk the parent repo's `.git/worktrees/<name>/` metadata directory and clear entries from THERE on a separate heuristic, leaving the working-tree dir intact but orphaned. If so, the path-placement rule in `31_worktree-path-safety.md` is insufficient; the reaper has a metadata-walk path independent of the working-tree-path predicate.

Per lead's routing: filing this as a follow-up comment so the predicate work in #323 (5 skip-predicates already merged) can be extended in a follow-up if anyone has metadata-walk evidence. NOT blocking #323 itself.

Concrete clue: my worktree at `/work/worktrees/fix-telegram-wake/` survived as a working dir, but `/work/.git/worktrees/fix-telegram-wake/` (the gitfile-pointer metadata dir) was gone — `git status` from the worktree returned `fatal: not a git repository: /work/.git/worktrees/fix-telegram-wake`. Same shape as the PR #323 commit incident, which I'd attributed to the standard `.claude*` reap but in retrospect was probably this same metadata-walk path.

PR #337, opened 2026-06-07
fix(runtime): redirect $TMPDIR to a per-agent /opt/tmp host bind (#50 option 4)

ywatanabe1989:
## Summary

Closes operator-blocking bug #50 — capsule `/tmp` overflow under apptainer's `--writable-tmpfs --containall`, blocking proj-paper-scitex-clew cohort-A (stalled at 19/47 capsules) and downstream the operator's paper deliverable.

**Root cause:** heavy ML workloads (TF/Keras tokenizer caches, BERT preprocessing, multi-GB pickles) write via `tempfile.gettempdir()`, which honours `$TMPDIR` before falling through to `/tmp`. With `$TMPDIR` unset, those writes land on either the small writable-tmpfs overlay or the 64M `--containall` session tmpfs and the process stalls on ENOSPC.

**Fix (lead-decision Option 4, a2a `7d14d69b`):** auto-inject `TMPDIR=/opt/tmp` into the container env, and bind a per-agent host scratch dir at `/opt/tmp` inside the container so `$TMPDIR` writes hit real disk regardless of overlay choice. Proved live by proj-paper-scitex-clew on capsule-0238624 (`submission.json` produced T+7min after the override landed in their PR #112).

### Implementation

- **New helper** `opt_tmp_flags(config, state_dir)` in `runtimes/_apptainer_tmpfs.py` — emits `--bind <state_dir>/opt-tmp:/opt/tmp --env TMPDIR=/opt/tmp` under the default `ApptainerSpec`.
- **Wired** into `runtimes/_apptainer_runtime.build_run_argv` right after `tmpfs_workdir_flags` — the two defences compose (workdir relocates `/tmp` + `/var/tmp` to host disk; `opt_tmp` redirects `\$TMPDIR` to a separate per-agent bind).
- **Operator escape hatches** respected (never silently overridden):
  - `spec.apptainer.tmpfs_size=""` opts out of BOTH measures (consistent with existing `tmpfs_workdir_flags`)
  - `spec.apptainer.env.TMPDIR=<any>` skips the env injection
  - `spec.apptainer.binds=[<host>:/opt/tmp]` skips the bind (substring match handles plain + `:rw`/`:ro` mode suffix)
- **No new spec fields. No `.def` edits.** `tmpfs_size` default (`"2G"`) unchanged.

### Why Option 4 over the earlier Option 2 (bind /tmp:/tmp fallback)

Option 2 (`--bind /tmp:/tmp` when state_dir filesystem overflows) was rejected mid-flight by lead in favor of Option 4 because:
- No host-`/tmp` leak (per-agent scratch dir, not the host's global `/tmp`)
- Proved live in production by clew on capsule-0238624 (Option 2 was theoretical)
- Smaller surface (~30 LOC vs. ~50 LOC fallback logic)

## Test plan

- [x] 9 new unit tests on `opt_tmp_flags()` covering: default emission, scratch-dir creation, opt-out via `tmpfs_size=""`, no `apptainer` block, `TMPDIR` operator override (env-only skip + bind still emitted), bind operator override (bind-only skip + env still emitted, plain + `:rw` mode), combined opt-out
- [x] 3 new integration tests on `build_run_argv` asserting `--bind .../opt-tmp:/opt/tmp` + `--env TMPDIR=/opt/tmp` appear in the assembled argv plus the operator-owns-bind dedup
- [x] All 16 targeted tests green locally
- [x] Full suite green: 6506 passed / 28 skipped (excluding the pre-existing `tests/develop/test_audit.py::test_audit_all_clean` failure already being addressed by the in-flight `fix/audit-clean-on-develop` branch — NOT introduced by this PR)
- [ ] CI green
- [x] proj-paper-scitex-clew notified (a2a `94d8eb75`) — they will retire their `_emit.py` spec-side TMPDIR override once this lands

## Cross-references

- Issue #50
- Lead a2a `7d14d69b` (Option 4 confirmation)
- Lead a2a `0c4446c1` (Option 4 re-confirmation after crossed messages)
- proj-paper-scitex-clew a2a `d346a786` (production proof + capsule-0238624)
- proj-paper-scitex-clew PR #112 (their spec-side workaround — removable on merge)
- Investigation report at `.tmp-audit/issue-50-investigation.md`

### DIFF
--- a/src/scitex_agent_container/runtimes/_apptainer_runtime.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_runtime.py
@@ -272,9 +272,18 @@ class ApptainerContainerRuntime(RuntimeBase):
         # when the operator already declared --workdir in raw_args. The
         # flag is curated (emitted before raw_args) so an operator's own
         # --workdir still wins via the helper's raw_args skip.
-        from ._apptainer_tmpfs import tmpfs_workdir_flags
+        from ._apptainer_tmpfs import opt_tmp_flags, tmpfs_workdir_flags
 
         argv += tmpfs_workdir_flags(config, state_dir)
+
+        # Second tmpfs-overflow defence: bind a per-agent host scratch
+        # dir at /opt/tmp inside the container and redirect $TMPDIR
+        # there. Stops heavy ML capsules (TF/Keras tokenizer caches,
+        # BERT preprocessing) from wedging on the small writable-tmpfs
+        # overlay when they use ``tempfile.gettempdir()`` instead of
+        # literal /tmp. Per #50 option 4 (lead a2a 7d14d69b); confirmed
+        # live on proj-paper-scitex-clew capsule-0238624.
+        argv += opt_tmp_flags(config, state_dir)
 
         # Anthropic-auth argv — emits the backend wiring (env + creds
         # bind). Branches internally on whether spec.claude.provider is
--- a/src/scitex_agent_container/runtimes/_apptainer_tmpfs.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_tmpfs.py
@@ -1,4 +1,4 @@
-"""Sized ``/tmp`` scratch for the apptainer runtime.
+"""Sized ``/tmp`` scratch + ``TMPDIR=/opt/tmp`` redirect for apptainer.
 
 Why this exists
 ---------------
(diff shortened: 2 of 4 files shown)


ywatanabe1989:
Closing — SUPERSEDED by a better fix that is already in develop and already live in production.

This PR proposed `opt_tmp_flags()`: bind a host scratch dir at `/opt/tmp` and inject `TMPDIR=/opt/tmp`. develop instead ships `tmpfs_workdir_flags()` in the same module (`runtimes/_apptainer_tmpfs.py`), which uses apptainer's `-W/--workdir` to relocate the in-container **`/tmp` and `/var/tmp` themselves** onto host disk, plus a fail-loud free-space precheck (`TmpfsSpaceError`) so a shortfall surfaces at launch instead of as a mid-run ENOSPC.

**The landed fix is strictly more general.** `TMPDIR=/opt/tmp` only helps code that honours `$TMPDIR` (i.e. goes through `tempfile.gettempdir()`). Anything that writes to a hardcoded `/tmp` — and plenty does — would still hit the 64 MB `--containall` session tmpfs and still die. Relocating the mount fixes both classes at once. Same root cause (#50), addressed one level deeper.

**Verified live, from inside a running agent container just now** — not inferred from code:

```
$ df -h /tmp /var/tmp
/dev/sdd   2.0T  1.7T  141G  93%  /tmp
/dev/sdd   2.0T  1.7T  141G  93%  /var/tmp
```

`/tmp` is a 2.0T host filesystem with 141G free — **not** the 64 MB tmpfs that caused #50. The `--workdir` relocation is active and working. Meanwhile `/opt/tmp` does not exist in the container and `$TMPDIR` is not `/opt/tmp`, confirming this PR's change never landed by any route — and does not need to.

State of this PR for the record: OPEN 37 days, CONFLICTING, red on all three pythons, auto-merge never armed. Nothing here is worth reviving; the bug it targeted is fixed.

The card `sac-tmpfs-overflow-50` claimed "Option 4 implementation landed in PR #337". That was false — it never landed. I am correcting the card to record what actually shipped.

PR #347, opened 2026-06-09
fix(hooks): block 体言止め (サ変名詞止め) sentence endings in TG structured-format hook

ywatanabe1989:
## Summary

Rule 3 of `enforce_telegram_structured_format.sh` only caught lines closing with 「。」. Three real crew Telegram messages slipped past it because they ended on a サ変動詞名詞 with no trailing 「。」 (体言止め in the noun-stem form):

- 「spartan_preamble.sh, spartan_one_capsule.sh の scratch default 除去」
- 「spartan_submit_array.sh 新設」
- 「background subagent 起動」

This adds rule 3b to the same hook: when a line carries Japanese characters, has content before the final noun, and the stem ends with one of the canonical サ変名詞 (除去 / 阻止 / 新設 / 起動 / 進行 / 完了 / 確認 / 実装 / 作成 / 削除 / 報告 / 更新 / 登録 / 設定 / 取得 / 対応 / 調査 / 検証 / 修正 / 追加 / 開始 / 終了 / 抽出 / 移行 / 統合), it is a 体言止め sentence and must be rewritten as 〜します / 〜しました.

## Guards (so the existing self-test stays green)

- ASCII-only lines (e.g. `- config/SPARTAN.yaml`) are skipped — they have no Japanese, so they are file references, not 体言止め sentences.
- Lines ending with a known source/config file suffix (`.sh` / `.py` / `.yaml` / `.yml` / `.toml` / `.md` / `.json` / `.txt` / `.cfg` / `.ini` / `.lock` / `.conf`) are skipped even if they share a noun string.
- Bare header labels (`1. 完了`, `2. 残`) where the noun IS the whole post-list-marker content stay allowed — only sentences with content BEFORE the noun trip the rule.
- Lines already ending with 「。」 are deferred to rule 3 to avoid double-fire.

## Base branch note

Branch is rooted on `fix/telegram-format-hooks-ext` because `develop` does not yet contain the `_baseline_assets/telegram_hooks/` directory at all (that PR introduces the directory + the sibling hook scripts + the pytest wrapper). The operator instruction was "off develop", but rooting off develop would require duplicating the entire telegram_hooks infrastructure here. Recommended merge order: land `fix/telegram-format-hooks-ext` first, then this PR rebases cleanly.

## Test plan

- [x] Added 4 new self-test cases: 3 EXPECTED-BLOCK for the operator-escalation crew lines, 1 EXPECTED-ALLOW for `- config/SPARTAN.yaml`.
- [x] Confirmed RED before implementation (3 new BLOCK cases initially failed with got=0 want=2).
- [x] Confirmed GREEN after implementation: 25/25 shell self-test cases pass (was 21/21 before, no regression on the original 21).
- [x] Registered `enforce_telegram_structured_format.sh` in the existing pytest wrapper (`test_telegram_hooks_self_tests.py`); 12/12 pytest cases pass.
- [x] Hook script is executable (`chmod +x`) and registered in the `test_hook_script_exists_and_is_executable` parametrize list.

### DIFF
--- a/src/scitex_agent_container/_baseline_assets/telegram_hooks/enforce_telegram_structured_format.sh
+++ b/src/scitex_agent_container/_baseline_assets/telegram_hooks/enforce_telegram_structured_format.sh
@@ -1,0 +1,400 @@
+#!/bin/bash
+# -*- coding: utf-8 -*-
+# Timestamp: "2026-06-09 (OP-PRIO-FMT rule 6 / structured-format)"
+# File: ~/.claude/hooks/pre-tool-use/enforce_telegram_structured_format.sh
+#
+# OP-PRIO-FMT rule 6 (operator 2026-06-09 — structured-format directive,
+# extended by lead 2026-06-09 amendment for です/ます tone + dangling-
+# prose ban): a Telegram message MUST be structured when it carries
+# 3+ lines of content. Four failure modes are blocked here; the
+# fifth (blank-line spacing between top-level items) is already
+# enforced by the companion ``telegram_line_spacing.sh`` and
+# intentionally NOT duplicated.
+#
+# Rule 1 — TOP-LEVEL NUMBERING
+#   Top-level items MUST be numbered (1./2./3. or 1a./1b. style).
+#   If a message has 3+ top-level hyphen-bullets and ZERO numbered
+#   top-level items, the operator cannot quote by index on a phone
+#   and is forced to re-read. BLOCK with a nudge to renumber.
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Closing per operator: this hook is a user-private hook and should not live in the public repo. The tightened version is already deployed at ~/.claude/hooks/pre-tool-use/enforce_telegram_structured_format.sh (self-test 25/25 green) and is in active use.

PR #435, opened 2026-06-20
fix(apptainer): fail loud when credentials placeholder cannot reach container $HOME

ywatanabe1989:
## Context — handoff item 2 (scitex-arm capsule experiment, 2026-06-20)

Old ADR-0007 specs (`relaxed: true` + raw_args `--home /home/agent` /
`--overlay <dir>` / `--userns`) FATAL at boot under current SAC with the
cryptic:

```
FATAL: ... mount /home/agent/.claude/.credentials.json ...
destination doesn't exist in container
```

apptainer FILE binds require the in-container destination to pre-exist. SAC
pre-creates an empty placeholder at the host path backing the container
`$HOME`, but a raw_args `--home` mounts a **fresh tmpfs** at the container
HOME that shadows the workspace-home bind. The only host path that survives
that shadow is the overlay **upper-home** (directory overlays only). With an
`.img` (loopback) overlay or no overlay, there is no upper-home bind → the
placeholder is lost → FATAL.

## What this PR does (priority #1 — highest value, lowest risk)

**Fail loud before the cryptic FATAL.** New
`runtimes/_apptainer_creds_guard.assert_credentials_placeholder_delivers`,
called from `build_run_argv` right before the credentials `:rw` bind is
appended. It raises a precise `CredentialPlaceholderUndeliverableError`
naming the missing placeholder path + the fix, when:

- a creds bind will be emitted AND the spec declares a raw_args `--home`
  (tmpfs shadow) with **no usable overlay upper-home directory**, or
- the placeholder could not be created at all (`None` despite a bind —
  the previously-swallowed `OSError` path).

It is a **pure no-op for every layout that already delivers** (managed /
directory overlay, non-relaxed workspace-home bind, provider / no-credential
specs), so no working launch changes behaviour.

Also documents the FATAL + the raw_args→managed-overlay migration in skill
`40_troubleshooting.md`.

## Priority #2 (migration) — already works on develop; documented, not re-coded

The recommended target — `spec.apptainer.overlay` (managed) — **already**
emits `--home /home/agent` (via `compute_iso_prepend`) and places the
credential placeholder in `<overlay>/upper/home/agent/.claude/.credentials.json`
(verified end-to-end). So #2 is a **spec rewrite** (raw_args → managed
overlay; keep `--userns` only for rootless gh/git), not a SAC code change.
The migration recipe is in the troubleshooting skill.

## Priority #3 (overlay reset/prune) — DEFERRED with a plan

`rm -rf` inside a live directory-overlay `upper/` layer (stale
`work-ci-runner-*` / `uvwork/*`) is the riskiest item: destructive,
needs a hold/lifecycle check (registry is shaky for never-started agents),
and must respect fuse-overlayfs whiteout/opaque semantics. Per the
"stop if too risky to do correctly" guidance, deferred rather than guessed.
Plan: a `sac overlay prune <agent>` that (1) refuses if the overlay is held
by a running agent, (2) prunes only an allowlisted set of cache subdirs
under `<overlay>/upper/<home>/` (never `.claude`/credentials), (3) is
`--dry-run` by default and prints freed bytes — modeled on
`_apptainer_build._create_overlay_image`'s subprocess+fail-loud shape.

## Tests

7 new guard tests (no mocks; real `AgentConfig`/`ApptainerSpec` on tmp
dirs; AAA markers; one-assert). Full `tests/.../runtimes/` suite:
**957 passed, 14 skipped** (skips need a real SIF/apptainer), 0 failures.


ywatanabe1989:

PR #436, opened 2026-06-20
fix(channels): register TUI channel MCP into $HOME/.claude.json (user scope)

ywatanabe1989:
## Problem (handoff item 3, FROM_PAPER_SCITEX_CLEW)

A spec with `channels: [server:sac]` boots into the TUI showing
`server:sac · no MCP server configured with that name` (confirmed in a live
pane: proj-scitex-dev). a2a/channel comms never wire. The SDK path
auto-registers the sac channel MCP; the TUI path did not effectively.

## Root cause (verified against the SDK-bundled `claude` v2.1.150 binary)

`_apptainer_inner_argv.tui_channel_config` already passes the `sac mcp
channel` subscriber on an inline `--mcp-config`. But claude's dev-channels
resolver (`--dangerously-load-development-channels`) only scans MCP servers
from the **enterprise / user / project / local SETTINGS scopes** (binary
`t2(scope).servers`):

- `project` — `.mcp.json` files walking up from cwd (`--pwd`)
- `user` — top-level `mcpServers` of `$HOME/.claude.json`
- `local` — `$HOME/.claude.json`'s `projects[<cwd>].mcpServers`

Servers passed via `--mcp-config` (the inline sac subscriber **and** the
`--mcp-config $HOME/.mcp.json` file) are stored separately (`K.mcpConfig`)
and are **never** in the scope set the resolver checks → the channel is
silently dropped.

The `server:` flag prefix is **correct**: the binary parser keys on it
(`server:<name>` → `{kind:server, name:<name>}`; an untagged entry errors
`entries must be tagged`). So the earlier #427 bare-name strip was wrong, and
the #428 prefix-kept flag is right — the bug is **registration scope**, not
the flag value.

## Fix

New `runtimes/_tui_channel_mcp.py` writes each `spec.claude.channels`
member's backing MCP into the **top-level `mcpServers`** of the agent's OWN
`$HOME/.claude.json` (claude's `user` scope — never an operator project dir):

- `server:sac` → the `sac mcp channel` subscriber, synthesised from the SAME
  shared `tui_channel_plan` the inline `--mcp-config` uses (no drift).
- any other `server:<name>` → its backing entry from `$HOME/.mcp.json`
  (deployed from `to_home/.mcp.json`) or `spec.mcp_servers`; **fail-loud**
  (`ChannelMcpMissingError`) when absent — a declared channel with no MCP
  must stop the start, not boot a mute agent (no silent fallback).

Wired into `TuiSessionRuntime.materialize_workspace` over the SAME home pair
`ensure_project_onboarding` writes (workspace-home + overlay upper-home), so
it lands for relaxed `--home`/directory-overlay specs (e.g. proj-scitex-dev)
too. This is the TUI mirror of the SDK `apply_channels` auto-register.
Idempotent — an operator's existing same-name entry wins.

## Tests / quality

- 14 new tests in `tests/.../runtimes/test__tui_channel_mcp.py` (real
  `AgentConfig` via `load_config` + real `tmp_path` files; no mocks; AAA
  markers).
- Full runtimes suite green: **964 passed, 14 skipped**.
- New module 265 lines; ruff clean.


ywatanabe1989:

PR #437, opened 2026-06-20
feat(to_home): aggregate resolved skills into CLAUDE.md at deploy time

ywatanabe1989:
## Summary

Implements the operator's "guarantee" (paper-scitex-clew handoff item 5):
because `@`-import and skill-chaining are unreliable triggers, sac now
**forcefully inlines an agent's resolved skill content into its CLAUDE.md at
deploy time**, between idempotent markers, so the skills are always in
context (CLAUDE.md is loaded every turn).

## What changed

- **`runtimes/_to_home_skills.py`** (new) — `aggregate_skills_into_claudemd(workspace_home)`:
  - Reads the agent's now-materialized `*.md` skills under
    `<workspace_home>/.claude/skills/`. (In v3 `spec.skills` is removed and
    skills are *materialized* into the workspace home via `to_home`, not
    bind-mounted — so post-materialization these files ARE the agent's
    resolved skills.)
  - Inlines them into CLAUDE.md wrapped in
    `<!-- sac:skills:start -->` … `<!-- sac:skills:end -->`.
  - **Forceful + idempotent**: each deploy strips any existing block and
    re-renders. This stays correct even though CLAUDE.md is *also*
    marker-protected by `_deploy_marker_protected` (a separate Start/End
    generated-section pair): strip-then-append keeps exactly one skills block
    regardless of how the user-tail carries it forward across re-deploys.
  - No skills → no-op (but a stale block in an existing CLAUDE.md is still
    cleaned up, so removing the last skill self-cleans). Skills present but no
    CLAUDE.md → the canonical root `CLAUDE.md` is created. Targets both
    `CLAUDE.md` and `.claude/CLAUDE.md` when present.
- **`runtimes/_to_home.py`** — `deploy_to_home` and `materialize_to_home`
  call the aggregator after to_home materialization + `fold_envrc_into_env`.
  The overlay twin (`deploy_to_home_overlay`) inherits it. Sibling-module
  split keeps `_to_home.py` near its line ceiling (mirrors `_to_home_text` /
  `_to_home_overlay`).

## Tests (no mocks)

- 26 new tests in `test__to_home_skills.py` — real files on `tmp_path`, real
  `materialize_to_home` / `deploy_to_home` entrypoints. Cover: inlining
  (single/multiple/nested), marker emission, content preservation,
  idempotency (byte-identical re-run, changed-skill refresh, single-block),
  removal cleanup, create-when-absent, both-CLAUDE.md targeting, and
  **marker-protection coexistence across re-deploys + user-tail**.
- AAA `# Arrange`/`# Act`/`# Assert`; one assertion per test.
- No regression: existing to_home suites (93 pass), full runtimes suite
  (976 passed, 14 pre-existing skips).

## Note on the legacy `SkillsSpec`

`config._types.SkillsSpec` (`required`/`available`/`injection_mode="at-import"`)
is parsed but **unconsumed** in the deploy path — the v3 docs confirm
`spec.skills` was removed and at-import was never wired. This PR deliberately
sources skills from the materialized `.claude/skills/` tree (the v3 reality),
not from that vestigial spec field.

### DIFF
--- a/src/scitex_agent_container/runtimes/_to_home.py
+++ b/src/scitex_agent_container/runtimes/_to_home.py
@@ -72,6 +72,7 @@ from ._to_home_errors import (
     WorkspaceCredentialLeakError,
     WorkspaceMcpMergeError,
 )
+from ._to_home_skills import aggregate_skills_into_claudemd
 from ._to_home_text import (
     END_MARKER,
     extract_user_tail,
@@ -220,6 +221,10 @@ def materialize_to_home(spec_dir: Path, workspace_home: Path) -> None:
     if root.is_dir():
         _walk_and_apply(root, root, workspace_home, config=None)
     fold_envrc_into_env(workspace_home)
+    # Inline the agent's now-materialized skills into CLAUDE.md (forceful,
+    # idempotent) so they are always in context regardless of @-import /
+    # skill-chaining reliability. See _to_home_skills for the rationale.
+    aggregate_skills_into_claudemd(workspace_home)
 
 
 def deploy_to_home(config: AgentConfig, workspace_home: str) -> None:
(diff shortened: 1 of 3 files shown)


ywatanabe1989:

PR #467, opened 2026-06-25
fix(spawn): host-reachable SAC_LISTEN_BASE_URL + spec-dir env for in-container spawn

ywatanabe1989:
## Summary
A maintainer agent INSIDE a sac container could not spawn/start peers.

- **Root cause (spawn timeout):** `runtimes/_apptainer_listen_env.listen_env_flags` injected `SAC_LISTEN_BASE_URL` from `_listen/_config.listen_base_url()`, which builds `http://{listen_host()}:{port}` — and `listen_host()` defaults to the loopback `127.0.0.1` the listen server binds (`cli_pkg/listen_cmds.py:175`). So every agent container got `SAC_LISTEN_BASE_URL=http://127.0.0.1:7878`. The in-container spawn client (`_lifecycle/_spawn_client._resolve_base_url`) dialed that, hitting the container's OWN loopback where nothing listens → POST timed out. a2a worked because it derives the host from `host_config.canonical_host()` (the canonical hostname), not loopback (`_listen/_registry_endpoints.py:113`).
- **Root cause (agent_start no specs):** `build_run_argv` never injected `$SCITEX_AGENT_CONTAINER_YAML_DIRS`, so in-container spec resolution (`config/_resolve._search_dirs`) found nothing.

## Fix
- New `container_reachable_host()` / `container_listen_base_url()` in `_listen/_config.py` resolve the container-facing URL via the same canonical-host path a2a uses (explicit non-loopback `listen.host` wins; else `host_config.canonical_host()`; else loopback fallback).
- `listen_env_flags` now injects that host-reachable `SAC_LISTEN_BASE_URL` (not 127.0.0.1) and forwards `$SCITEX_AGENT_CONTAINER_YAML_DIRS` verbatim when set on the host.

## Test plan
- [x] New `test__apptainer_listen_env.py`: asserts injected `SAC_LISTEN_BASE_URL` is host-reachable (not 127.0.0.1) and uses the canonical hostname; spec-dir env forwarded only when set.
- [x] New `TestContainerReachableHost` in `test__config.py`: loopback default resolves to canonical host; explicit non-loopback host honoured.
- [x] Updated 5 `test__apptainer_runtime.py` cases that previously asserted the buggy loopback URL.
- [x] 191 tests pass.

## Verify live
On the real two-host fleet: from inside a sac container, `agent_spawn` should POST to `http://<host-hostname>:7878/agents` (not 127.0.0.1) and succeed; `agent_start` should resolve specs via the forwarded spec-dir env. Note: forwarded spec dirs must also be bind-mounted/readable inside the container (typically true for `~/.scitex/...`); if an operator's spec dir is outside the bound home, add a bind.


ywatanabe1989:
## Do not merge as-is — proven to regress same-host spawn

I tested reachability from **inside a live sac container** on ywata-note-win (apptainer shares the host netns by default, so the container and host see the same interfaces):

| URL | result |
|---|---|
| `http://127.0.0.1:7878/v1/sac/health` (current behaviour) | **401 in 3ms** — reachable, server up, bearer-gated |
| `http://ywata-note-win:7878/v1/sac/health` (**what this PR injects**) | **000 — connection refused** |

On this host `container_listen_base_url()` resolves to `http://ywata-note-win:7878`. The listen binds **loopback-only** (this PR's own report cites `listen_cmds.py:175`), so the canonical-hostname interface is not bound → refused. This PR replaces a working same-host URL (`127.0.0.1`, 401 = alive) with a refused one, **breaking same-host spawn** — the common case (agents on ywata-note-win spawning peers on the same host).

The loopback→hostname swap can't fix cross-host either: a loopback-bound server is unreachable from any peer host regardless of the URL string. Cross-host spawn needs the **server** to bind a reachable interface (`0.0.0.0`, gated by the bearer it already requires) — a separate, security-relevant change that is not in this PR.

Most likely the original "spawn timeout" is a **401 misreported as a timeout** (same class as the listen-restart 401-as-down fix): the in-container spawn client reaches the loopback listen but isn't sending the bearer → 401 → surfaced as "cannot reach / timed out". That is the thread to pull — audit `_lifecycle/_spawn_client.py` bearer/`Authorization` handling, not the URL.

**Keep:** the `$SCITEX_AGENT_CONTAINER_YAML_DIRS` forwarding half is correct and independently valuable — worth landing on its own.

Converting to **draft**. Reworked plan: (1) spawn client sends the bearer and surfaces a 401 as auth-needed, not unreachable; (2) keep loopback for same-host; (3) cross-host = a separate listen `0.0.0.0`-bind decision (security). Full evidence recorded on card `sac-agent-cannot-spawn-agents-listen-7878-unreachable-20260625`.

ywatanabe1989:
Closing — NOT superseded-by-another-route, and not needed. Settled empirically from inside a live agent container rather than by reading code.

## The change never landed
develop still injects the loopback verbatim:
```
src/scitex_agent_container/runtimes/_apptainer_listen_env.py:73
    flags: list[str] = ["--env", f"SAC_LISTEN_BASE_URL={listen_base_url()}"]
```
and `_listen/_config.py` has no canonical-host resolver — only `DEFAULT_LISTEN_HOST = "127.0.0.1"`. So the card claiming this fix "re-shipped" was wrong: it never merged, by this route or any other.

## But the premise is false, so the fix is unnecessary
This PR's docstring says the loopback "points the spawn client at the *container's own* loopback (**nothing listens there**)". That is not true of the runtime the fleet actually uses. Measured from inside a running agent container just now:

- `SAC_LISTEN_BASE_URL=http://127.0.0.1:7878` (i.e. exactly the value this PR calls broken)
- `curl $SAC_LISTEN_BASE_URL/health` -> **HTTP 401 in 0.01s** — something is listening and it *answered*; not a refusal, not a timeout
- same request **with** the injected `SAC_LISTEN_BEARER` -> **HTTP 404**, not 401 — the bearer was **accepted**, so it is the real `sac listen` (only the route is absent)
- `/proc/self/net/tcp` shows **4 sockets on port 7878 visible in the container's own netns**

Root cause of the misdiagnosis: **apptainer shares the host network namespace by default** — sac passes no `--net` anywhere in `runtimes/` — so the container's `127.0.0.1` *is* the host's loopback. In-container spawn / `host_exec` work today over precisely this path; this agent is using it.

## Caveat worth keeping
The concern becomes real the moment agents run in an **isolated** netns (a Docker bridge, or apptainer with `--net`). If sac ever moves the agent runtime there, `SAC_LISTEN_BASE_URL` must resolve to a host-reachable name and this PR is the sketch of that fix. That is a note for the day it happens — not a reason to keep a 37-day-old red PR (on a `claude/agent-<hash>` branch) rotting.

PR #507, opened 2026-07-01
feat(preflight): fail-fast guard against $HOME clobbering operator dotfiles

ywatanabe1989:
## Why
The agent's canonical `$HOME=/home/agent` is overlay/tmpfs-backed and per-agent — that isolation is what stops the agent's Claude Code + shell writes (`~/.claude/.credentials.json`, session transcripts, `~/.bash_history`, `~/.config`, caches, git config) from **clobbering the operator's real dotfiles**. A `relaxed: true` spec can move `$HOME` via `--home` in `raw_args` onto a real host directory that is also bind-mounted read-write, defeating the isolation. The existing in-container preflight (`_apptainer_preflight.py`, asserts `$HOME == /home/agent`) is **bypassed by relaxed specs** and runs inside the container (too late).

## What
A host-side, fail-fast, fail-loud guard in `agent_start`, run **before** the container spawns and **unconditionally** (not gated by `--no-preflight`, since this is a data-safety invariant, not a liveness check).

**Invariant:** refuse launch if the resolved container `$HOME` (`resolve_container_home`) **equals or is a subpath of** the destination of any *operator-declared* `:rw` bind (from `apptainer.binds` or `raw_args --bind`). sac's own default binds and `:ro` identity mounts are excluded.

- Passes every current spec (neurovista etc. keep `$HOME=/home/agent`, which is never under the rw whole-home bind). Verified: no `--home` in the repo points anywhere but `/home/agent`, so zero false positives.
- Refuses e.g. `--home /home/ywatanabe` + `/home/ywatanabe:/home/ywatanabe:rw` with a message naming the offending bind and home.

## Files
- `runtimes/_home_isolation_guard.py` — new, pure function `assert_home_not_operator_writable(config)`.
- `_lifecycle/_start.py` — wired in after the existing telegrammer-wiring validator.
- `tests/scitex_agent_container/runtimes/test_home_isolation_guard.py` — 5 tests (2 pass-paths, 3 clobber-refusals incl. raw_args `--bind`), no mocks, all green.

## Note
Stacked on `fix/remove-stray-fakeroot-junk` (the develop CI-unblock PR). Once that merges, this PR's diff reduces to just the guard.

## Test plan
- [ ] New guard tests green in CI
- [ ] Full matrix green

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_start.py
+++ b/src/scitex_agent_container/_lifecycle/_start.py
@@ -213,6 +213,14 @@ def agent_start(
         agent_name=config.name,
     )
 
+    # Data-safety invariant (UNCONDITIONAL — not gated by ``no_preflight``):
+    # refuse specs whose $HOME resolves onto an operator-writable host bind,
+    # which would let the agent's home writes clobber the operator's real
+    # dotfiles. See :mod:`..runtimes._home_isolation_guard`.
+    from ..runtimes._home_isolation_guard import assert_home_not_operator_writable
+
+    assert_home_not_operator_writable(config)
+
     runtime_factory = runtime_factory or _get_runtime
     runtime = runtime_factory(config)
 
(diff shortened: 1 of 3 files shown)


ywatanabe1989:

PR #508, opened 2026-07-01
feat(cards): deterministic card-ownership / false-blocker guard

ywatanabe1989:
## Why (incident 2026-07-01)
An agent looked at its OWN in-progress cards and reported them as "gated / not on my side" — including a card whose only blocker was that agent's own mergeable PR. Operator flagged this as a serious incident: agents wrongly deciding their job is not theirs. This makes the rule deterministic.

## What
A pure, no-mocks validator over the shared task store (reads tasks.yaml as data — no scitex_todo import, honors the standalone mandate). For each active card it fails loud on:
- **self-block** — a blocker card is assigned to the SAME agent (it's your job, not a block)
- **void-blocker** — a blocker card is in a terminal state (done/cancelled/deferred) — the exact "consumer was cancelled" case
- **blocked-no-blocker** — status `blocked` with nothing registered

Runnable now via `python -m scitex_agent_container._lifecycle._card_hygiene [--agent X]` (exit 1 + list on violations). Wiring into the per-agent Stop-hook (so a turn can't end leaving a card falsely-blocked) and the liveness-tick reconciler is the next layer, tracked on the incident card.

## Tests
6 tests, real fixtures, one assert each, AAA — all green.

## Test plan
- [ ] CI green

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_card_hygiene.py
+++ b/src/scitex_agent_container/_lifecycle/_card_hygiene.py
@@ -1,0 +1,149 @@
+"""Deterministic card-hygiene validator (blocker-validity pillar).
+
+Incident 2026-07-01: an agent looked at its OWN in_progress cards and labelled
+them "gated / not on my side" — including a card blocked only on its own
+mergeable PR. A valid blocker is a CARD owned by a DIFFERENT appropriate agent;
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

PR #509, opened 2026-07-01
feat(listen): zombie-agent reconciler — auto-clear stale 'running' registrations

ywatanabe1989:
## Why (operator directive, 2026-07-01)
After a host SIF rebuild or a `sac listen` restart, agents can become **zombies**: the registry still marks them `running` (an active `instances` lease), but there's no live apptainer container and their a2a inbox is dead. A plain `sac agents restart` then no-ops ("Agent X is already running. No-op. Use --force") — so the dead agent stays dead until someone force-fresh restarts it. Hit today on scitex-scholar, scitex-writer, openalex-local, crossref-local, scitex-clew (all fixed manually). Operator: "zombie clearing must be implemented in sac listen server."

## What
A reconciler in the listen lifespan that each tick finds agents the registry marks running with **no live container** and clears their stale lease, so the registry becomes truthful and a normal restart relaunches cleanly.

- **Detect + clear only** (auto-relaunch + bus events are deliberate follow-ons).
- "Live container" = the runtime's `apptainer_pid` file exists **and** its pid is alive.
- **Subtlety handled:** zombie lease rows have `pid IS NULL`, so the existing `clear_stale_instance_lease` would *skip* them. We close the row directly via `record_instance_stop` **after** proving no live container.
- All blocking IO (registry read, pid probe, lease close) routes through `_off_loop.run_blocking_or` so it can never starve the uvicorn bind — same defense as the liveness-tick loop. Gated by `SAC_ZOMBIE_RECONCILE_DISABLED`; interval via `SAC_ZOMBIE_RECONCILE_INTERVAL_S` (default 120s). Teardown-cancelled with the other lifespan tasks.

## Files
- `_listen/_zombie_reconciler.py` — pure `find_zombie_agents` / `reconcile_zombies` behind injected seams, + production seams + the loop.
- `_lifecycle/_listen_lifespan.py` — launch the loop alongside the liveness tick.
- `tests/.../_listen/test_zombie_reconciler.py` — 12 tests, no mocks (plain fakes), one assert each.

## Notes
Verified: all production-seam imports resolve, lifespan imports cleanly, 12/12 tests green. This is the deterministic replacement for today's manual force-fresh sweeps. Follow-ons (separate cards): auto-relaunch cleared zombies; a health signal when an a2a send hits a dead inbox; `sac agent health`.

## Test plan
- [ ] CI green (tests + audit + ruff)


ywatanabe1989:

PR #527, opened 2026-07-04
feat(a2a): default-allow messaging cross-group (keep host_exec/lifecycle gated) + expose agent role/owner in peers

ywatanabe1989:
## Summary

Two independent, surgical changes in the `sac listen` a2a layer. The
messaging-vs-privileged split is the core principle:

- **a2a MESSAGING = DEFAULT-ALLOW** (collaboration, not a security boundary).
- **PRIVILEGED ACTIONS = still allowlist/mesh-gated** (host_exec + lifecycle).

### 1. a2a messaging default-allow (cross-group)

Previously a cross-group `a2a_send` with no shared group / mesh / grant was
**denied** — which ACL-blocked agents in non-mesh groups (e.g.
paper-scitex-clew's paper-group) from a2a-ing developer agents like
scitex-python / sac. The mesh from PR #12/#524 only covered
developer/researcher/generalist/privileged.

**Check changed:** `_listen/_acl.py::check_send_acl` — the final cross-group
fallthrough flips from `("deny", …)` to `("allow", None)`.

**Overrides preserved** (evaluated *before* the default, unchanged):
- explicit block — `state_db_blocks.has_block` → `"block"` (a blocked sender
  is still denied + 403'd).
- per-spec `spec.comms` parent/siblings=deny — `_phase3_relationship_deny`
  → `"deny"` (operator's explicit per-agent isolation still wins).
- identity-spoof / missing-target / no-identity → still deny.

**NOT touched (the security boundary):**
- `host_exec` eligibility — `_listen/_host_exec.py::ELIGIBLE_GROUPS`
  (developer / researcher / privileged) — unchanged.
- agent lifecycle management (start/stop/restart/delete/status/tail) —
  `_listen/_acl.py::check_lineage_acl` — unchanged.

Consequence: the receiver "approve-prompt" (grant-or-block on a cross-group
deny) can no longer fire from a send — under default-allow there is no
un-granted sender awaiting approval; the receiver's control is now BLOCK.
The block/unblock primitives + CLI are unchanged; the obsolete approve-prompt
integration tests are converted to direct unit coverage of the helpers.

### 2. Expose agent role + owned project in a2a peers

`GET /agents` (backing `mcp__sac__a2a_peers`) now includes each agent's
`role` and owned `project` so ownership is discoverable fleet-wide
("scitex-dev owns X" visible to all).

**Added:** `_listen/_registry_endpoints.py::enrich_row_with_role_owner` +
`resolve_role_and_project`, wired into `_listen/server.py::list_agents`
alongside the existing endpoint enrichment. `role` = spec
`metadata.labels.role`; `project` = basename of the agent's resolved
workdir. Best-effort + idempotent (a row already carrying role/project keeps
it; an unresolvable spec degrades the field to `null`).

## Tests

- Cross-group send that was DENIED is now ALLOWED (`check_send_acl` +
  end-to-end `message:send` < 400); a BLOCKED sender is still denied
  (`"block"`).
- host_exec eligibility UNCHANGED — a non-eligible group is still
  host_exec-denied (`test__host_exec.py`, kept); `ELIGIBLE_GROUPS` pinned.
- lifecycle-management ACL UNCHANGED — non-mesh solver still unmanageable
  cross-group; non-developer with no lineage edge still denied
  (`test__acl_group.py`, kept).
- Peers response includes `role` + `project` per agent (injected-resolver
  unit tests + real on-disk-spec resolution).
- Deny-notification machinery (comms item D + synthetic notify) re-based
  onto the surviving per-spec-deny path.

## Deploy

Needs a **`sac listen` daemon restart** to take effect (the ACL check and
peers serializer run inside the daemon).

## Note on CI

7 pre-existing failures in this repo's `_listen`/`a2a` suite are
environmental (`SacBinaryNotFoundError` — no `sac` on PATH in the bare test
env) and fail identically on `develop`; they are unrelated to this change.


ywatanabe1989:
Rescuing this PR — auto-merge has been armed since 2026-07-04 and could never fire: the branch was CONFLICTING and red, and nobody was watching.

## What I found

This PR has two halves, and **one of them has already shipped without it.**

**Half 1 — "expose agent role/owner in peers": SUPERSEDED.** develop already has a function with the *same name*, `enrich_row_with_role_owner`, and it is a strict superset of this PR's:

| | this PR | develop |
|---|---|---|
| resolver | `resolve_role_and_project` → `(role, project)` | `resolve_agent_identity` → full spec identity |
| fields | `role`, `project` | `role` **+** `responsibilities`, `groups`, `purpose`, `project` (omit-if-missing) |
| source | ad-hoc spec read | delegates to `a2a._card_identity.spec_identity` — the **same projection the AgentCard uses**, so the two a2a surfaces cannot drift |
| directive | operator 2026-07-03 | operator **2026-07-06** |

develop also added `enrich_row()` = `enrich_row_with_role_owner(enrich_row_with_endpoint(row))` — precisely the composition this PR was doing by hand in `server.py` — and wires it into the peers list at `_agents_list.py:87-89`:

```python
from ._registry_endpoints import enrich_row
rows = [enrich_row(row) for row in rows]
```

So role/owner in `a2a peers` is **already live**, via a later and better implementation of the same operator directive. Meanwhile develop had refactored `list_agents` out of `server.py` into `_agents_list.py`, which is what made this branch conflict in the first place.

**Half 2 — the ACL default-allow change: still wanted, and this is the only place it exists.** That is the real remaining contribution.

## What I did

Resolved the conflict by **dropping the superseded half and keeping the ACL half**:

- `_registry_endpoints.py`, `server.py`, `test__registry_endpoints.py` → take develop's (this PR's `resolve_role_and_project` + its weaker `enrich_row_with_role_owner` + their 135 lines of tests are redundant, and would have *regressed* the peers row by dropping `responsibilities` / `groups` / `purpose`).
- `_acl.py` + the four ACL test files + `test_server.py` → keep this PR's. `test_server.py`'s changes are ACL semantics (cross-group send now allowed; deny driven by a per-spec `inbound.siblings=deny`), not role/owner, so they stay.

## The ACL change, reviewed

The policy is sound and matches the operator's 2026-07-03 directive: a2a **messaging** becomes default-allow cross-group (collaboration is not a security boundary), while the boundary **stays on privileged actions** — `host_exec` (`ELIGIBLE_GROUPS`) and lifecycle management (`check_lineage_acl`) are explicitly untouched. The overrides that still deny are evaluated *before* the default allow: an explicit block (`has_block` → `"block"`) and a per-spec `spec.comms` parent/siblings deny. Unauthenticated, identity-spoof, and missing-target all still deny.

Watching CI to merge — not arming auto-merge and walking away.

ywatanabe1989:
Second red, second root cause — and this one was a real gap in the PR, not just staleness.

## The smoke layer still encoded the old contract

The ACL change was applied to the `_listen` unit tests but not to `tests/smoke/`, where **16 e2e tests still asserted "a cross-group send with no grant is a 403"** — exactly the contract this PR retires. Under default-allow that send now succeeds, so they failed. (Same shape as the sibling miss elsewhere in this rescue: the *second* test file always encodes the old contract.)

## The trap in fixing them

Re-triggering the deny is not a free choice. `_node_channel` splits the verdicts, and the two are **not** interchangeable:

| verdict | sender | receiver |
|---|---|---|
| `"block"` | 403 | **deliberately nothing** — no denied_attempt, no prompt, so the block flag cannot leak |
| `"deny"` | 403 | denied_attempt notification + persisted rows |

These tests exist to pin the *receiver-side* surface. Had I reached for `block_send` — the obvious choice — the 403 assertions would have gone green while every receiver-side assertion silently stopped being exercised. So the deny has to be a `"deny"`: new helper `_set_up_denied_send` makes alpha and gamma **siblings** and gives gamma a per-spec `inbound.siblings = deny`, an override `check_send_acl` evaluates *before* the default allow. Every receiver-side assertion (403, empty content column, deny reason, `denied_attempt` kind, no body leak) is preserved.

`test_cross_group_send_without_grant_returns_403_with_reason` is **inverted, not deleted** — it now asserts the send is allowed. The contract changed; the test should state the new one plainly.

## The approve-prompt: a real casualty, and intentional

`_node_channel` emits the task-#27 receiver approve-prompt on a **cross-group deny only** — *"the only deny reason the receiver can REMEDY via grant/block"*. This PR removes that deny, so **no send can raise a prompt any more**. That is correct: a grant is no longer the remedy for anything, because nothing is denied for want of one.

I want to be clear that this was **not my call to make** — *this PR's own author already made it*, in `test__acl_approve_flow.py`:

> "Under messaging DEFAULT-ALLOW … a cross-group send no longer denies, so the receiver approve-prompt no longer fires from a send. The surviving, still-meaningful behaviour this file covers is the BLOCK/UNBLOCK decision primitives…"

They converted those tests from send-triggered e2e to direct unit tests of the primitives. I applied the same decision to the smoke layer they missed. **Every assertion survives** as a unit test — dedupe, embeds `sac a2a unblock`, embeds `sac a2a block`, no body leak. The four smoke tests that drove the prompt are replaced by a comment naming their replacements, and the denied-send **no-leak guarantee is untouched** and still pinned in the smoke layer on the rows a denied send does still persist.

## Verification

- **29 passed** in `tests/smoke` (was 16 failing)
- **776 passed** in `tests/scitex_agent_container/_listen`
- both run against the **worktree** source (`PYTHONPATH=<worktree>/src` — a bare pytest here imports the *installed* sac and proves nothing)
- ruff delta vs develop: **zero**

The `scitex-dev-quality-audit` red on the last run was **not the code** — `astral-sh/setup-uv@v3` died with `Unknown system error -116` unpacking its tarball. A runner flake; re-running.

## Disclosed, not hidden

`_looks_like_cross_group_deny` and the approve-prompt branch it guards in `_node_channel.py` are now **unreachable** for messaging. Dead code that still *looks* live is how the next reader gets misled. I deliberately kept it out of this diff rather than widen a security-relevant PR, and carded it: `sac-acl-approve-prompt-unreachable-after-default-allow`.

PR #528, opened 2026-07-04
feat(runtime): SCITEX_AGENT_CONTAINER_RUNTIME_DIR knob + prune runtime/overlay for restart.policy:never on stop (inode hygiene)

ywatanabe1989:
## Why

Permanent fix for the **sac-runtime-state-hygiene** inode-accumulation incident: per-capsule runtime dirs + overlay uppers accumulate one-per-run and exhaust a GPFS-home fileset across a large launch wave (the incident: 1.25M+ inodes from a cohort's capsules).

The two changes are independent in the diff.

## Change 1 — one relocatable runtime base (env knob)

- New `_runtime_paths.runtime_base_dir()` SSOT: returns `$SCITEX_AGENT_CONTAINER_RUNTIME_DIR` (expanded/abspath) when set, else the historical `~/.scitex/agent-container/runtime` default (**unchanged**).
- Rerouted every hardcoded runtime-base call site through it:
  - `_state/state_db.py` → `state.db`
  - `_state/registry.py` → `registry/`
  - `_runners/_session_state.py` → per-agent state root (`DEFAULT_STATE_ROOT`)
  - `_lifecycle/_session_reset.py` → `--force` session-reset dir
  - `cli_pkg/db_group.py` → `sac db migrate` registry default
- Per-file overrides (`SCITEX_AGENT_CONTAINER_STATE_DB` / `_REGISTRY_DIR`) still win. Setting the **one** env relocates the whole tree node-local (`SCITEX_AGENT_CONTAINER_RUNTIME_DIR=$TMPDIR/sac-rt`); unset ⇒ byte-identical to before.
- `state_db.py` crossed the 512-line cap once the import landed, so the DDL string constants moved verbatim into `_state/state_db_schema.py` (re-exported; no behaviour change).

## Change 2 — prune runtime + overlay for ephemeral agents on stop

- New `_lifecycle/_prune_runtime.py`. **Gate (conservative):** `restart.policy == "never"` **AND** `restart.prune_on_stop is True`.
- **Why the explicit opt-in** (not never-alone / default-on): `restart.policy` **defaults to `"never"`**, so a persistent coordinator that simply omits a `restart:` block is policy==never yet must keep its runtime (session_id, home/) across restarts. Requiring the flag guarantees persistent agents are **never** pruned.
- New `spec.restart.prune_on_stop` (bool, default False) + parser + validation.
- Wired into `agent_stop(prune_runtime=False)`; only the terminal `sac agents stop` CLI passes `prune_runtime=True`. The internal `agent_stop` calls in `agent_restart` / force-`agent_start` keep the default, so a restart never nukes a runtime it is about to reuse.
- Best-effort + **fail-loud-log**; prunes only the runtime dir + overlay dir, never launcher output dirs.

## Tests
- Resolver: env set/unset, module-constant relocation, per-file override precedence.
- Prune: gate matrix, runtime-dir + overlay removal, missing-dir → no crash, persistent + non-opted-in never pruned.
- `agent_stop` wiring: ephemeral pruned / persistent kept.
- Parser + validation for `prune_on_stop`.

Do NOT merge (per task).


ywatanabe1989:
Rescuing this PR — auto-merge has been armed since 2026-07-04 and could never fire, because the PR was red and nobody was watching.

## Why it was red

Two things, neither of them the feature.

**1. The test doubles never learned the new kwarg.** This PR teaches the terminal-stop path to opt into the inode-hygiene prune:

```python
agent_stop(name, force=force, prune_runtime=True)
```

but all eight stand-ins for `agent_stop` in `test__stop.py` still had the old signature, so each one raised `got an unexpected keyword argument 'prune_runtime'` the instant the CLI called it. The CLI swallowed the `TypeError` as a failed stop, the recorders recorded nothing, and eight tests died on `assert [] == [...]`. A stand-in that cannot be *called* the way production calls it is not testing production. Fixed by giving the doubles the real signature (keyword-only `prune_runtime`, default `False`).

**2. Staleness.** The branch was ~10 days / 580 files behind develop, and the ruff leg was red on 3 real errors (F401 `os`, F401 `Path`, I001). develop merged in cleanly (no conflicts); ruff delta against develop is now **zero**.

## Guarding the seam that broke

Tolerant doubles are why this failed loudly today — but they also mean that *deleting* `prune_runtime=True` from the call site would now regress in **silence**. So the forwarding gets its own assertion: `test_terminal_stop_forwards_prune_runtime_to_agent_stop`. The flag is passed unconditionally on the terminal path; the gate that restricts pruning to opted-in ephemeral agents (`restart.policy: never` + `prune_on_stop: true`) lives inside `agent_stop`, so persistent agents stay untouched.

## Verification

61 green (51 stop + 10 prune_runtime), run with `PYTHONPATH=<worktree>/src`.

That detail is load-bearing and worth flagging for everyone working in this repo: **pyproject sets no pytest `pythonpath`, so a bare local `pytest` imports the INSTALLED sac from site-packages, not your checkout.** My first run of this very file reported **50/50 passed** while testing a package that does not contain this PR at all. The giveaway was a collection `ImportError` for `_prune_runtime` — a file this PR *adds*, absent from site-packages. If you validate a source change here without that env var, your green means nothing.

Watching CI to merge — not arming auto-merge and walking away.

PR #538, opened 2026-07-05
fix(apptainer): scrub legacy scitex-todo env vars from the launch environment

ywatanabe1989:
## Incident

Reported convergently by two peer agents, `claude-code-telegrammer` and
`scitex-dev`: agent containers were seeing the LEGACY env var
`SCITEX_TODO_AGENT` present in their process environment alongside the
current `SCITEX_TODO_AGENT_ID`, even though:

1. `/home/agent/.mcp.json` in these containers only forwards
   `SCITEX_TODO_AGENT_ID`.
2. Their `spec.yaml` only sets `--env=SCITEX_TODO_AGENT_ID=<name>`,
   nothing named `SCITEX_TODO_AGENT`.
3. `src/scitex_agent_container/runtimes/_to_home_text.py` already lists
   `SCITEX_TODO_AGENT` in `_RUNTIME_ONLY_VARS` as a "Legacy pre-0.7.30
   name — kept as a GUARD only (never injected by sac anymore)".
4. Yet `/proc/1/environ` inside affected containers showed BOTH vars
   baked into the apptainer instance's env snapshot at container-start
   time.

Net effect: scitex-todo's MCP write path (`add_task` / `update_task` /
`comment_task`) does a strict "old var present → hard reject" check
regardless of whether old and new values agree, so the leaked legacy
var broke writes fleet-wide.

I searched the shared `scitex-todo` task store for an existing
`incident-sac-apptainer-ambient-env-leak-20260705`-style card (by
`id_prefix` and by keyword across the `scitex-dev` / `scitex-todo`
scopes) and did not find one — either it hasn't landed yet or lives
under a different id. Happy to cross-link once it surfaces.

## Root cause

Apptainer, by default, passes the FULL environment of whatever
host shell/process invokes `apptainer exec` / `instance start` into
the container — NOT just sac's explicit `--env=` list. This repo
already relies on and documents exactly this passthrough class
elsewhere (`_apptainer_host_env.py`'s `APPTAINERENV_APPEND_PATH`
mechanism, whose docstring says "This works even under
`--containall`"). So a stale `SCITEX_TODO_AGENT` export left over in
an operator's launching shell (from before the scitex-todo 0.7.30
rename) leaked straight through into every container launched from
that shell, regardless of what sac's own `--env=` list or to_home
materialization said. The existing `_to_home_text.py` guard only ever
stopped the value from being baked into on-disk template files
(`.mcp.json`, `CLAUDE.md`) — it never touched the *live
process-environment* handed to the `apptainer` subprocess.

## Fix

- **Single source of truth**: extracted the legacy-name denylist out
  of `_to_home_text.py`'s `_RUNTIME_ONLY_VARS` into a standalone
  `LEGACY_RENAMED_ENV_VARS = {"SCITEX_TODO_AGENT", "SCITEX_TODO_TASKS"}`
  (`_to_home_text.py:24-42`). `_RUNTIME_ONLY_VARS` now composes it in,
  so the materialized-file guard is unchanged behaviourally.
  (`SAC_NAME` / `CLAUDE_AGENT_ID` / `CLAUDE_AGENT_ROLE` stay out of
  this denylist — those are per-agent-identity vars that must stay
  literal `${VAR}` at deploy time for a *different* reason, not
  renamed-legacy vars that must never leak; scoped per the task's
  instruction to read that comment carefully.)

- **`_apptainer_host_env.py`** (`runtimes/_apptainer_host_env.py`):
  added `LEGACY_ENV_DENYLIST` (= `LEGACY_RENAMED_ENV_VARS`) and a pure
  `scrub_legacy_env(env) -> dict` that returns a copy of `env` with the
  denylisted keys removed — the mirror-image of the existing
  `host_cargo_bin_append_env` (an addition helper) in the same module.

- **Exec site 1 — SDK-Popen path**
  (`runtimes/_apptainer_runtime.py`, the `start()` method): the
  `launch_env` construction now wraps in `scrub_legacy_env(...)`:
  ```python
  launch_env = scrub_legacy_env(
      {**os.environ, **host_cargo_bin_append_env(os.environ)}
  )
  ```

- **Exec site 2 — tmux/TUI path**
  (`_runners/_tmux/tmux.py`, `TmuxManager.start`): this call site was
  actually **worse** — `subprocess.run(argv, check=False)` passed NO
  `env=` kwarg at all, so it silently inherited the full ambient
  environment of the sac process into the tmux pane, in addition to
  the explicit `-e KEY=VAL` overrides. Now builds an explicit
  `launch_env = scrub_legacy_env(os.environ)` and passes it via
  `env=launch_env`.

- **Defense-in-depth (belt-and-suspenders)**
  (`runtimes/_apptainer_inner_argv.py`, `build_inner_argv`): added
  `_LEGACY_ENV_UNSET_STEPS`, an `unset SCITEX_TODO_AGENT
  SCITEX_TODO_TASKS` shell step that now runs FIRST, before even the
  `SAC_GIT_*` alias steps, inside the container's `bash -lc` wrapper.
  This wrapper is regenerated fresh on every `sac agents start` — no
  `.def`/image rebuild needed — so a leak that bypasses sac's Python
  env construction entirely (e.g. an operator running raw `apptainer
  instance start` directly) is still caught before `exec`ing the real
  runner.

This generalizes the same pattern already used by
`_apptainer_jail.py`'s `scrub_bind_env` (strip disallowed env vars
from the launch environment before exec, belt-and-suspenders alongside
a fail-loud check). A future scitex-todo (or other) env-var rename
only needs one new entry in `LEGACY_RENAMED_ENV_VARS` — every exec
site and the in-container defense pick it up automatically, no fresh
audit required.

## Files changed

- `src/scitex_agent_container/runtimes/_to_home_text.py` — new
  `LEGACY_RENAMED_ENV_VARS` constant; `_RUNTIME_ONLY_VARS` composes it.
- `src/scitex_agent_container/runtimes/_apptainer_host_env.py` — new
  `LEGACY_ENV_DENYLIST` + `scrub_legacy_env()`.
- `src/scitex_agent_container/runtimes/_apptainer_runtime.py` —
  scrub applied at the SDK-Popen `launch_env` construction.
- `src/scitex_agent_container/_runners/_tmux/tmux.py` — scrub applied
  + explicit `env=` now passed to `subprocess.run` for `tmux
  new-session` (previously implicit/none).
- `src/scitex_agent_container/runtimes/_apptainer_inner_argv.py` —
  new `_LEGACY_ENV_UNSET_STEPS`, prepended first in
  `build_inner_argv`'s shell wrapper.
- `CHANGELOG.md` — Unreleased entry.
- Tests (new/updated):
  - `tests/scitex_agent_container/runtimes/test__apptainer_host_env.py`
    — `scrub_legacy_env` unit coverage (removes denylisted keys, keeps
    others, does not mutate input).
  - `tests/scitex_agent_container/runtimes/test__to_home_text.py` —
    `LEGACY_RENAMED_ENV_VARS` content/shape lock.
  - `tests/scitex_agent_container/runtimes/test__apptainer_inner_argv.py`
    — `_LEGACY_ENV_UNSET_STEPS` content, ordering (before git-alias,
    before `exec`), and a real-bash-execution proof it actually strips
    the var.
  - `tests/scitex_agent_container/runtimes/test__apptainer_runtime.py`
    — new `apptainer_env_dump_on_path` fixture + test:
    `rt.start(cfg, foreground=True)` with a stale `SCITEX_TODO_AGENT`
    in the test process's ambient env; asserts the REAL subprocess env
    (dumped by the fake `apptainer` binary) does not contain it.
  - `tests/scitex_agent_container/_runners/_tmux/test_tmux_start_legacy_env_scrub.py`
    (new file) — real `tmux` binary + real `subprocess.run`;
    `TmuxManager.start` with a stale `SCITEX_TODO_AGENT` in the test
    process's env; reads back production's own env-snapshot-before-exec
    file (`/tmp/sac-tui-env-<session>.txt`) and asserts the var is
    absent. Skips when `tmux` is not on `$PATH` (matches existing
    `test_tmux_session_activity.py` convention).

## Test plan

Ran with an absolute interpreter path against this worktree's source
(`PYTHONPATH=<worktree>/src /opt/venv-sac/bin/pytest ...` since the
installed venv package is a separate non-editable copy):

- [x] `tests/scitex_agent_container/runtimes/test__apptainer_host_env.py`
- [x] `tests/scitex_agent_container/runtimes/test__to_home_text.py`
- [x] `tests/scitex_agent_container/runtimes/test__apptainer_inner_argv.py`
- [x] `tests/scitex_agent_container/runtimes/test__apptainer_runtime.py` (179 passed)
- [x] `tests/scitex_agent_container/_runners/_tmux/test_tmux_start_legacy_env_scrub.py` (real tmux)
- [x] Full sweep: `tests/scitex_agent_container/runtimes/` +
      `tests/scitex_agent_container/_runners/` — **1686 passed, 1
      failed, 20 skipped**. The 1 failure
      (`test_claude_session.py::test_state_dir_for_default_root_is_under_user_home`)
      is a **pre-existing, order-dependent flake unrelated to this
      diff** — it passes cleanly in isolation, and none of this PR's
      changed files touch `_session_state.py` or state-root
      resolution. Confirmed by re-running
      `tests/scitex_agent_container/_runners/test_claude_session.py`
      alone (exit 0, all green).

Not merging this myself per instructions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,29 @@ versioning follows [SemVer](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- **fix(apptainer): scrub legacy scitex-todo env vars from the launch
+  environment** — INCIDENT 2026-07-05 (reported convergently by
+  claude-code-telegrammer and scitex-dev): apptainer passes the FULL
+  ambient environment of whatever host process invokes it into the
+  container, so a stale pre-0.7.30 `SCITEX_TODO_AGENT` /
+  `SCITEX_TODO_TASKS` export left in an operator's launching shell
+  leaked straight through and tripped scitex-todo's old-var-present
+  hard-reject on the MCP write path, even though neither `.mcp.json`
+  nor `spec.yaml` ever named the legacy var. Adds
+  `_apptainer_host_env.scrub_legacy_env` (reusing the single
+  `_to_home_text.LEGACY_RENAMED_ENV_VARS` denylist already used by the
+  materialized-file guard) and applies it at BOTH exec sites —
+  `_apptainer_runtime.py`'s SDK-Popen `launch_env` construction and
+  `_runners/_tmux/tmux.py`'s `TmuxManager.start` (which previously
+  inherited the full ambient env implicitly, with no `env=` kwarg at
+  all). Also adds a defense-in-depth `unset` step, regenerated fresh
+  every boot in `_apptainer_inner_argv.build_inner_argv` (no image
+  rebuild required), for launch paths that bypass sac's Python env
+  construction entirely (e.g. a raw `apptainer instance start`).
+  Extensible: a future var rename only needs one new entry in
+  `LEGACY_RENAMED_ENV_VARS`.
+
 ## [0.21.11] — 2026-06-09
 
 PATCH release. Re-cut of v0.21.10's accounts-list redesign after the
--- a/src/scitex_agent_container/_runners/_tmux/tmux.py
+++ b/src/scitex_agent_container/_runners/_tmux/tmux.py
@@ -159,7 +159,21 @@ class TmuxManager:
             for key, value in effective_session_env.items():
                 argv += ["-e", f"{key}={value}"]
         argv += ["bash", "-c", shell_script]
-        subprocess.run(argv, check=False)
+
+        # ``subprocess.run`` with no ``env=`` kwarg inherits THIS process's
+        # full ambient environment into the tmux pane — the same
+        # full-ambient-env passthrough class that leaks a stale legacy
+        # var (e.g. SCITEX_TODO_AGENT, left over from before the
+        # scitex-todo 0.7.30 rename) straight through into the eventual
+        # ``exec apptainer ...`` inside the pane (INCIDENT 2026-07-05).
+        # Explicitly scrub the denylisted names from the inherited base
+        # before handing it to tmux, matching the SDK-Popen exec site in
+        # ``_apptainer_runtime.py``. The ``-e KEY=VAL`` overrides above
+        # (cargo-bin append etc.) still layer on top via tmux itself.
+        from ...runtimes._apptainer_host_env import scrub_legacy_env
+
+        launch_env = scrub_legacy_env(os.environ)
+        subprocess.run(argv, check=False, env=launch_env)
 
         time.sleep(2)
         return TmuxManager.exists(session_name)
--- a/src/scitex_agent_container/runtimes/_apptainer_host_env.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_host_env.py
@@ -23,6 +23,17 @@ ADDITIONS (never mutates ``os.environ`` itself) so the exec sites
 (:mod:`_apptainer_runtime` SDK-Popen path and :mod:`tui_session` tmux
 path) can merge them into the env they hand to the ``apptainer``
 process.
+
+It also carries the mirror-image SUBTRACTION helper,
+:func:`scrub_legacy_env` — apptainer passes the FULL ambient
+environment of whatever host process invokes it into the container
+(the exact same passthrough class that motivates
+``APPTAINER_APPEND_PATH`` above), so a stale legacy env var left over
+in the launching shell (e.g. ``SCITEX_TODO_AGENT`` from before the
+scitex-todo 0.7.30 rename) leaks straight through regardless of what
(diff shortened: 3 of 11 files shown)


ywatanabe1989:
Superseded by #549 (generic clean-env launch). The name-scrub approach in this PR was the wrong fix per operator directive 2026-07-05: scrubbing a denylist of variable NAMES still means sac knows those names. #549 replaces it with always-`--cleanenv` + a name-agnostic curated launch env, so no ambient var of any name reaches the container and sac names zero packages. Closing in favor of #549.

PR #548, opened 2026-07-05
refactor(channels): remove claude-code-telegrammer knowledge from sac; generic wake-on-push

ywatanabe1989:
## Summary

Per operator directive (2026-07-05): **sac must have ZERO knowledge of the `claude-code-telegrammer` package.** sac is a generic container-lifecycle runtime; it must not hardcode any specific downstream channel package's names, env-var names, or ship package-specific content. This PR removes all `claude-code-telegrammer` coupling from sac source and replaces the package-specific wake-on-push wiring with a **generic, opt-in** mechanism.

### 1. Deleted the Telegram enforcement-hook baseline assets
- Removed `src/scitex_agent_container/_baseline_assets/telegram_hooks/` (5 `enforce/encourage_telegram_*.sh` hooks, `forbidden-words.yaml`, `README.md`, `settings.local.json.fragment.json`) and their integration tests (`tests/integration/telegram_hooks/`).
- **No source copier materialized them** — they were manual reference assets (their README documented a `cp`/`ln -s` recipe). Nothing in sac source referenced `telegram_hooks`, so there was no wiring to unwind and no baseline-settings merge referenced the fragment.

### 2. Deleted the Telegram skill doc
- Removed `_skills/scitex-agent-container/23_telegram-integration.md` and every index/cross-reference: `SKILL.md`, `27_credentials-relogin.md`, `30_responsiveness-background-work.md`.

### 3. Genericized wake-on-push (`runtimes/_sdk_channels.py`)

**Removed** the hardcoded `_TELEGRAMMER_CHANNEL = "server:claude-code-telegrammer"`, `_TELEGRAMMER_MCP_KEY`, `_TELEGRAMMER_TURN_URL_ENV = "CLAUDE_CODE_TELEGRAMMER_TURN_URL"`, `_wire_telegrammer_wake()`, `validate_telegrammer_wake_wiring()`, `TelegrammerWakeWiringError`, and the `ChannelPlan.telegrammer_turn_url` field.

**Replaced** with one generic, package-agnostic value:
- New placeholder name **`SAC_AGENT_TURN_URL`** (matches the `SAC_`-prefix convention, e.g. `SAC_BIN`, `SAC_LISTEN_BASE_URL`).
- `ChannelPlan.agent_turn_url` = the agent's own loopback `http://127.0.0.1:<a2a_port>/v1/turn`, present whenever an a2a port resolves (no channel gating).
[...]

PR #549, opened 2026-07-05
fix(apptainer): generic clean-env launch (no ambient env leak, zero package-name knowledge; supersedes the name-scrub #538)

ywatanabe1989:
## Operator directive (2026-07-05)

sac must have ZERO knowledge of any specific downstream package. The prior
approach — SCRUBBING a denylist of variable NAMES (`SCITEX_TODO_AGENT`,
`SCITEX_TODO_TASKS`) from the launch environment, as in the open **PR
#538** — is itself a violation, because sac's code must NAME those
variables to scrub them. This PR replaces that with a GENERIC mechanism
that names ZERO downstream-package variables: sac controls what
environment the container gets, rather than naming specific vars to
remove. **This PR supersedes PR #538** (please close #538 in favour of
this).

## Root cause

`runtimes/_apptainer_runtime.py` passed the ENTIRE ambient process
environment (`{**os.environ, ...}`) as the env of the host-side
`apptainer` subprocess, and `runtimes/_apptainer_iso_flags.py` only added
`--cleanenv` when `not relaxed`. So for any spec with `apptainer.relaxed:
true` (scitex-agent-container, neurovista, figrecipe, …), apptainer
forwarded the whole ambient env into the container — any stale var in the
launching shell/daemon (e.g. a legacy `SCITEX_TODO_AGENT`) rode through
into the agent's `/proc/1/environ`.

### Primary: `--cleanenv` decoupled from `relaxed`
`runtimes/_apptainer_iso_flags.py` now applies `--cleanenv`
**unconditionally** (unless the operator declared it in `raw_args`), no
longer gated on `not relaxed`. Env-cleanliness is orthogonal to the
filesystem-isolation `relaxed` opt-out (which is about whether
`--containall` / `--writable-tmpfs` / `--home` drop default binds).
`--containall` stays coupled to `relaxed` (filesystem concern); only
`--cleanenv` is decoupled.

**Why this is safe for relaxed agents** — verified empirically on this
host (apptainer 1.5.2) against a real container:

```
WITHOUT --cleanenv:  LEAK_TEST_CANARY=SHOULD_NOT_APPEAR   ← leaks
                     SCITEX_TODO_AGENT=POISON_LEGACY       ← leaks
WITH    --cleanenv:  (neither present)
                     PATH=…:/host/cargo/bin                ← APPTAINERENV_APPEND_PATH STILL applied
```

`--cleanenv` still honours the `APPTAINERENV_*` injection directives, so
sac's `APPTAINERENV_APPEND_PATH` cargo-bin PATH append (the only ambient
directive relaxed agents rely on) survives. Everything else the container
needs already comes from explicit `--env` / `--env-file` (state-db, bus
auth, listen URL/bearer, quota path, `spec.env`, the agent `.env`), never
from ambient passthrough — so nothing breaks.

### Defense-in-depth: curated launch env
`runtimes/_apptainer_host_env.py` adds `minimal_launch_env(base_env)` — a
pure, name-agnostic allowlist: generic system vars (`PATH`, `HOME`,
`USER`, `TERM`, `LANG`, …) + prefix rules for apptainer/system-owned
namespaces (`APPTAINER_*`, `APPTAINERENV_*`, `SINGULARITY*`, `LC_*`,
`XDG_*`). Applied at the SDK-Popen launch site
(`runtimes/_apptainer_runtime.py`) and the tmux/TUI launch site
(`_runners/_tmux/tmux.py`, which previously passed no `env=` at all and
inherited the full ambient env into the pane). Even if `--cleanenv` were
ever off, there is no downstream-package var in the process env to
forward. Contains ZERO downstream-package variable names — the rule is
purely "generic + apptainer-owned namespaces in, everything else out", so
a future package env-var rename needs no change here.

### Attestation card
`a2a/_card.py` now reports `cleanenv: True` unconditionally to match the
always-on behaviour.

## No name-based scrub added / removed
This branch is off `develop`, which does not contain PR #538's
`scrub_legacy_env` / `LEGACY_RENAMED_ENV_VARS`, so there was nothing to
delete. The existing `_apptainer_jail.scrub_bind_env` (scrubs
`APPTAINER_BIND` / `SINGULARITY_BIND…` for jailed capsules) and
`_to_home_text._RUNTIME_ONLY_VARS` (on-disk template-file guard) are
LEFT UNTOUCHED — they are separate, legitimate concerns, not launch-env
name-scrubbing. `grep` proof that the launch/env cleanliness path names
zero downstream vars:

```
$ rg 'SCITEX_TODO|CCT_|CLAUDE_CODE_TELEGRAMMER' \
    _apptainer_iso_flags.py _apptainer_host_env.py _apptainer_runtime.py
(no matches)
```

The `CLAUDE_CODE_TELEGRAMMER_TURN_URL` at `_apptainer_build_argv.py:340`
is an explicit `--env` INJECTION (not ambient scrubbing) and is left to
PR #548, which genericises it to `SAC_AGENT_TURN_URL`; not touched here.

## Verification

**Gold-standard, real apptainer (1.5.2) on this host** — proof above:
without `--cleanenv` a poison `LEAK_TEST_CANARY` + `SCITEX_TODO_AGENT`
leak into a real container; with `--cleanenv` neither does and the
cargo-bin PATH append survives. (Full-agent boot was not possible here —
no fuse device for SIF mount / no agent SIF — so this used an extracted
sandbox rootfs; the argv/env path is identical.)

**Real-subprocess integration test** — a fake `apptainer` that dumps its
own process env, driven by `rt.start(cfg, foreground=True)` with poison
vars set in the test's ambient env; asserts the REAL launch_env handed to
the subprocess excludes both `LEAK_TEST_CANARY` and `SCITEX_TODO_AGENT`,
for BOTH relaxed and non-relaxed specs.

**Unit tests** — `minimal_launch_env` drops arbitrary/legacy vars, keeps
`PATH` + `APPTAINERENV_APPEND_PATH` + `LC_*`, does not mutate input.

Sweep (`PYTHONPATH=<worktree>/src /opt/venv-sac/bin/python -m pytest`,
interpreter confirmed resolving to worktree source):
`tests/…/runtimes/` + `_runners/_tmux/` + `integration/test_v3_spec_structure.py`
→ **1269 passed, 20 skipped**. `scitex-dev ecosystem audit-all` clean
except one PRE-EXISTING `[§4]` CLI `--help` version-line error unrelated
to this diff.


ywatanabe1989:

PR #569, opened 2026-07-07
fix(runtimes): auto-degrade to no-bus on listen-less node instead of fatal abort

ywatanabe1989:
## What fataled

`runtimes/_apptainer_listen_env.py::listen_env_flags` guarded the bus:
when `spec.claude.channels` contained `server:sac` but no `sac listen`
bearer token resolved on the node, it RAISED `RuntimeError`, aborting
the container launch. On a node with no `sac listen` — e.g. a Spartan
SLURM compute node running a one-shot cohort-solver capsule (no per-node
`sac listen`) — the capsule supervisor restarted straight back into the
same abort, producing an **infinite fresh-restart loop**. `channels: []`
did not help because `config/_loaders.py` force-appends `server:sac` onto
every agent's channels unless the label `sac-builtin: "off"` is set.

## The fix (drop + warn + continue)

In the `elif wants_bus:` branch, instead of raising:
1. DROP `server:sac` from `config.claude.channels` in place. This helper
   runs (build argv line ~327) BEFORE the inner-argv builders
   (`tui_channel_config`, `build_inner_argv`) read the same `channels`
   list, so the bus adapter is never registered — no 401 auth attempt,
   no message-reader fatal.
2. Emit a loud WARNING explaining the bus is disabled because no
   `sac listen` token is present.
3. Fall through and launch normally (bus disabled).

A listen-less node now **auto-degrades to "no bus"** instead of aborting.

The token-PRESENT path is UNCHANGED: `server:sac` + `SAC_LISTEN_BEARER`
are still injected. The `config/_loaders.py` force-append is untouched,
and the label `sac-builtin: "off"` remains the pre-existing FULL opt-out.

## Tests

Extended `tests/scitex_agent_container/runtimes/test__apptainer_listen_env.py`:
- bus wanted + token ABSENT → no `RuntimeError`, `server:sac` dropped,
  other channels kept, warning emitted.
- bus wanted + token PRESENT → `server:sac` + bearer still injected
  (regression guards).

No existing test asserted the raise behavior, so none needed updating.
Only existing files were edited (no new source/test files → orphan
test-parity holds).

```
PYTHONPATH=<worktree>/src /opt/venv-sac/bin/python3 -m pytest \
  <worktree>/tests/scitex_agent_container/runtimes/test__apptainer_listen_env.py -q
# 24 passed
```

### DIFF
--- a/src/scitex_agent_container/runtimes/_apptainer_listen_env.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_listen_env.py
@@ -21,12 +21,17 @@ specs (``--containall`` + explicit ``raw_args``) bypass the preflight
 wrapper but still need bus auth, otherwise their adapter can never
 subscribe.
 
-Fail-loud contract: when ``server:sac`` is registered but the bearer
-cannot be resolved, this raises ``RuntimeError`` rather than launching an
-agent whose adapter can never authenticate (the silent wake-on-push
-failure this guard exists to prevent). When ``server:sac`` is absent a
-missing token is harmless (nothing subscribes) — we inject only the base
-URL and log a loud warning.
+Auto-degrade contract: when ``server:sac`` is registered but the bearer
+cannot be resolved (e.g. a Spartan SLURM compute node running a one-shot
+cohort-solver capsule, where there is no per-node ``sac listen``), this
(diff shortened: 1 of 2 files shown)


ywatanabe1989:

PR #592, opened 2026-07-09
fix(accounts): timer refreshes the active account (total-fleet-stall root cause)

ywatanabe1989:
## Summary

- Root cause of the 2026-07-09/10 total-fleet stall: the host-side `sac.accounts-refresh` systemd timer ran `--all --skip-active`, which was correct under the pre-2026-07-08 two-refresher model but became self-defeating once every agent's credential bind went `:ro` and the timer became the sole refresher. `--skip-active` was starving the one account the whole fleet is pinned to until its ~8h access_token expired and 401'd everything.
- `_jobs_plugin.py`'s JobSpec now emits `--include-active --sync-active-login`; `--skip-active` stays as a tested, explicit opt-in.
- Fixes a live crash the flip exposed: `_iso_ms()` passed the `datetime.timezone` class instead of `timezone.utc`, which raised on the "still fresh, skip" branch and aborted the whole `--all` run (invisible only because `--skip-active` never reached a fresh account).
- Adds the three fields the 2026-06-23 explicit-placement directive requires (`host`, `apptainer.binds`, `restart.max_retries`) to the in-repo `sdk-test` smoke spec, which had been failing `sac agents list` validation since that directive landed.
- Excludes self-peer registration markers (`agents/self/spec.yaml`) from `sac agents list`'s defined-agent validation. That file is deliberately schema-incompatible by design (its own header says not to add `apiVersion`/`spec:`) so `_listen/_self_peers.py` can recognize it as a self-registration marker rather than a launchable Agent -- the generic validator always reported it "invalid" even though it was working as designed.
- Adds `sac agents list --tags` -- a free-form, multi-value lifecycle/status label (e.g. `active-development`), distinct from the ACL-gated `groups` label and from `--capability`. Mirrors the existing `--capability` filtering convention exactly.

Already hand-applied as a host systemd override for immediate mitigation; this PR makes it canonical so a fresh `sac dev systemd install` (or new host provisioning) doesn't regress back to `--skip-active`.

## Test plan
- [x] `tests/scitex_agent_container/cli_pkg/test__account_refresh_include_active.py` (new, 8 tests) — still-fresh branch no longer raises, fresh account no longer aborts a later stale refresh, `--include-active` actually refreshes the pinned account, `--skip-active` still works as opt-in, mutual-exclusion + stderr messaging.
- [x] `tests/scitex_agent_container/cli_pkg/test__account_refresh_pinned_running.py` (existing, 16 tests) — unaffected, still green.
- [x] `tests/scitex_agent_container/test__jobs_plugin.py` (updated, 8 tests) — JobSpec command pinned to the new flags, guard against `--skip-active` reappearing.
- [x] `tests/scitex_agent_container/cli_pkg/_helpers/test__agent_list.py` (extended, 13 new tests total) — self-peer marker classification, sibling agents still discovered, tags filter (both registered + defined-on-disk paths, OR-matching, untagged exclusion).
- [x] `tests/scitex_agent_container/_listen/test__self_peer*.py` (35 tests) — unaffected, still green.
- [x] `tests/scitex_agent_container/cli_pkg/test_status_cmds.py` (42 tests) — unaffected, still green.
- [x] CI is the authoritative gate for the full suite.


ywatanabe1989:
Reopening to force a pull_request-context CI trigger — the required pytest-matrix checks never fired (opened against the wrong base, then retargeted via an edit that GitHub doesn't treat as a synchronize/reopened event).

PR #610, opened 2026-07-10
fix(account,creds,runtimes): incident 2026-07-10 — dead OAuth endpoint root cause, honest failure classes, loud refresh alarms, self-diagnosing picker, :rw credential bind

ywatanabe1989:
## Incident

`incident-account-pool-all-expired-boot-failure-20260710` — 2026-07-10 ~23:2x JST: `sac agents start --group infra` failed for all 5 agents with `NoHealthyAccountError`; all 3 stored accounts read EXPIRED; the refresh timer had been failing silently for hours.

## Root cause (empirically verified, live)

Anthropic **moved the OAuth token endpoint**. The hardcoded `https://console.anthropic.com/v1/oauth/token` ("verified live 2026-05-30") now returns **HTTP 404 `not_found_error` for every refresh_token grant** — the exact body in the timer log for every "dead" account. Probe from this host (dummy token):

| endpoint | answer |
|---|---|
| `console.anthropic.com/v1/oauth/token` | **404 not_found_error** (gone) |
| `platform.claude.com/v1/oauth/token` | **400 invalid_grant** (alive, evaluates the grant) |

The old handler collapsed every failure into `None`, and the CLI labelled it *"refresh endpoint rejected the refresh_token — needs `claude /login`"* — a dead URL misdiagnosed as dead tokens. Nothing had refreshed headlessly for days; snapshots survived only via live-login captures, then aged out.

**Live validation of the fix:** running `refresh_account_credentials` from this branch against the two "dead" accounts **succeeded with their stored refresh_tokens** — the tokens were never invalid. All three accounts are VALID again (`ywata1989-gmail-com` +8h, `ywatanabe-scitex-ai` +8h, `wyusuuke-gmail-com` +7h). **No operator `claude /login` was needed for this incident.** Rotate-persist (write-back of the rotated refresh_token under flock + atomic rename) was verified present and exercised live.

## Changes (4 separable commits)

1. **Endpoint fix + honest failure classes** (`_account/token_refresh.py`, split out of the 817-line `claude_usage.py`; re-exports keep every importer unchanged)
   - `_TOKEN_URL` → `https://platform.claude.com/v1/oauth/token`, resolved per call, overridable via `$SAC_ANTHROPIC_OAUTH_TOKEN_URL` (next move needs no release)
   - failures classified: `rejected` (invalid_grant → genuinely-dead token, prescribes `claude /login` + `sac accounts save`) vs `transport` (404/5xx/network → message says **"NOT a token problem"** and names the URL + status) vs `response` (2xx without access_token); `failure_kind` rides the result dict
   - re-read-and-retry once on invalid_grant when the on-disk refresh_token changed under us (concurrent-writer rotation)
2. **Loud refresh-failure alerting** (`_account/refresh_alarm.py`, hooked into every `sac accounts refresh` invocation — the systemd timer runs exactly this CLI)
   - existing rails only: ADR-0013 lead `blocker` push (`push_to_lead`), falling back to the scitex-todo BLOCKING-YOU help card (`scitex_todo._help_wait.help_wait`, card `help-accounts-refresh-<account>-waiting`) since this fleet currently has **no `lead:` block configured**
   - alert = account name + classified error + canonical recovery line; dedupe via `runtime/refresh-alarm-state.json`; recovery re-arms (recover-then-die alerts again); failed delivery records nothing (retried next run) and prints loudly
3. **Self-diagnosing `NoHealthyAccountError`** (`_creds/_pick_healthy.py`, `_account/creds_sync.py`)
   - `AccountHealth` carries `snapshot_path` + `expires_at_raw` (one parse supplies judgment and evidence); the error quotes per account: file path, literal `expiresAt`, UTC rendering, and the probe timestamp
   - regression locks: a future `expiresAt` can never read EXPIRED / never raises
4. **Credential bind `:ro` → `:rw`** (operator decision 2026-07-11; both the single-file bind and the `/tmp/sac-claude` dir bind)
   - a read-only bind cannot stop the in-container claude from rotating the token server-side; it only stops the rotation being RECORDED, stranding the snapshot on a dead refresh_token — worst of both worlds. Shared-writable is the normal supported claude shape; flock + atomic write + the invalid_grant re-read-retry cover the narrow residual race. `config/_types.py` already documented the writable model (doc/code divergence resolved toward the doc). Evidence note: today's incident was **not** this path (both stored refresh_tokens were still redeemable — no in-container rotation had consumed them), so the flip removes a standing hazard, not the manifested one. Cross-host single-writer stays the `sac-host-token-broker-single-writer` follow-up.

## FACT 2 — the "false-EXPIRED" read of wyusuuke: resolved, no divergent reader

The picker's read was **correct at read time**; the file changed minutes later:

- creds-watch saved the wyusuuke snapshot at 00:30:07Z with `expires_at=1783672205` (= 08:30:05Z)
- creds-watch itself logged the live cred **"expired 20973 seconds ago"** (= 5.83h ≈ the reported **-5.8h**) at 14:19:38Z — a second reader independently confirming the same pre-repair value
- a fresh login/rotation landed at 14:22:23Z; creds-watch captured it into the snapshot at 14:22:24Z (`expires_at=1783722143` = 22:22:23Z — the +8h state the operator later probed)
- "-5.8h" against 1783672205 pins the picker's read to 14:15–14:21Z (23:15–23:21 JST); the dead pair's "-5.9h" (expiries 08:25/08:28Z) pins the same window — the group start probed *before/straddling* the 14:22:24Z capture, and the terse error hid the evidence needed to see that

Ruled out: ms/s unit bug (would render ±490,000h, not -5.8h); project-scope/`$SCITEX_DIR`/container-home store divergence (the infra specs pin absolute `credentials_files` under the host store and the pool derives `store_dir` from those exact paths; the `/home/agent` store is empty → would read ABSENT, not EXPIRED); quota-cache-as-expiry (quota data only orders fresh candidates, cannot mark EXPIRED); refresher-writes-elsewhere (same `store/<name>/.credentials.json`); dotfiles-git-revert (live snapshots are gitignored). Not fully adjudicable from surviving evidence: a stale mount-namespace view in whatever context ran the start vs. simple invocation-time uncertainty — which is exactly why the error now prints file + raw value + probe time (commit 3): the next occurrence is decidable from one line.

## Tests

- 214 passed across `_creds/` and `runtimes/` (bind flip) plus the new suites: token-refresh classification (404-never-says-login regression lock, invalid_grant classes, env override, rotation retry), refresh-alarm (dedupe / re-arm / retry / recovery), pick-healthy evidence (path + raw value + probe time in the error; future-expiry locks), CLI wiring (real CLI → real no-network failure → help card written into a sandboxed store; cross-run dedupe)
- new hermeticity fixture: the quota-aware pick tests were reading the LIVE fleet `/var/sac/quota-cache.json` bound into agent containers (pre-existing environment-dependent failures, reproduced on unmodified develop in-container); `SAC_QUOTA_CACHE_PATH` is now pinned to an absent tmp file
- `ruff --select F401,F811` clean; `scitex-dev ecosystem audit-all --path <worktree>`: project/skills/mcp-tools/python-apis clean after fixes (removed a stray root junk file; registered `scitex_todo._help_wait` in the PS-140 cross-package gate); `audit-cli: not-auditable: unknown` is pre-existing/environmental (fires identically before this branch); the `tests/develop/test_audit.py` one-shot hits its documented 120s subprocess cap under current host load

## Deploy notes / follow-ups

- the host timer keeps 404-ing until this deploys (host sac is an editable install → next timer tick after merge+pull uses the fixed URL); both revived accounts hold ~8h tokens meanwhile
- the dotfiles **hourly** snapshot-refresh script (`account-snapshots-refresh.log`, the "operator 1a" rail) hits the same dead URL — separate repo, needs the same one-line fix
- `lead:` block is unconfigured fleet-wide, so alert leg 1 is inert until configured; leg 2 (scitex-todo card) delivers today
- July-8 credential backups are **committed in the dotfiles git repo** (`src/.scitex/agent-container/accounts/_backup/.../*.credentials.json` tracked) — security-hygiene follow-up
- `claude_usage.py` is down 817 → 584 lines; step 2 of the split (per-account usage block) is the noted follow-up

### DIFF
--- a/src/scitex_agent_container/_account/claude_usage.py
+++ b/src/scitex_agent_container/_account/claude_usage.py
@@ -31,7 +31,6 @@ returned dict for back-compat with downstream consumers, but are always
 
 from __future__ import annotations
 
-import fcntl
 import json
 import time
 import urllib.error
@@ -39,6 +38,39 @@ import urllib.request
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
+
+# Headless OAuth token rotation lives in ``token_refresh`` (split for the
+# 512-line cap after INCIDENT 2026-07-10 — see that module's docstring).
+# Every moved name is re-exported here (``__all__``) so pre-split
+# importers (``cli_pkg/_account_refresh``, ``mint_token``, tests) keep
+# working unchanged.
+from .token_refresh import (
+    _CLAUDE_CODE_CLIENT_ID,
+    _REFRESH_USER_AGENT,
+    _TOKEN_URL,
+    _USAGE_BETA_HEADER,
+    _credentials_path,
+    _load_json,
+    _read_tokens_at,
+    _refresh_access_token,
+    _refresh_access_token_at,
+    refresh_account_credentials,
+)
+
+__all__ = [
+    "_CLAUDE_CODE_CLIENT_ID",
+    "_REFRESH_USER_AGENT",
+    "_TOKEN_URL",
+    "_USAGE_BETA_HEADER",
+    "_credentials_path",
+    "_load_json",
+    "_read_tokens_at",
+    "_refresh_access_token",
+    "_refresh_access_token_at",
+    "fetch_usage",
+    "fetch_usage_for_credentials",
+    "refresh_account_credentials",
+]
 
 # ---------------------------------------------------------------------------
 # Public return-shape sentinel keys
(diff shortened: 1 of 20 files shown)


ywatanabe1989:
CI triage (coordination note): the earlier matrix red on this PR (run 29106433698) was **diff-local** — 4 failures: two tests still asserting the pre-flip `/tmp/sac-claude:ro` bind (`tests/scitex_agent_container/runtimes/test__apptainer_auth.py`, `tests/integration/test_provider_column_specs.py`) and the two new alarm-wiring tests importing the optional `scitex_todo` peer that leaf CI does not install. Fixed in 52334578 (+ merged origin/develop, resolving the `_pick_healthy` import conflict with #611's `_quota_rank` refactor). **Disjoint from PR #609's failures** (`test__in_sif_broker` / `test__stale_lease` config-validation + audit gate — being fixed in parallel by the #609 owner); develop itself is green on lint/quality/import-smoke with its matrix in flight. No shared-develop fix needed from this PR. Auto-merge is armed; this comment is the who-fixes-what marker.

PR #634, opened 2026-07-12
fix(containers): un-freeze scitex-todo in the SIFs + fail-loud symbol gate

ywatanabe1989:
## The harm

The fleet has been running **scitex-todo 0.7.50** for days. Current is **0.9.4**. Consequences (carded as `incident-fleet-drift-stale-scitex-todo-blocks-6-busiest-agents-20260712`): agents' WIP gates count `deferred` + `cancelled` cards as open and refuse legitimate card creation; agents "diagnose" bugs in peer code that was fixed releases ago.

Both live SIFs probe:

```
ImportError: cannot import name 'WIP_STATUSES' from 'scitex_todo._throughput'
```

## Root cause — structural, not a missed rebuild

scitex-todo is installed at the **:base** layer, and **:scitex inherits base's `/opt/venv-sac`**. A `>=` floor that the inherited version *already satisfies* is a **no-op** — uv audits it and moves on. So the :scitex layer could **never** carry a newer scitex-todo than :base, however often it was rebuilt.

Proof: `sac-scitex.sif` was baked **2026-07-11 21:14**, four hours **after** scitex-todo 0.8.4 published (20:27) — and still shipped the **0.7.50** it inherited from a 2026-07-10 `sac-base.sif`. The floor was satisfied; nothing upgraded; nothing complained.

Bumping the floor alone would have papered over this and let it recur at the next release.

## Changes

- **base**: `scitex-todo[mcp]` floor `0.3.0` → `0.9.4`. Most agent specs boot `sac-base.sif` **directly**, so this layer is the one that matters.
- **scitex**: `--upgrade-package scitex-todo` (uv) / a dedicated `--upgrade` invocation (pip) so the package is re-resolved against PyPI **even when the inherited version satisfies the floor**. This is what breaks the base→scitex freeze.
- **scitex**: carry the `[mcp]` extra on the upgrade, else fastmcp stays pinned to whatever the *old* version's extra resolved.
- **scitex**: drift sweep on the other pinned leaves — `scitex-dev` 0.20.1 → **0.29.0**, `scitex-writer` 2.22.0 → **2.29.0** (`scitex-audio` 0.3.0 / `scitex-cv` 0.2.0 already current). All floors checked against live PyPI — a floor *above* the newest published version makes the build UNSATISFIABLE, which is how a `>=0.7.38` floor bricked this recipe once already.
- **both layers**: a build-time gate that asserts **BY SYMBOL** and dies FATAL if it does not hold.

## The gate

```python
from scitex_todo._throughput import WIP_STATUSES   # pre-0.8.0 dies here
...
if "in_progress" not in WIP_STATUSES: fail
if {"deferred","cancelled","done"} & set(WIP_STATUSES): fail
```

Three deliberate choices:

1. **Symbol, never the version string** — the version string *lies*. An orphaned/duplicate `.dist-info` freezes what `importlib.metadata` reports while the live code differs (open incident). "The SIF says 0.9.4" is not evidence.
2. **Refuses to run if two distributions claim the name** — that is the failure mode that makes the version lie in the first place.
3. **Explicit `sys.exit(1)`, not `assert`** — a bare `assert` is stripped under `python -O`/`PYTHONOPTIMIZE`, which would silently disarm the very gate that exists to end silent staleness.

Re-run in **both** layers: base passing its gate proves nothing about what :scitex ships, because :scitex then mutates the inherited venv (scitex[all] resolve, sac force-reinstall, provider install). Guard the artefact you ship.

## Verification

- Gate exercised **both directions** against real interpreters (no mocks): exit 0 + `WIP_STATUSES: ['in_progress']` on 0.9.4; exit 1 with the incident's exact `ImportError` on 0.7.50.
- Both `%post` blocks pass `dash -n` (apptainer runs them under `/bin/sh`, not bash).

## Rollout

Rebuild is running on the host at low priority. Running agents are untouched — the SIF swap is atomic and they pick up the new image on their next natural restart. No force-restarts.


ywatanabe1989:
## Added scope: the sac source tree, gated by symbol too

Folded in on request — same drift class, second source.

### Correction on the premise

The defs do **not** pin sac to PyPI. Both layers install it from the `%files`-staged source tree:

- `apptainer-base.def:390` → `"/opt/scitex-agent-container-src[openai]"`
- `apptainer-scitex.def:147,216` → `--force-reinstall --no-deps /opt/scitex-agent-container-src`

So the image's sac is **whatever source tree ships the .def**. There is no PyPI pin to bump and **no release to cut** — the branch simply predated #633. Merging `develop` in *is* the fix.

### But the staging has failed silently before

Step 1's `scitex[all]` drags in a transitive `scitex-agent-container` wheel from PyPI, and uv once **skipped** the local install under its version-already-satisfied check — the SIF quietly shipped the **wrong sac tree**. The two-pass force-reinstall is the fix; nothing ever *asserted* it worked. Now something does.

### The gate now probes both packages

```python
from scitex_todo._throughput import WIP_STATUSES                              # scitex-todo #356
from scitex_agent_container.runtimes._apptainer_overlay import ensure_overlay_dirs  # sac #633
```

`ensure_overlay_dirs` is **newer than any published sac release**, which is precisely what makes it a good probe: it can only be present if the local source tree won the install.

### This is also the cleanest proof that version strings cannot be trusted

The fresh rehearsal image reports:

```
sac reported version         : 0.21.13
sac overlay auto-provision   : ..._apptainer_overlay.ensure_overlay_dirs
```

The version says `0.21.13` (the last release in `pyproject.toml`) while the code is develop-HEAD **with** the overlay fix. A version check would have called this image stale. The symbol tells the truth. That is the whole thesis of this PR, demonstrated by accident.

### Rehearsal (real interpreters, no mocks)

| scenario | result |
|---|---|
| scitex-todo 0.9.4 + branch sac | **exit 0** — `WIP_STATUSES: ['in_progress']` + overlay symbol |
| PyPI-wheel sac (wrong tree) | **exit 1** — `ModuleNotFoundError: ..._apptainer_overlay` |
| scitex-todo 0.7.50 (what the fleet runs) | **exit 1** — the incident's exact `ImportError: cannot import name 'WIP_STATUSES'` |

Both `%post` blocks pass `dash -n`.

ywatanabe1989:
## The freeze, reproduced in isolation

Rather than argue the mechanism from the docs, here it is on real interpreters. Simulate the `:base` layer baking whatever was newest that day, then run the **old** `:scitex` line and the **new** one against it.

```
### Simulate the :base layer: bake whatever was newest that day (0.7.50)
  inherited: 0.7.50

### (a) THE BUG — the :scitex layer's OLD line: a floor the inherited version already satisfies
$ uv pip install "scitex-todo>=0.7.32"
Checked 1 package in 49ms                 <-- uv did NOTHING
  after old line: 0.7.50 <-- FROZEN

### (b) THE FIX — the NEW line: -P re-resolves it against PyPI
$ uv pip install --upgrade-package scitex-todo "scitex-todo[mcp]>=0.9.4"
 + uvicorn==0.51.0
 + watchfiles==1.2.0
 + websockets==16.1
  after new line: 0.9.4 ['in_progress']
  fastmcp (from [mcp] extra): 3.4.4
```

**`Checked 1 package in 49ms`** is the whole incident in one line. The old recipe asked uv for `>=0.7.32`, uv looked at the inherited 0.7.50, agreed, and moved on. Every `:scitex` rebuild for the last three days did exactly this and reported success.

Note the third line of the fix: bumping the floor **alone** would not have been enough either — the freeze is in the *no-op*, not in the number. `-P` is what forces the re-resolve.

And the `fastmcp 3.4.4` line is why the `[mcp]` extra rides the upgrade: a plain `scitex-todo` upgrade would have left fastmcp pinned to whatever the **old** version's extra resolved.

ywatanabe1989:
## Built and verified — both layers

Rebaked on the host from **this branch's** recipes (`PYTHONPATH` override — see below), at low priority, detached. ~57 min total, no disruption to running agents.

### The gate fired inside both bakes

```
+ /opt/venv-sac/bin/python /tmp/sac-assert-image-fresh.py
scitex-todo reported version : 0.9.4
scitex-todo WIP_STATUSES     : ['in_progress']
sac reported version         : 0.21.13
sac overlay auto-provision   : scitex_agent_container.runtimes._apptainer_overlay.ensure_overlay_dirs
OK: scitex-todo WIP-gate + sac overlay symbols present and correct.
```

### Verbatim symbol probe of the PUBLISHED SIFs

Not the build log — `apptainer exec` on the shipped artefacts.

```
--- :base   -> sac-base-2026-0713-024755.sif
scitex-todo version : 0.9.4
WIP_STATUSES        : ['in_progress']
in_progress in WIP  : True
deferred in WIP     : False
cancelled in WIP    : False
sac version         : 0.21.13
overlay symbol      : scitex_agent_container.runtimes._apptainer_overlay.ensure_overlay_dirs

--- :scitex -> sac-scitex-2026-0713-031203.sif
scitex-todo version : 0.9.4
WIP_STATUSES        : ['in_progress']
in_progress in WIP  : True
deferred in WIP     : False
cancelled in WIP    : False
sac version         : 0.21.13
overlay symbol      : scitex_agent_container.runtimes._apptainer_overlay.ensure_overlay_dirs
```

The same probe on the **old** SIFs returned `0.7.50` / `WIP_STATUSES: ABSENT` — on **both**.

Drift sweep landed too: `scitex-dev` **0.30.0** (floor floated up), `scitex-writer` **2.29.0**, `scitex` **2.30.8**.

### One caveat worth stating plainly: deployed ≠ merged

The host's `sac` is a **non-editable** install (`~/.env-3.11`), so `sac image build` reads its `.def` files — and stages the sac source tree — out of *site-packages*, not the checkout. I therefore built with `PYTHONPATH` pointed at this worktree so the image came from exactly the code under review.

The consequence: **`develop` still carries the old recipe.** The live SIFs are fixed, but anyone rebuilding before this merges would bake the pre-fix recipe — no `-P`, no gate. The images would not regress to 0.7.50 (base now carries 0.9.4, which satisfies the old floor), but the freeze trap and the gate would both be absent again on the next scitex-todo release. **The merge is what makes this durable.**

### Rollout

Atomic symlink swap; old timestamped SIFs retained. Running agents keep their open image and pick the new one up on their next **natural** restart. Nothing was force-restarted. Rollback is a symlink swap.

ywatanabe1989:
Rescuing this PR — it has been sitting RED with auto-merge armed since 2026-07-12, so it could never fire.

## Why it was red

The blocker was not this PR's change. It was the guard in `tests/integration/test_apptainer_base_def_scitex_todo.py`, which claimed to check for *"a version pin (>=0.3 **or stricter**)"* but implemented it as a literal substring match:

```python
has_version_pin = ">=0.3.0" in block or "==0.3.0" in block or ">=0.3" in block
```

It recognises exactly ONE floor — the stale one. This PR raises the floor to `>=0.9.4`, a **strictly stronger** pin, and the guard failed it with *"scitex-todo install must carry a version pin"*. The guard could not express "or stricter" at all: it read like a pin check and behaved like a pin **freeze**.

That is not incidental to `incident-fleet-drift-stale-scitex-todo` — it is the mechanism. The .def sat at an ancient `>=0.3.0` while the live SIFs silently shipped scitex-todo 0.7.50, and the one guard that claimed to watch the floor was the thing holding it down.

## The fix

Implement what it always claimed: parse the requirement's lower bound and compare it **numerically** against the capability the fleet actually depends on — the WIP-gate fix (scitex-todo #356, first released in 0.8.0), the same capability the .def's section-6c gate asserts BY SYMBOL at build time. Declared floor and asserted symbol can no longer drift apart.

The floor is a **capability, not a snapshot**: any version at-or-above it passes, so raising the pin needs no test edit and produces no false-RED. Below it — or unpinned — fails loud.

I also guarded the guard: parametrised parser tests cover the floor written short (`>=0.8`), a floor with an upper bound, an exact pin, a future major, and the three shapes that must be REJECTED (unpinned, the stale `>=0.3.0`, and the `0.7.50` the SIFs actually shipped).

develop merged in (branch was 40 commits behind). 13/13 green locally; ruff + the PS-140 cross-package gate clean. Watching CI to merge — not walking away on an armed auto-merge.

PR #636, opened 2026-07-12
feat(accounts): readable JST Since, relative reset hints, fix fleet 7d-capacity semantics

ywatanabe1989:
## What

Three operator-requested improvements to `sac accounts list` (2026-07-13), all in that command's render path.

### 1. Readable JST `Since`
The `Since` field printed a raw ISO stamp (`2025-05-30T19:59:34.010055Z`). It now renders as `YYYY-MM-DD HH:MM (JST)` (e.g. `2025-05-31 04:59 (JST)`). The zone is `zoneinfo("Asia/Tokyo")` and the `(JST)` abbreviation is **derived from the tz object** (`strftime("%Z")`), not hard-coded.

### 2. Relative reset hints (5h AND 7d)
The usage-bars reset hints were absolute wall-clocks (`→09:19` / `→Sun 21h`). They now show the time **remaining** until the window resets: `29% (in 4h05m)` for the 5h bar, `66% (in 2d 3h)` for the 7d bar. Applied to 7d too — it was clean via a single adaptive `format_relative_until` helper (days+hours ≥ 1d, else hours+minutes, else minutes; `now` for a past/due reset).

### 3. Fix `Fleet effective utilization` semantics (the important one)
**Current formula** (`_account_usage_bars.py`): per account `effective% = clamp(reset_horizon_hours, 0, 168h)/168h × used_pct_7d`, then the mean across accounts. That reset-horizon weighting is exactly why it printed **15%** while the 7d bars read 17/88/88 (mean ≈ 64%): the two 88% accounts' windows reset soon, so their `frac_before_reset` was small (≈0.14) and each contributed only ~12%, dragging the fleet number to ~15%.

**New**: `Fleet 7d capacity used: 64% (3 accounts)` — the plain arithmetic mean of `used_pct_7d` across accounts that have usage. On 17/88/88 → **64%** (was 15%). It reads as a capacity-planning signal: ≈100% ⇒ saturated ⇒ add an account; low ⇒ over-provisioned ⇒ drop one. Relabelled from "Fleet effective utilization" so the number matches its meaning.

**avg vs total (not ambiguous):** the Anthropic OAuth usage API exposes utilisation *percentages* (`used_pct_7d` ∈ [0,100]); the absolute weekly quota (`limit_tokens_7d`) is `None`, so total-used / total-available is **not computable** from the data. The mean of the per-account percentages is therefore the only well-defined aggregate (and equals total-used / total-available when the accounts share a quota tier). Re-aggregates the same 7d numbers the bars already show — **no new network calls**.

## SSOT
New shared `cli_pkg/_timefmt.py` (`format_jst` + `format_relative_until`) so this surface and the forthcoming `sac agents list` Started-column JST work consume one implementation instead of a hand-rolled private copy (none was merged yet — searched for `zoneinfo`/`Asia/Tokyo`/`JST`/`_timefmt`). The Started-column change should `from ._timefmt import format_jst`.

Removed the now-dead reset-horizon machinery (`effective_utilization_pct`, `fleet_effective_utilization`, `WEEK_HOURS`, `_reset_horizon_hours`) and the absolute reset-hint helpers (`format_reset_hhmm`, `format_reset_day_hour`).

## Tests
No mocks — pure helpers driven with known inputs + real on-disk account fixtures (PA-306 style). **214 passing** across the touched `cli_pkg` tests, including the operator's exact 17/88/88 → 64% case, singular/plural/unavailable line shapes, and an end-to-end relative-hint CLI test (far-future reset ⇒ deterministic `(in ...)`). New `test__timefmt.py` covers the shared helpers (JST conversion incl. the operator's timestamp, date-only, None/unparseable; relative `in 4h05m` / `in 2d 3h` / `in 7m` / `now`).

Verified live against real credentials: `sac accounts list` prints `Since: 2025-05-04 15:41 (JST)`.

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_account_list_cmd.py
+++ b/src/scitex_agent_container/cli_pkg/_account_list_cmd.py
@@ -13,11 +13,11 @@ percentages; the table holds only what the bars cannot express"):
 2. The Stored-accounts table — exactly ``Account | Status | Last
    Update`` (:func:`._account_list_render.render_stored_table`).
 3. The usage-bars block — per-account 5h/7d bars, each percentage
-   carrying its compact reset hint (``29% (→09:19)`` /
-   ``66% (→Sun 21h)``), plus the rolling-window legend when a row has
+   carrying its compact relative reset hint (``29% (in 4h05m)`` /
+   ``66% (in 2d 3h)``), plus the rolling-window legend when a row has
    no cached reset timestamps
    (:mod:`._account_usage_bars` / :func:`._account_list_render.rolling_legend_line`).
-4. The one-line fleet effective-utilization figure.
+4. The one-line fleet 7-day capacity-used figure.
 
 The JSON path (``sac accounts list --json``) is schema-stable and keeps
 ``email_address`` / ``plan_label`` / the raw usage payload for machine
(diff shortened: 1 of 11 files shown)


ywatanabe1989:
Added commit `4e7d15b9`: removed 4 perpetually-skipped `@pytest.mark.skip` stubs in `tests/integration/test_config.py` (for `spec.skills`/`spec.remote`, removed in the v3-realign). These were the **4 pre-existing** `PA-307 §3 STX-TQ001` ("test function has no assertion") violations that were failing the `test_audit_all_clean` audit gate on this PR — unrelated to the accounts-list change, but they block the merge (the rule fires regardless of the skip marker). They exercise removed functionality and can never be un-skipped. Verified locally: the lint hook (same STX-TQ001 rule) no longer flags the file and it collects cleanly (52 tests). The remaining `[§4b] help is free-form` is only a WARN (non-blocking).

PR #594, opened 2026-07-09
ci: migrate to scitex-ai/.github reusable workflows

ywatanabe1989:
## What

Slice of the fleet-wide CI-content-drift remediation (operator-approved, 2026-07-10 incident; pattern piloted at `scitex-ai/scitex-stats#69`). Converts local CI job bodies into thin `workflow_call` stubs against reusable workflows hosted at `scitex-ai/.github@main`, where they clearly match the generic pattern with no repo-specific behavior lost.

## Converted

- **`auto-merge-to-develop.yaml`** (`automerge` job → `call`): body is byte-identical to the generic reusable `auto-merge-to-develop.yml` (confirmed against its actual content), so this is a pure mechanical extraction. `on:`/`permissions:` preserved verbatim. The file **stays on the default branch (`main`)** per GitHub's check_suite-workflow-must-live-on-default-branch requirement — this PR only changes the job body; a human still needs to promote `develop`→`main` per this repo's existing workflow for the change to take effect.

## Left alone (repo diverges materially from the generic reusable workflow — converting would silently drop behavior)

- **`pytest-matrix-on-ubuntu-py3-11-3-12-3-13.yml`** (`tests` job): runs on `ubuntu-latest` (not the reusable workflow's `self-hosted spartan-cpu`), and has three repo-specific steps the generic `pytest-matrix.yml` doesn't support: `fd-find` apt install (required by the mandatory `test_audit.py` gate), Anthropic-credentials provisioning (needed for creds-gated tests to actually run instead of skip), and a template-validity smoke pre-step.
- **`import-smoke-on-ubuntu-py3-12.yml`** (`install-check` job): runs on `ubuntu-latest` (not self-hosted), and has an extra CLI entry-point smoke step (`sac --version`) the generic `import-smoke.yml` doesn't include.
- **`quality-audit-on-ubuntu-latest.yml`** (`audit` job): uses a completely different tool surface (`scitex-dev quality audit-cli/audit-frontmatter/audit-docs/audit-lines/audit-scope`, all `continue-on-error` / warn-only, weekly cadence) vs. the generic `quality-audit.yml`'s single hard-fail `scitex-dev ecosystem audit-all`. Not the same check at all.
[...]

PR #650, opened 2026-07-13
ci(tests): pytest matrix on self-hosted Spartan runners — BLOCKED on 2 runner faults (nproc=1; 1 dispatchable runner)

ywatanabe1989:
> **STATUS: DRAFT — do NOT merge.** Measured end-to-end, this migration is currently
> **SLOWER** than the hosted baseline: **18m54s vs 11m41s**. Two runner-infrastructure
> faults cause that (below). Neither is a defect in this PR — the workflow itself is
> proven correct — but until scitex-hpc fixes them, merging would make CI worse.
> The code is ready and waiting.

## What

Move the **pytest matrix** — and only the pytest matrix — onto the self-hosted Spartan
CPU runners, running inside the prebuilt CI SIF (the recipe sac's own tag-publish
workflow already uses).

## Why only the pytest matrix

`tests` is the *entire* critical path. Median over the last 10 runs of each workflow:

| workflow | median |
|---|---|
| **tests** | **11m41s** |
| docs | 0m34s |
| import-smoke | 0m30s |
| quality | 0m29s |
| lint | 0m11s |

Everything else finishes in <=45s and runs in parallel with `tests`. Moving them would
gain ~0s and add risk.

## MEASURED: this PR's run (`29253369819`) vs baseline

| | | |
|---|---|---|
| baseline (ubuntu-latest, 3 legs **parallel**) | **11m41s** | median |
| this PR (Spartan, 3 legs **serialized**, 1-core runner) | **18m54s** | py3.11 4m37s + py3.13 7m20s + py3.12 6m47s |

**1.6x slower.** The workflow is correct — py3.12 and py3.13 both went green, the SIF step
ran and every github-hosted step correctly skipped — but two infra faults sink the result.

## Fault 1 — the Spartan runner is running on ONE core

Same runner, same `run-in-sif.sh`, two days apart:

```
2026-07-11 (publish run 29150607979):  xdist workers=48 (nproc=48)   9416 passed in 155.38s (2m35s)
2026-07-13 (this PR      29253369819):  xdist workers=4  (nproc=1)   9696 passed in 397.12s (6m37s)
```

`nproc=1`. The suite is the same size (+3% tests). `spartan-cpu-scitex-agent-container-01`
has lost its multi-core SLURM allocation — the runner `srun --overlap`s into an
already-running lease, and that lease is currently handing it **1 CPU**. `run-in-sif.sh`
floors xdist at 4 workers, so 4 workers now fight over 1 core.

**On 48 cores this suite takes 2m35s. On 1 core it takes 6m37s.** The hardware is genuinely
~4.5x faster than a hosted runner when it is allocated correctly — that is not speculation,
it happened on 2026-07-11.

## Fault 2 — only ONE dispatchable runner, so the matrix serializes

```
gh api /orgs/scitex-ai/actions/runner-groups
  group id=1 'Default'      allows_public_repositories=FALSE
  group id=3 'spartan-cpu'  allows_public_repositories=FALSE
```

**scitex-agent-container is a public repo.** Both org runner groups refuse public repos, so
`spartan-cpu-org-01`, `spartan-cpu-org-02` and `spartan-pooled-cpu-01` sit `online, busy=false`
while sac's jobs queue. sac can only use its own repo-level runner. Observed here — all three
legs on the one runner, py3.12 queued 12 minutes:

```
py3.11  13:20:07 -> 13:24:44   spartan-cpu-scitex-agent-container-01
py3.13  13:24:50 -> 13:32:10   spartan-cpu-scitex-agent-container-01   (starts 6s after 3.11 ends)
py3.12  13:32:14 -> 13:39:01   spartan-cpu-scitex-agent-container-01   (queued since 13:20:04)
```

`run-in-sif.sh` already documents the intended topology — *"one runner per node:
scitex-agent-container-01/02/03"* — so `-02` and `-03` are simply missing.

## The ask (scitex-hpc) — both are needed for the win

1. **Restore the multi-core SLURM lease** for `spartan-cpu-scitex-agent-container-01`
   (`nproc` should be 48, currently 1).
2. **Register the two missing repo-level runners** (`-02`, `-03`) so the three matrix legs
   run in parallel. (Alternative: set `allows_public_repositories: true` on the `spartan-cpu`
   org group — but that is org-wide and lets *any* public repo onto Spartan, so restoring the
   repo-level runners is safer.)

Projected once both land: per-leg ~3m, three legs in parallel -> **~3m vs 11m41s (~3.9x)**.
With only one fixed: ~9m (cores only) or ~7m20s (runners only) — both marginal.

## What this PR gets right regardless

- **`timeout-minutes: 30`** — the guard that was MISSING. With no cap, a hung leg burned
  **1h39m42s** in run `29052895608` before the platform killed it.
- **Job names kept verbatim.** `pytest-matrix-on-ubuntu-py3.{11,12,13}` are required status
  checks on **both `develop` and `main`**; renaming them (or moving the job under a reusable
  workflow, which re-renders checks as `<caller-job> / <name>`) would leave all three
  permanently pending and block every PR in the repo.
- **Fork PRs pinned to `ubuntu-latest`** unconditionally — untrusted code never runs on shared HPC.
- **`Provision Anthropic credentials` is github-hosted only**, deliberately: the Spartan runners
  share the operator's real `$HOME`, and writing `~/.claude/` there would clobber live agent
  credentials on the HPC box.

## One-step revert

```bash
gh variable set CI_TESTS_RUNS_ON --repo scitex-ai/scitex-agent-container --body '["ubuntu-latest"]'
```

Undo: `gh variable delete CI_TESTS_RUNS_ON`. Git-level: `git revert fa96db24`.

The escape hatch is **real, not decorative**: the previous github-hosted step path is retained
verbatim behind `runner.environment`, so flipping the variable lands CI on a known-green path
rather than red. A dedicated `CI_TESTS_RUNS_ON` (not reusing `CI_RUNS_ON`) means reverting the
tests does not drag the **publish** workflow back onto hosted runners, where it cannot build.

## Why not the `scitex-ai/.github` reusable workflow

Not adoptable today: (1) it would rename all three required status checks and block every PR;
(2) its bare-runner `uv venv` install is documented *in this repo* to fail on Spartan
(`actions/setup-python` -> *"version not found for this OS"*; no passwordless apt for `fd-find`,
without which the mandatory `tests/develop/test_audit.py` gate hard-crashes); (3) `on:
workflow_call: {}` with zero inputs and a hardcoded `runs-on` = no escape hatch; (4) it is a
stale draft nobody consumes — its header cites scitex-stats' `ci.yml`, which no longer exists,
and scitex-stats' real matrix runs the **SIF** pattern, not the SSOT's uv pattern.

Worth a follow-up in `scitex-ai/.github`: give it a `runs-on` input and align it with the SIF
pattern, then repos can actually consume it.

## The 1h39m42s job in #594 was never on a self-hosted runner

`pytest-matrix-on-ubuntu-py3.11` in run `29052895608` reports `runner=GitHub Actions 1000150225`,
`labels=ubuntu-latest` — **GitHub-hosted**. Conclusion `cancelled`; failure annotation literally
**`repository transferred`** (the repo moved `ywatanabe1989/…` -> `scitex-ai/…` mid-run); the two
sibling legs carry the matching `git ... exit code 128`. The SSOT workflow it supposedly tested was
authored **2026-07-10**, the day *after* that run. The self-hosted runners were never exercised.

## The red py3.11 leg here is a GitHub outage, not this change

`Set up job` failed with `Failed to resolve action download info. Error: Service Unavailable`.
The same error is failing `lint` and `import-smoke` on **`develop`** right now (run `29252868295`,
job `ruff-on-ubuntu-latest`, a **GitHub-hosted** runner, different SHA). py3.12 and py3.13 both went
green on the same Spartan runner minutes later.

## Second commit — `newb` removal (droppable)

`59ab2992` deletes `newb-docs-quality-on-ubuntu-latest.yml` per the standing decision. Correcting
the record: it was *not* adding to the PR queue — weekly cron + `workflow_dispatch` only, never on
push/PR, not a required check. Separate commit so it can be dropped independently.


ywatanabe1989:

PR #673, opened 2026-07-14
fix(listen): sac listen span forever behind a holder it never really checked, and published a turn_url that could not connect

ywatanabe1989:
Two real bugs in `sac listen`, the host control plane. Both were hurting the operator right now. 「このインフラがぶれると死活問題」

---

### 1a. A HEALTHY holder made `sac listen` spin forever (operator, verbatim)

```
$ sac listen
# sac listen: standing by behind healthy holder PID 745734 on 127.0.0.1:7878
# sac listen: standing by behind healthy holder PID 745734 on 127.0.0.1:7878
   ... (forever) ...
^C# sac listen: shutdown signal received while standing by — exiting without binding
$ kill 745734                      <-- HE HAD TO DO THIS BY HAND
$ sac listen
# sac listen: holder PID 823966 not answering health (1/2) — re-checking in 4.0s before taking over
# sac listen: holder PID 823966 is wedged (not serving) — taking over
# sac listen: acquired 127.0.0.1:7878 — serving
```

> 「なぜ私がいちいち手でキルしないといけないんでしたっけ そこは自動化できてないんですか」

**"Someone else is already serving" is a SUCCESS, not a reason to poll every 4s until the human gives up.** Standing by bought nothing — a standby holds no lock and serves no request — and the spinner was the only way out, so he Ctrl-C'd and killed the holder by hand. Every abandoned spinner left junk behind (two orphaned `sac-listen` screen sessions were found this way; the flock guard held, only one ever bound — the *spinner* was the junk).

Taking over a healthy holder would be worse than useless: a listen restart tears down the in-process a2a `Broker` and **deafens every agent's inbox at once**.

### 1b. A DEAD holder was announced as "healthy" — forever

```
# sac listen: holder PID 738982 on 127.0.0.1:7878 not answering health (1/2) — re-checking in 4.0s before taking over
# sac listen: standing by behind healthy holder PID 738982 on 127.0.0.1:7878
# sac listen: standing by behind healthy holder PID 738982 on 127.0.0.1:7878
... (forever)
```

…while PID 738982 answered nothing and the fleet was cut off from the host. Note the shape: it **SAW** the failed check, then went back to calling the same holder healthy.

**Root cause, reproduced verbatim in a test:**

```python
if _probe_health(...):
    consecutive_unhealthy = 0          # <-- the whole record, gone
    _log("standing by behind healthy holder PID …")
```

A single lucky reply erased the record of the failed check, so a **flapping** holder could never accumulate the *consecutive* failures the take-over threshold demanded. It oscillated `1/2` → "healthy" → `1/2` → "healthy" forever.

Flapping is not hypothetical — I measured the live 7878 daemon on this box answering `HTTP 200` one minute and `Connection refused` five minutes later.

Second cause: the probe was `_http_get(...) > 0`, so **any** HTTP status — including a 5xx from a daemon whose health route is erroring — read as a healthy holder.

### What a failed health check does now

The startup path is a **bounded decision** with exactly three outcomes and no "spin":

| Holder state | Action |
|---|---|
| **Nobody holding** (or holder died) | take the flock, heal any untracked port remnant, **BIND** |
| **Verified SERVING** (answers `/v1/health`) | `# sac listen: already serving: PID N on 127.0.0.1:7878 (/v1/health answered HTTP 200) — nothing to do` → **EXIT 0 immediately.** One probe. No sleep. No loop. |
| **NOT SERVING**, corroborated | **SIGTERM only** → holder exits → re-acquire → bind |
| **NOT SERVING** and won't leave | **FAIL LOUD**, exit non-zero, naming the PID + the exact command |

Health is now a three-state verdict (`_holder_health.py`), never a bool:

- `SERVING` — it answered `/v1/health`.
- `NOT_SERVING` — it answered, with a 5xx. An answer, but not health.
- `UNREACHABLE` — **"I asked and got nothing."** This is the state two states could not express. It is NOT health, and it is never logged as one.

**A failure is a fact; one later success does not un-happen it** (`_standby_ledger.py`). A failed check is cleared only by a *sustained* serving streak, and that recovery is logged loudly. A holder that blips once is not destroyed; a holder that keeps failing is always eventually acted on.

**Never destroy on a negative signal.** The automatic take-over sends **SIGTERM and nothing else** — it never escalates. A probe-based "wedged" verdict can be wrong, and the remedy for a wrong one is to SIGKILL a healthy control plane. The false-RED is strictly worse than the false-green. A holder that ignores SIGTERM is handed to the human:

```
ERROR: sac listen cannot take over 127.0.0.1:7878.
  The flock holder PID 738982 FAILED its health check 2x (/v1/health did not answer
  AT ALL (connection refused / timed out) — no evidence of health), so it is NOT
  serving — but it also did not exit on SIGTERM after 3 take-over attempt(s): …
  Refusing to SIGKILL it automatically: a probe-based 'wedged' verdict can be wrong,
  and force-killing a healthy control plane would cut the whole fleet off from this host.
  Refusing to stand by behind it either: it is not serving, so waiting silently would
  hide the outage indefinitely.
  To force the take-over, run:  sac listen restart --force
  To inspect the holder first:  sac listen status  /  lsof -nP -iTCP:7878 -sTCP:LISTEN
```

Also fixed: the take-over **no longer unlinks a live holder's pidfile**. That broke the flock's own atomic-bind invariant — a racing standby flocking the OLD inode plus a fresh acquirer creating a NEW one would *both* believe they held the lock.

### systemd interaction (`Restart=always`) — checked, as asked

`scripts/systemd/sac-listen.service` has `Restart=always`, `RestartSec=5s`, `StartLimitIntervalSec=60s`, `StartLimitBurst=5`. Its comment says `Restart=always` was chosen **deliberately** over `on-failure` because "a listen that exits 0 for ANY reason … must STILL come back. `Restart=on-failure` does NOT cover a 0-exit, which is exactly how the fleet lost a2a comms silently."

**So yes: `Restart=always` restarts even on exit 0.** The new "already serving" 0-exit *is* retried under the unit. But the churn is **bounded and self-limiting**, and it does not produce a new spin loop:

- `RestartSec=5s` + `StartLimitBurst=5`/`60s` ⇒ at most **5 restarts (~25s)**, then the unit parks in `failed` — visible in `systemctl --user status sac-listen`. Not a hot loop.
- The companion `sac-listen-health.timer` does **not** resurrect it: its probe is **port-based** (`curl http://127.0.0.1:7878/v1/health`), so while the foreign instance is serving, the probe is **green** and the timer neither alarms nor restarts. **Fleet comms stay up throughout.**

A duplicate `sac listen` is a configuration problem; the unit now surfaces it as `failed` while the port keeps being served. Documented in the unit file. If that `failed` state ever becomes a nuisance, the clean remedy is a dedicated exit code + `RestartPreventExitStatus=` — **not** reverting to `Restart=on-failure`, which would reopen the silent-0-exit outage of 2026-06-26. I did not change the restart policy: it was a deliberate, incident-driven choice and is out of scope here.

---

## Bug 2 — the derived `turn_url` was unreachable, 100% of the time

Reported by `claude-code-telegrammer`, independently re-measured on this box:

```
$ getent hosts ywata-note-win
127.0.1.1   ywata-note-win.localdomain ywata-note-win      <-- .1.1, not .0.1

127.0.0.1:19017        -> OPEN
ywata-note-win:19017   -> CONNECTION REFUSED
```

My own measurement (same network namespace, live 7878):

```
resolve ywata-note-win -> ['127.0.1.1']
TCP 127.0.0.1:7878      OPEN
TCP ywata-note-win:7878 ConnectionRefusedError: [Errno 111] Connection refused
GET http://127.0.0.1:7878/v1/health      -> status 200
GET http://ywata-note-win:7878/v1/health -> URLError: [Errno 111] Connection refused
```

`derive_turn_url()` built `http://<canonical-hostname>:<port>/v1/turn`. The a2a sidecar binds `127.0.0.1` (`a2a/_server.py:555`: `host: str = "127.0.0.1"`). The machine's own hostname resolves to `127.0.1.1` — the stock Debian/Ubuntu/WSL self-host convention. **Different loopback address ⇒ refused, deterministically, for every local consumer.**

### Chosen fix: derive `127.0.0.1`, do NOT rebind the sidecar to `0.0.0.0`

Binding every agent's sidecar to `0.0.0.0` would expose `/v1/turn` — an endpoint that **injects a prompt into a live Claude session** — on every interface of the host. sac's transport doctrine is loopback-only with orochi's tunnel mesh as the sanctioned external path (`SAC_OROCHI_SCOPES.md §4.4`); `sac listen` itself refuses a non-loopback bind without `--allow-non-loopback`. Fixing an address-derivation bug by widening a listener's exposure trades a connectivity bug for a security regression. Deriving the address the sidecar *actually binds* costs nothing and changes no listener.

A host that names **this machine** normalises to `127.0.0.1`. A genuinely **remote** host keeps its name. The predicate (`_local_host.py`) treats **any** loopback address as local — a naive `host == "127.0.0.1"` check would miss `127.0.1.1` and happily emit the dead URL again.

### Every consumer of `derive_turn_url`, as asked

| Consumer | Affected? |
|---|---|
| `_listen/server.py:161` → `GET /agents` (`enrich_row`) | **YES** — published the dead URL |
| `_listen/server.py:229` → `GET /agents/<name>/status` | **YES** — same |
| `cli_pkg/_send_broker.py:277` `_host_from_turn_url()` → **`sac agents send`** | **YES, load-bearing** — see below |
| `cli_pkg/a2a_group.py:557` → `sac a2a peers` table | display only |
| scitex-todo's board (external) — POSTs directly to `turn_url` | **YES** — this is the `Connection refused [Errno 111]` that `_listen/_notify.py`'s docstring blames on *container isolation* |
| `_mcp/channel.py` wake-on-push (cct's suspect) | **NO** — see below |

**Were card-event wakes affected? NO.** `_mcp/channel.py`'s `turn_url` does not come from `derive_turn_url` at all — it arrives via `sac mcp channel --turn-url`, which `runtimes/_sdk_channels.py:241` builds as a **hardcoded** `f"http://127.0.0.1:{a2a_port}/v1/turn"`. Likewise `_listen/_card_event_delivery.py` POSTs to `SAC_LISTEN_BASE_URL` (`http://127.0.0.1:7878`) `/v1/notify`, also loopback. cct's suspicion is not borne out — sac's own card-event rail was never on this URL.

**Worth flagging (not fixed here):** `_listen/_notify.py`'s docstring attributes the board's `Connection refused` to "the agent's a2a port is NOT reachable inbound from the board's host context" (container isolation). That diagnosis looks wrong: `127.0.0.1:<a2a_port>` **is** open from the host (cct measured `127.0.0.1:19017 -> OPEN`; I reached the host's 7878 from inside a container). The refusal matches the `127.0.1.1` derivation, not container isolation. The `/v1/notify` rail works and is unchanged — but the "interim workaround" may have been built on a misdiagnosis. Follow-up.

### Blast radius caught: `sac agents send` would have regressed onto ssh

`cli_pkg/_send.py:198` classified local-vs-remote by **string-comparing the host**:

```python
peer_host = endpoint.host if endpoint.host != current_host else ""
...
if peer_host and peer_host != current_host:
    url = f"ssh://{peer_host}:{a2a_port}/v1/turn"
```

The brokered lookup an in-container agent uses parses the registry's `turn_url` back into a host (`_send_broker._host_from_turn_url`). Once that URL correctly says `127.0.0.1`, a bare `!= current_host` would have classified the loopback literal as a **remote peer and tried to `ssh://127.0.0.1:<port>`** — turning a URL fix into a comms outage. The classification is now loopback-aware, with tests in both directions (loopback stays local; a real remote still routes over ssh).

---

## Tests — made to fail first

Both suites were run against the **unfixed** code first: **16 failed**. The flapping-holder test reproduced the operator's log verbatim:

```
'# sac listen: holder PID 738982 ... not answering health (1/2) — re-checking ...'
'# sac listen: standing by behind healthy holder PID 738982 on 127.0.0.1:40982'
'# sac listen: holder PID 738982 ... not answering health (1/2) — re-checking ...'
'# sac listen: standing by behind healthy holder PID 738982 on 127.0.0.1:40982'
```

and the turn_url tests failed with:

```
AssertionError: the turn_url derived for a LOCAL agent from this machine's own
hostname ('ywata-note-win') does not connect to the sidecar bound on 127.0.0.1
```

**No mocks.** Every holder is a REAL HTTP server on a REAL ephemeral loopback socket (never 7878). The "never SIGKILL" guarantee is proved against a REAL child process that installs `SIG_IGN` for SIGTERM — the only way it could end up dead is a SIGKILL, so the assertion that it survives *is* the proof. `derive_turn_url`'s output is asserted by actually `connect()`-ing to a real socket bound exactly like the sidecar; a string-shape assertion would have passed happily throughout the outage.

New: `test__standby_false_green.py`, `test__standby_ledger.py`, `test__standby_signal.py`, `test__holder_health.py`, `test__local_host.py`, `test__graceful_stop.py`, `test__turn_url_reachability.py`, `test__send_loopback_dispatch.py`.

One pre-existing expectation was updated — `test_enrich_row_with_endpoint_adds_both_fields_when_sources_present` asserted `http://test-host:21000/v1/turn`, where `test-host` is the fixture's *own canonical host*. That expectation encoded the bug.

The live 7878 control plane was never touched.

### DIFF
--- a/scripts/systemd/sac-listen.service
+++ b/scripts/systemd/sac-listen.service
@@ -57,6 +57,28 @@ StandardError=journal
 # restart can't double-bind the port; the kernel releases the flock
 # on every dirty exit, so the next start cleanly takes it.
 Restart=always
+#
+# INTERACTION WITH THE "already serving" 0-EXIT (read before changing this).
+# `sac listen` now EXITS 0 immediately when ANOTHER instance is already
+# VERIFIED SERVING on the port ("already serving: PID N ... nothing to do")
+# instead of standing by in a poll loop forever. Restart=always restarts
+# even on a 0-exit — deliberately, see above — so under this unit that
+# 0-exit IS retried. The churn is BOUNDED and self-limiting, by design:
+#   * RestartSec=5s + StartLimitBurst=5 / StartLimitIntervalSec=60s ⇒ at
+#     most 5 restarts (~25s) and then the unit parks in `failed`, which is
+#     VISIBLE (`systemctl --user status sac-listen`). It is NOT a hot loop.
+#   * The companion sac-listen-health.timer does NOT resurrect that loop:
+#     its probe is PORT-based (`curl http://127.0.0.1:7878/v1/health`), so
+#     while the foreign instance is serving, the probe is GREEN and the
+#     timer neither alarms nor restarts. Fleet comms stay up throughout.
+# In other words: a duplicate `sac listen` is a configuration problem, and
+# the unit surfaces it as `failed` while the port keeps being served. The
+# fix is to stop the foreign instance, not to make this unit spin.
+# If that `failed` state ever becomes a nuisance, the CLEAN remedy is a
+# dedicated exit code for "already serving" plus `RestartPreventExitStatus=`
+# — NOT reverting to Restart=on-failure (which would reopen the silent
+# 0-exit outage of 2026-06-26).
+#
 # Brief debounce: avoid hot-looping a launch that always fails (e.g.
 # uvicorn import error after a bad pip upgrade). 5s is the same
 # value scitex-orochi's persistent-services use.
--- a/src/scitex_agent_container/_listen/_graceful_stop.py
+++ b/src/scitex_agent_container/_listen/_graceful_stop.py
@@ -1,0 +1,79 @@
+"""SIGTERM-only stop primitive — the automatic path may ASK, never destroy.
+
+Why this exists alongside ``_restart._terminate_then_kill``
+==========================================================
+``_terminate_then_kill`` escalates SIGTERM → SIGKILL after a grace
+window. That is correct for ``sac listen restart``: an operator typed
+the command, so escalating is *sanctioned destruction*.
+
+The hot-standby loop (:mod:`._standby`) is a different animal. It acts
+on a **probe-based inference** that the holder is wedged — and a probe
+can be wrong. A 2-second health timeout under load, a momentary stall,
+a GC pause: any of these can read as "not answering". The remedy for a
+wrong verdict there is to SIGKILL a perfectly healthy control plane and
+cut the entire fleet off from the host.
+
+    A false-RED is strictly worse than a false-green: the false-green
+    merely fails to act, while the false-RED actively destroys the
+    working thing it misdiagnosed.
+
+So the automatic take-over may only ever ASK the holder to leave. If the
+holder ignores SIGTERM, the standby loop FAILS LOUD and names the one
+destructive command a HUMAN can authorise (``sac listen restart
+--force``). No SIGKILL is ever sent on the strength of a probe.
+
+Seams (``_kill`` / ``_sleep`` / ``_alive``) are module-level swappable
+callables per PA-306 / STX-NM001-003 — no MagicMock.
+"""
+
+from __future__ import annotations
+
+import os
+import signal
+import time
+from typing import Callable
+
+from ._restart import pid_alive
+
+__all__ = ["DEFAULT_POLL_INTERVAL_SECS", "terminate_gracefully"]
+
+DEFAULT_POLL_INTERVAL_SECS: float = 0.2
+
+_kill: Callable[[int, int], None] = os.kill
+_sleep: Callable[[float], None] = time.sleep
+_alive: Callable[[int], bool] = pid_alive
+
+
+def terminate_gracefully(
+    pid: int,
+    *,
+    grace_secs: float,
+    poll_interval: float = DEFAULT_POLL_INTERVAL_SECS,
+) -> bool:
+    """SIGTERM ``pid``, poll for its exit, and **NEVER escalate**.
+
+    Returns ``True`` iff the process is GONE by the time we stop waiting
+    (including the case where it was already gone — that is the goal
+    state, not a failure). Returns ``False`` when it ignored SIGTERM and
+    is still alive; the caller must then surface a loud, actionable
+    failure rather than reach for SIGKILL.
+
+    Bounded by ``grace_secs`` — this must never become an unbounded wait.
+    """
+    try:
+        _kill(pid, signal.SIGTERM)
+    except ProcessLookupError:
+        return True  # already gone — the goal state
+    except PermissionError:
+        # Someone else's process. We cannot stop it and we must not
+        # pretend we did; the caller fails loud with the pid.
+        return False
+
+    remaining = grace_secs
+    step = poll_interval if poll_interval > 0 else DEFAULT_POLL_INTERVAL_SECS
+    while remaining > 0:
+        _sleep(step)
+        remaining -= step
+        if not _alive(pid):
+            return True
+    return not _alive(pid)
--- a/src/scitex_agent_container/_listen/_holder_health.py
+++ b/src/scitex_agent_container/_listen/_holder_health.py
@@ -1,0 +1,154 @@
+"""Three-state health verdict for the ``sac listen`` flock holder.
+
+Why three states (the false-green this closes)
+==============================================
+The standby loop used to ask ``_probe_health() -> bool``. A two-state
+predicate cannot distinguish the two things an operator MUST be able to
+tell apart:
+
+* "the holder ANSWERED ``/v1/health``"  — an OBSERVATION, and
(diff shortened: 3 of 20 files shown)


ywatanabe1989:

PR #677, opened 2026-07-14
feat(freshness): sac warns on EVERY invocation when it is not what shipped — ghost tags, stale installs, stale daemons

ywatanabe1989:
## Why

**We are not failing to FIX these bugs. We are failing to SHIP them.**

On 2026-07-13 seven PRs fixed the `sac listen` wedge (#658, #663, #644, #656, #657, #661, #662). All merged. The host ran **0.21.14 the entire time — which predates every one of them.** Agents hand-drained the wedged pool and re-diagnosed the same already-fixed bug for a day, because three consecutive tags never reached PyPI and **nothing anywhere said a word**.

A "ghost tag" is the exact failure mode: `git tag` succeeds, the release workflow dies, and nothing alarms. The tag exists, so it *looks* released.

The drift is **structural, not bad luck.** A release is `tag → test → build → publish`. The tag is a local action that always succeeds; everything downstream is async and can fail silently. **No consumer ever compared what it was RUNNING against what actually SHIPPED**, so drift was monotonic — it could only accumulate, never self-correct.

This closes the loop at the consumer.

## What this does

Every `sac` invocation now warns on **stderr** when the install is behind PyPI, naming the exact fix:

```
$ sac --version
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
sac-freshness WARNING: this sac is not what shipped.
  * installed 0.21.13 is BEHIND PyPI 0.21.17
      fix: /opt/venv-sac/bin/python -m pip install -U 'scitex-agent-container==0.21.17'
  (silence: SAC_FRESHNESS_QUIET=1   details: sac freshness check)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Typing `sac` is how you find out. No README required — a doc nobody opens is not a control.

## Four INDEPENDENT detections

Each is blind to the other three. On 2026-07-13 **all four were true at once** and the fleet saw none of them.

| check | detects |
|---|---|
| `ghost-tag` | a tag exists, PyPI has no such release |
| `host-behind-pypi` | the install is older than the newest release |
| `running-vs-installed` | a live daemon still executes **pre-upgrade code** (installed is NOT running — Python never reloads a module in a live process) |
| `release-run` | the release workflow ended in `failure`/`cancelled` |

### It found EIGHT ghost tags, not the two anyone remembered

`v0.6.0`, `v0.17.1`, `v0.21.6`, `v0.21.8`, `v0.21.10`, `v0.21.12`, **`v0.21.15`**, **`v0.21.16`**

A third of the 0.21 line never shipped. That is the real answer to 「これ何回目？」.

## Design constraints

- **SYMBOL-PROBE, NEVER THE VERSION STRING.** The number lies in both directions here (stale in memory / stale on disk / fossil `.dist-info`). `_symbols` asks `hasattr(mod, name)` of the module *actually loaded*, so a fix is **provable** rather than claimed. Seeded with #658's `_off_loop.abandoned_call_count`.
- **PyPI is the only truth** about what shipped. Not the tag, not the GitHub release, not the changelog — all three lied during the incident.
- **Three states: FRESH / STALE / UNKNOWN.** Only STALE ever speaks. UNKNOWN is **silent** — never "fine", and never a remedy either, because a false RED gets acted on and the action destroys a healthy thing.
- **No auto-remedy.** Never upgrades, never restarts. Surfaces only.
- **Never slow, never fatal.** Cron does the network and writes a cache; the CLI reads that cache and nothing else. Any failure degrades to silence — a staleness warning that can break the CLI is worse than the staleness it reports.

## Performance (measured, not asserted)

- `warn_if_stale()`: **1.4 ms** median (p95 4.2 ms) on a load-60 host
- imports added: **~7.5 ms** on a ~150 ms CLI budget
- `_freshness/__init__` is lazy (PEP 562): the hot path pulls in **NO** urllib / subprocess / scitex_config — verified

`scitex_config._ecosystem` (the canonical path resolver) costs **1.59 s** to import — measured — so `_cache` mirrors its one-line `$SCITEX_DIR` contract in stdlib instead. Importing it would make every `sac` call ~10x slower, which is a guaranteed revert. **The equivalence is pinned by a test** (`test_cache_path_matches_ecosystem_resolver`) so the mirror cannot silently fork.

## Hook point

`cli_entry_point`, **not** `main`'s group callback: click's eager flags skip the callback, so `sac --version` — the very command you run to ask "am I current?" — would have been the one invocation that never answered it.

## Knobs

- `SAC_FRESHNESS_QUIET=1` — silence (precedent: `SCITEX_DEV_LINTER_QUIET`)
- `SAC_FRESHNESS_SEVERITY=silent|warn|error` — the single knob to escalate warn → hard exit (exit 3) once the signal has earned that trust

## Tests

67 new, driving **real seams with real recorded evidence** (no mocks): the actual PyPI release list, the actual tag list, the actual `gh run` conclusions.

The centrepiece — `test_would_have_caught_the_incident_as_stale` — replays the fleet exactly as it stood at 2026-07-13 23:30 (head tag `v0.21.16`, its run FAILED, PyPI still 0.21.14) and asserts the alarm fires. **It proves this would have caught the incident.**

152 passed across `test_cli` + `_provenance` + `_freshness` — no regressions.

## Verified live

`sac freshness refresh` run against **real PyPI / real git tags / real gh runs** correctly reports `host-behind-pypi` STALE (0.21.13 → 0.21.17), names all 8 ghost tags, returns `running-vs-installed` as **UNKNOWN** (no systemd in-container — correctly *not* "fine"), and confirms #658's symbol is present in the loaded code.

## Follow-up (not in this PR)

The cron line is host-only (`~/.dotfiles/.crontab_list`, dotfiles-owned). The script is deployed to `~/.scitex/agent-container/bin/deploy-freshness-alarm.py`; it needs one hourly crontab entry to go live. Until then the alarm is inert but the CLI warning works as soon as this ships.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -324,6 +324,10 @@ exclude = [".git"]
 
 [tool.hatch.build.targets.wheel.force-include]
 "src/scitex_agent_container/cron/post-merge-pull.sh" = "scitex_agent_container/cron/post-merge-pull.sh"
+# The deploy-freshness alarm (cron half of `sac freshness`). Must ship in
+# the wheel: it is what writes the cache the every-invocation CLI warning
+# reads, so a host that installs sac from PyPI gets the alarm with it.
+"src/scitex_agent_container/cron/deploy-freshness-alarm.py" = "scitex_agent_container/cron/deploy-freshness-alarm.py"
 # Bundle pyproject.toml + README.md inside the wheel so the
 # source-bundled SIF build (cli_pkg/_image_source_build.py) can
 # reconstruct a pip-installable source tree from the INSTALLED
--- a/src/scitex_agent_container/_freshness/__init__.py
+++ b/src/scitex_agent_container/_freshness/__init__.py
@@ -1,0 +1,100 @@
+"""Deploy freshness: is the code we are running the code that shipped?
+
+Born from the 2026-07-13/14 comms outage. Seven PRs fixed a `sac listen`
+wedge; every one merged; the host ran 0.21.14 the entire time, which
+predates all of them. Agents hand-drained the wedged pool and re-diagnosed
+the same already-fixed bug for a day, because three consecutive tags never
+reached PyPI and NOTHING anywhere said a word.
+
+    "We are not failing to FIX these bugs. We are failing to SHIP them."
+
+The drift is structural, not bad luck. A release is
+``tag -> test -> build -> publish``. The tag is a local action that always
+succeeds; everything downstream is asynchronous and can fail silently. No
+consumer ever compares what it is running against what actually shipped,
+so drift is monotonic — it can only accumulate, never self-correct. (Six
+of the eighteen v0.21.x tags never published. The operator knew about two.)
+
+This package closes the loop: the consumer checks itself, and speaks up.
+
+* :mod:`._checks`  — the four failure modes, which are genuinely different
+  and do not catch each other: ghost tag / host behind PyPI / running !=
+  installed / release run failed.
+* :mod:`._symbols` — probes a fix by the SYMBOL it introduced, never by a
+  version string (which lies in both directions here).
+* :mod:`._warn`    — the every-invocation stderr warning. The operator's
+  primary ask: typing ``sac`` is how you find out.
+* :mod:`._cache`   — cron writes, CLI reads. The CLI never does I/O beyond
+  one small JSON file.
+
+Three states, and the middle one is the reason this works:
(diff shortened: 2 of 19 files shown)


ywatanabe1989:

PR #679, opened 2026-07-14
fix(listen-watchdog): the 5s fuse was RESTARTING a healthy control plane and deafening the whole fleet

ywatanabe1989:
The `sac listen` health watchdog was **restarting a healthy control plane**, and every restart **deafens every agent's inbox at once**. 「このインフラがぶれると死活問題」

## The bug: a probe that manufactured the outage it claimed to detect

`sac-listen-health-probe.sh` restarted `sac-listen.service` after **ONE** failed `curl` with a **5-second** deadline. On this box — which **idles at load 60-70** — five seconds is not a health check, it is a coin flip.

Measured, against a **REAL** server that answers `HTTP 200` in 8s (healthy, merely *busy*):

```
probe #1  ERROR: sac-listen DOWN — ... did not answer (transport failure).  -> RESTART
probe #2  ERROR: sac-listen DOWN — ... did not answer (transport failure).  -> RESTART
probe #3  ERROR: sac-listen DOWN — ... did not answer (transport failure).  -> RESTART

TOTAL RESTARTS of a HEALTHY (slow) daemon: 3 / 3 probes
```

**The daemon answered every single time.** The probe called it a "transport failure" every single time, because it could not tell **SLOW** from **DOWN**.

And the remedy is catastrophic: **every `sac listen` restart tears down the in-memory a2a `Broker`, which deafens EVERY agent's inbox at once.** So a slow probe did not merely mis-report — **it manufactured the outage it claimed to detect**, then re-probed *during its own restart*, saw a genuinely-down daemon, and restarted **again**. Live journal, 26 seconds apart:

```
10:57:26  health-probe: ERROR ... incident-class=sac-listen-watchdog
10:57:26  systemd: Stopping sac listen            <- restart #1 (of a HEALTHY daemon)
10:57:45  Started sac listen
10:57:52  health-probe: ERROR: sac-listen DOWN    <- probed DURING its own restart
10:57:52  watchdog is RESTARTING                  <- restart #2
```

**A probe that mutates is not a probe.** This single 5-second fuse plausibly explains the listen churn (3 pids in 7 min), the "registered but deaf" agents, the duplicate standby loops, and a peer's report that the auth pool "re-wedged within a minute of a drain" — that was not a wedge, it was a restart severing in-flight requests.

*(The listen running on the box right now was born at `10:57:52` — it **is** the daemon created by that second, false restart.)*

---

## The fix

Both directions are bugs. **Never acting** hides an outage; **over-acting CREATES one.** The false-RED is the worse of the two, because its remedy destroys a healthy thing — so only a **corroborated** verdict may restart.

### 1. Three states, never a bool

Two states cannot express *"I asked and got nothing"* — which is exactly what a loaded box produces.

| Verdict | Meaning | Weight |
|---|---|---|
| **UP** | It **answered** — any HTTP status `< 500` | — |
| **DOWN** | Connection **REFUSED** — kernel sent RST: *nothing is listening* | **2** (hard) |
| **DOWN** | **HTTP 5xx** — answered, but its health route errors | 1 (soft) |
| **UNKNOWN** | We asked and got **nothing** (timeout/DNS/reset) | 1 |

**Absence of evidence is not evidence of death.** `curl`'s `%{time_connect}` is used to say *why*: if the TCP handshake completed, the port **is** bound and the daemon has not exited, so the evidence line reads *"no HTTP answer within 20s (but TCP connected — the port IS bound)"*. That is the measurable difference between **busy** and **gone**.

### 2. Any HTTP status < 500 stays "UP" — preserved verbatim

A `401`/`403` **proves** the daemon is up: bound, speaking HTTP, auth-gating. Card `sac-listen-restart-healthcheck-bearer` (PR #463) exists because gating liveness on `status == 200` re-classified a live, 401-answering daemon as "down" — a false-RED that killed a **healthy** process. That lesson is load-bearing and is kept. Matches `_listen/_holder_health.py`.

### 3. Corroboration — and a failure is a FACT

Failure weight must reach `SAC_LISTEN_FAIL_THRESHOLD` (3):

* **2 consecutive REFUSALS** (2+2) → act fast. **Crash coverage is not weakened** — a genuinely dead listen still comes back (incident 2026-06-26, which is why this watchdog exists).
* **3 consecutive UNKNOWNs** (1+1+1) → act. **Wedge coverage kept** — a daemon that cannot answer a trivial route within 20s, three times running, is not "busy". Corroboration is what **promotes** a repeated UNKNOWN into a DOWN verdict.

**A single success does not wipe the ledger.** That bug — `consecutive = 0` reset by any one lucky reply — is this class's *other half*, and @-the-peer just fixed it in `sac listen`'s own holder check (**PR #673**, `_listen/_standby_ledger.py`), where a flapping holder oscillated `1/2` → "healthy" → `1/2` forever and was **never acted on**. This is consistent with that: a success builds a **serving streak**, and the ledger clears only after 2 consecutive UPs — logged **LOUD**, because "the thing I said was broken now looks fine" is exactly the transition an operator must never have hidden.

Blip once and you are not destroyed; keep failing and you are always eventually healed.

### 4. Never restart something that is already restarting

Two **independent** guards, so losing one cannot resurrect the 26s double-restart:

* **(a) Post-restart backoff** — after issuing a restart, the probe does not probe **at all** for 90s. Restart #2 above happened *because* it probed during its own restart.
* **(b) Unit-state guard** — if systemd reports the unit `activating`, it is already coming back (someone else restarted it, or it crashed and `Restart=always` caught it) → stand down. **Holds even if the ledger is lost or corrupt.**

### 5. Rate-limit the remedy

At most **2 restarts per 10 minutes**. Beyond that: **ALARM LOUDLY and STOP restarting** (`incident-class=sac-listen-watchdog-giving-up`, exit 1). If 2 restarts did not fix it, the 3rd will not either — and an unbounded restarter on a bad signal is how you take a fleet down at 3am. A human is needed, and it says so instead of thrashing.

### The ledger

The script is invoked **fresh** every ~30s, so an in-memory counter would always read zero and "N consecutive failures" could never be observed. State persists at `~/.scitex/agent-container/runtime/listen-health.state`. It is **never `source`d** — a corrupt or hostile state file must not become code — and any non-integer value resets that field to `0` (**fail-safe: no history == do not restart**). There is a test that writes `restarts=$(touch /tmp/pwned)` into the ledger and asserts the canary never appears.

**Deadline 5s → 20s.** The live daemon answers `/v1/health` in **~30 ms**, so 20s is ~600× the median and still catches a wedge. The alarm rail (`sac fleet notify blocker`) is unchanged and **not forked**; a second alarm case (giving-up) rides the same rail, capped at one page per window.

New: `--status` (what is the watchdog thinking?) and `--reset`.

---

## Tests — no mocks, and made to fail first

**41 new tests.** Every case drives the **real** script against a **real** HTTP server on a **real** ephemeral loopback socket — slow, refusing, 5xx-ing, 401-ing, dying. Nothing ever touches **port 7878**: that is the live control plane the whole fleet depends on, and a test that restarted it would repeat the incident.

Run against the **unfixed** script, the same 41 tests: **26 FAIL.** Against this one: **41 pass.**

The headline test is deliberately driven at the probe's **shipped default deadline** — no override. That is the whole point: the old default (5s) restarted an 8s-answering server 3 times out of 3, and a test that overrode the deadline would have passed on the broken script too and proved nothing.

Both directions are pinned: it must **not** restart on a slow-but-alive daemon / a single failure / an uncorroborated verdict / during its own restart / past the cap — and it **must** still restart a genuinely dead (`test_second_refusal_restarts_dead_listen`) or wedged (`test_three_timeouts_restart_wedged_listen`) one.

---

## Deployed + demonstrated live

A merged PR is not a fix. This is deployed to the host and the operator's stop-gap drop-in (`30-loaded-box-fuse.conf`, `SAC_LISTEN_PROBE_TIMEOUT=30`) is **removed** — its value is folded in, and superseded: a fixed deadline on a loaded box is still a race, whereas corroboration + backoff + rate-limit are not.

Deployment order was load-bearing: the new script went in **first**, because removing the drop-in while the old 5s-fuse script was still deployed would have brought the fuse straight back live.

```
deployed probe sha == repo sha   (byte-identical)
DropInPaths=                     (stop-gap gone)
Environment=                     (no override; the script's own 20s default applies)
sac-listen: UNTOUCHED by the deploy (same PID, same start timestamp)
```

### Live evidence

**1. The real listen, made to miss its deadline — OLD vs NEW, side by side.**

I did **not** hammer the shared control plane until it degraded: endangering the whole fleet to prove a point is the same mistake as the bug. "The daemon takes longer than the deadline" is a *relation*, and it can be produced from either side — so both watchdogs were pointed at the **REAL live listen** with a deadline it genuinely cannot meet (5 ms, vs its real 14 ms). Identical physics, zero risk. `SAC_LISTEN_UNIT` pointed at a nonexistent unit, so no real restart was ever issued; the **decision** is what was wrong, and the decision is what is measured.

```
--- OLD watchdog (develop) ---
  probe #1..#6: *** RESTART *** x6
  => 6 restarts of the HEALTHY live listen in 6 probes

--- NEW watchdog (this PR, as deployed) ---
  probe #1: no restart | UNKNOWN — no HTTP answer within 0.005s (but TCP connected in 0.00097s — the port IS bound)
  probe #2: no restart | answered (HTTP 200), but a failure ledger of 1/3 still stands
  probe #3: no restart | UNKNOWN — no HTTP answer within 0.005s (but TCP connected — the port IS bound)
  probe #4: RESTART                     <- corroborated: 3 misses out of 4
  probe #5: no restart | within post-restart backoff (0s < 90s) — NOT probing
  probe #6: no restart | within post-restart backoff (0s < 90s) — NOT probing
  => 1 restart in 6 probes
```

Note the new one **does** restart at probe #4 — and that is correct, not a leak. A daemon that misses its deadline 3 times out of 4 is exactly what the ledger is for, and note probe #2's answer did **not** erase the record (the PR #673 rule). What is gone is the 6× thrash and, decisively, **the double-restart**: the backoff refuses to even *probe* while the daemon is coming back. The safety in production comes from the deadline being ~600× the median, not from the ledger declining to act.

**2. The real listen under the operator's own load, with the real systemd timer.** 12 concurrent authenticated requests, 45s:

```
p50 0.029s   p90 0.185s   p99 0.473s   MAX 0.524s
samples over  5s (the OLD fuse -> instant restart): 0 / 1068
samples over 20s (the NEW deadline):                0 / 1068
watchdog restart decisions during the load: 0 / 5 probes
listen PID unchanged: 2350357.  Fleet stayed audible.
```

(Honest caveat: at 12 concurrent the live listen never actually got slow — max 0.52s — so this run alone does **not** exercise the slow path. That is what demo 1 is for.)

**3. Guard (b), against REAL systemd.** The `activating` stand-down is the one thing a container cannot unit-test, so it was exercised on the host with a real throwaway unit deliberately parked in `activating`:

```
guard-test unit ActiveState = activating
  probe #1: UNKNOWN — could not connect within 5s
  probe #2: UNKNOWN — could not connect within 5s
  probe #3: sac-wdog-guardtest.service is already 'activating' — standing down.
            It is coming back on its own, and restarting a restart is what
            produced the 2026-07-14 double-restart.
```

Corroborated at probe #3 — and instead of restarting, it stood down. The real `sac-listen` was never pointed at.

**4. The real deployed watchdog, now.** Timer firing every 30s, every probe quiet:

```
journalctl --user -u sac-listen-health: ... Finished. Finished. Finished.  (no ERROR lines)
journalctl --user -u sac-listen:        (no Stopping / Started since the deploy)

$ sac-listen-health-probe.sh --status
unit:             sac-listen.service (ActiveState=active)
probe_timeout:    20s (connect 5s)
failure_weight:   0 / 3
serving_streak:   0 / 2
restarts_in_600s: 0 / 2
last_restart:     never

sac-listen NRestarts=0, PID 2350357 (born 10:57:52 — the false restart's child, still serving)
```

### DIFF
--- a/scripts/systemd/README.md
+++ b/scripts/systemd/README.md
@@ -134,28 +134,186 @@ companion watchdog closes the gap:
   ~30s (`OnBootSec=30s`, `OnUnitActiveSec=30s`).
 * `sac-listen-health.service` (oneshot) runs
   `sac-listen-health-probe.sh`, which HTTP-probes
-  `http://127.0.0.1:7878/v1/health`. "Up" == *any* HTTP response
-  (incl. a 401/403 under bearer auth); only a transport failure
-  (connection refused / timeout) counts as "down" — the same
-  auth-change-proof liveness contract as
-  `_listen/_restart.py::wait_for_health`.
-* On a down probe the script:
-  1. logs a **LOUD ERROR** to the journal (`journalctl --user -u
-     sac-listen-health`) so the restart is VISIBLE, not silent;
-  2. best-effort emits the anomaly on the operator alarm path
-     (`sac fleet notify blocker`) — this may fail (the listen is the
-     transport for that notify), but is tried so a peer/lead listen or
-     a recovered inbox still surfaces it;
-  3. runs `systemctl --user reset-failed && restart sac-listen.service`
-     (the `reset-failed` clears a tripped `StartLimitBurst` so a
-     transient burst self-heals while a genuine hard-down stays
-     visible).
+  `http://127.0.0.1:7878/v1/health`.
 * The watchdog is itself decoupled — it does NOT `Requires=`/`After=`
   the listen (it must run and alarm precisely when the listen is
   down), and the timer is enabled independently.
-* Tunables via env on the probe: `SAC_LISTEN_HEALTH_URL`,
-  `SAC_LISTEN_UNIT`, `SAC_LISTEN_PROBE_TIMEOUT`, `SAC_LISTEN_NOTIFY`.
-  `--check-only` probes without any side effects (used by tests).
+
+## The health watchdog: how it decides
+
+**Incident 2026-07-14 — the watchdog WAS the outage.** The probe used to
+restart `sac-listen.service` after **ONE** failed `curl` with a **5-second**
+deadline. On a box that idles at **load 60-70** that is not a health check,
+it is a coin flip. Measured against a REAL server that answered `HTTP 200`
+in 8s (healthy, merely *busy*):
+
+```
+probe #1 -> "sac-listen DOWN ... (transport failure)" -> RESTART
+probe #2 -> "sac-listen DOWN ... (transport failure)" -> RESTART
+probe #3 -> "sac-listen DOWN ... (transport failure)" -> RESTART
+TOTAL RESTARTS of a HEALTHY daemon: 3 / 3
+```
+
+The daemon **answered every time**. The probe called it a "transport
+failure" every time, because it could not tell SLOW from DOWN.
+
+And the remedy is catastrophic: **every `sac listen` restart tears down the
+in-memory a2a `Broker`, which deafens EVERY agent's inbox at once.** So a
+slow probe did not merely mis-report — **it manufactured the outage it
+claimed to detect**, then re-probed *during its own restart*, saw a
+genuinely-down daemon, and restarted again. Live journal, 26s apart:
+
+```
+10:57:26  health-probe: ERROR ... incident-class=sac-listen-watchdog
+10:57:26  systemd: Stopping sac listen            <- restart #1 (of a HEALTHY daemon)
+10:57:45  Started sac listen
+10:57:52  health-probe: ERROR: sac-listen DOWN    <- probed DURING its own restart
+10:57:52  watchdog is RESTARTING                  <- restart #2, 26s later
+```
+
+**A probe that mutates is not a probe.** That single 5-second fuse plausibly
+explains the listen churn (3 pids in 7 min), the "registered but deaf"
+agents, the duplicate standby loops, and the auth pool appearing to
+"re-wedge within a minute of a drain" (it was not a wedge — it was a restart
+severing in-flight requests).
+
+Both directions are bugs: never acting hides an outage, over-acting
+**creates** one. The false-RED is the worse of the two, because its remedy
+destroys a healthy thing. So **only a corroborated verdict may restart.**
+
+### 1. Three states, never a bool
+
+Two states cannot express *"I asked and got nothing"* — which is precisely
+what a loaded box produces.
+
+| Verdict | Meaning | Weight |
+|---|---|---|
+| **UP** | It **answered** — any HTTP status `< 500` (see §2) | — |
+| **DOWN** | Connection **REFUSED** — the kernel sent RST: *nothing is listening* | **2** (hard) |
+| **DOWN** | **HTTP 5xx** — it answered, but its health route is erroring. Bound and speaking HTTP, yet not healthy | 1 (soft) |
+| **UNKNOWN** | We asked and got **nothing** (timeout / DNS / reset). Under load this is what a HEALTHY-but-busy daemon looks like | 1 |
+
+**Absence of evidence is not evidence of death.** A timeout is UNKNOWN, and
+`curl`'s `%{time_connect}` is used to say *why*: if the TCP handshake
+completed, the port **is** bound and the daemon has not exited — so the
+evidence line reads "no HTTP answer within 20s (but TCP connected — the port
+IS bound)". That is the difference between *busy* and *gone*.
+
+### 2. Any HTTP status < 500 is "UP" (deliberate — keep it)
+
+A `401`/`403` **proves** the daemon is up: bound, speaking HTTP, and
+auth-gating. Card `sac-listen-restart-healthcheck-bearer` (PR #463) exists
+because gating liveness on `status == 200` re-classified a live,
+401-answering daemon as "down" — a false-RED that killed a **healthy**
+process. Only a *server error* (5xx) or *no answer at all* counts against
+the daemon. Matches `_listen/_holder_health.py`.
+
+### 3. Corroboration — and a failure is a FACT
+
+Accumulated failure weight must reach `SAC_LISTEN_FAIL_THRESHOLD` (3):
+
+* **2 consecutive REFUSALS** (2+2) → act fast. **Crash coverage** — a
+  genuinely dead listen still comes back (incident 2026-06-26).
+* **3 consecutive UNKNOWNs** (1+1+1) → act. **Wedge coverage** — a daemon
+  that cannot answer a trivial route within a 20s deadline, three times
+  running, is not "busy". Corroboration is what **promotes** a repeated
+  UNKNOWN into a DOWN verdict.
+
+**A single success does not wipe the ledger.** That bug — `consecutive = 0`
+reset by any one lucky reply — is this class's *other half*, and a peer just
+fixed it in `sac listen`'s own holder check (PR #673,
+`_listen/_standby_ledger.py`), where a flapping holder oscillated
+`1/2` → "healthy" → `1/2` forever and was **never acted on**. Here a success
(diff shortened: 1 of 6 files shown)


ywatanabe1989:
## CI caught 2 failures — and the second one led somewhere worse

**2 failed, 10086 passed.** Both failures were in the *pre-existing* `tests/integration/test_sac_listen_health_probe.py`, and neither was a regression: **both assertions were pinning the OLD, BUGGY contract.**

```
assert 'UNHEALTHY' in stderr
  actual: 'sac-listen-health: DOWN http://... — connection refused
           (nothing is listening on the port)'
```
`UNHEALTHY` was a **two-state** word, and two states cannot distinguish *"connection refused"* from *"I asked and got nothing"*. Collapsing those is precisely what restarted a healthy control plane. The vocabulary changed with the decision model.

```
assert 'ERROR: sac-listen DOWN' in stderr          # after ONE probe
  actual: 'WARN: ... Failure weight 2/3 — NOT restarting: an uncorroborated
           verdict is not grounds to destroy a control plane the whole fleet
           depends on.'
```
**This test literally asserted the false-RED** — that one failed probe emits the loud restart line. It now earns the verdict first (a refusal is weight 2, threshold 3, so a second one crosses it).

---

## The thing that actually mattered: the test suite could have restarted the real listen

Chasing that second failure surfaced a hole **I introduced**, and it is worse than the two test failures.

`_run_probe()` never set `SAC_LISTEN_HEALTH_STATE`. My new probe persists its failure ledger at `$HOME/.scitex/agent-container/runtime/listen-health.state` — so **every heal-mode test was writing the REAL ledger.**

Two consequences:

**1. Order-dependence — a green test proving nothing.** `test_heal_mode_logs_restart_intent` (`assert "RESTARTING"`) *passed* in CI. It should not have: one refusal is weight 2, below the threshold of 3. It passed only because the test that ran before it left `failures=2` behind, which silently satisfied its corroboration. It was asserting a restart it had not earned.

**2. The test suite could cause the outage it tests for.** On the host, a test leaving `failures=2` in the live ledger means the **next real 30s timer tick** reaches the threshold and **restarts the real `sac listen`** — tearing down the in-memory a2a Broker and deafening every agent's inbox. The suite would trigger the exact incident the watchdog exists to prevent.

**CI cannot see this.** There is no fleet on a runner, so the real ledger does not exist there and the tests pass either way. It would only ever have fired on the operator's box — which is the one place it must not.

Fixed in two layers:
- a **force-set floor in `tests/conftest.py`** (holds even for code paths that bypass fixtures, and protects future tests, not just these);
- a **per-test ledger** in the probe suite (kills the order-dependence).

Real env + real file + teardown restore — no `monkeypatch` (STX-NM002; the linter was right to block my first attempt).

Also added `test_uncorroborated_down_does_not_restart`, so the regression that made this whole rewrite necessary is now pinned in the very file that used to demand it.

---

## Verified

```
71 passed in 119.25s        (30 in the old probe suite + 41 new)

=== REAL host ledger AFTER a full suite run ===
failures=0 serving_streak=0 last_restart=0 last_alarm=0 restarts=
LEDGER SAFE: the test suite did NOT write the live watchdog's ledger.
```

And the live control plane is still the same process throughout: PID 2350357, `NRestarts=0`, zero `RESTARTING` lines.

PR #690, opened 2026-07-14
feat(host): sac host sync — one-way code sync, centre → remote, and drift stops being silent

ywatanabe1989:
## The gap

The WSL box is the brain. All config and authority live at the centre, and code flows one way: centre → remote. A remote never originates it.

**Today the centre can LAUNCH an agent on Spartan but cannot control WHICH CODE runs there — and nothing announces the difference.**

Spartan's sac checkout was **five releases stale** (v0.21.14 while develop was v0.21.20). There is no post-merge pull anywhere in the fleet. It was found *by hand*, an hour before this PR. Fixing it by hand is a one-off patch that rots on the next merge, and Constitution §4 is explicit: a change is only live when *every* layer that caches a version is reconciled — PyPI, GitHub, **both hosts**, the base image, agent overlays, CI, editable installs.

Two more silent divergences surfaced the same day, and both are the *same shape* — "the remote diverged from master and nobody was told":

1. **The Spartan sac checkout doubles as the self-hosted CI runner's audit workspace.** An agent left a branch checked out there; a `develop` CI run then audited *that branch* while reporting itself as a test of develop. It broke the release for ~40 minutes. Nothing announced it.
2. **Spartan's `~/.scitex` was a symlink into an unrelated paper project.** The registry was right; reality had drifted. Silently, for weeks.

## What this adds

`sac host sync` — in two halves, and **detection is the product**:

```bash
# READ-ONLY. Mutates nothing. Exits non-zero on drift, so it is an ALARM,
# not a report nobody reads. Cron-ready; usable as a pre-flight before any
# remote dispatch.
sac host sync --check spartan
sac host sync --check --all --json

# The remedy: fast-forward only, behind loud preconditions.
sac host sync spartan
sac host sync --all
```

The detector is the half that would have caught **both** incidents above, and it is shipped first deliberately.

## The invariants

**AHEAD is an ALARM, not a merge.** This is the load-bearing decision. If a remote holds commits the centre lacks, the one-way property has *already* been violated — something wrote code on a machine that is not allowed to originate it. So sac:

- does **not** auto-merge them back — that would ratify the violation and make the remote a source of truth;
- does **not** fast-forward or reset over them — that would destroy them, unexamined;
- **prints them by subject line and REFUSES.**

A diverged remote is a **bug report, not a branch to reconcile**. The error message states exactly that, and names the two commands a human may choose between (push them, or discard them explicitly).

The rest:

- **Dirty trees are refused**, never stashed, with the offending files listed.
- **UNKNOWN is not clean.** An unreachable peer, a failed fetch, or an unreadable CI state all refuse. sac has repeatedly shipped the bug where a failed probe collapses into a reassuring pole (`pid<=0` → "dead" destroyed live agents; a probe timeout → "DOWN" restarted a healthy daemon). Absence of evidence is not evidence.
- **CI guard.** Refuses while the peer's runners are busy *or* a run is merely **queued** — an idle runner is one queued job away from busy, which is less time than a merge takes to land. Checking only the `busy` flag would be a race with a corrupt CI run as the prize.
- **Object graph, not mtimes.** `rev-list --count` both ways, plus `status --porcelain`. mtimes are wrong here: a plain `git pull` rewrites them without changing content, and GPFS clock skew across hosts makes them unreliable. The graph is content-addressed, exact, clock-independent.
- **Never silent.** There is no quiet success path. A no-op still prints what it verified; a refusal prints the actionable reason; an override prints what it overrode.

## Verified by symbol, never by a version string

Version strings are proven liars in this repo — eleven tags shipped **nothing** while reporting success, and the centre's own `.dist-info` still reports 0.21.11 next to current code. So after the fast-forward, sac **re-probes the peer** and asserts:

1. HEAD is **exactly** the sha it aimed at;
2. the interpreter that runs sac **loads its code from inside that very checkout** — this is what catches a wheel or a fossil `.dist-info` shadowing an editable install, the failure mode that hides stale code for months;
3. a **real symbol imports** out of it.

If any disagree, the sync reports **FAILED**. It does not report a success it cannot substantiate.

The same discipline resolves the remote checkout path in the first place: sac asks the peer's own interpreter where it loads sac from, rather than guessing `~/proj/scitex-agent-container`. Expanding `~` for a remote path expands it on the *calling* process — the centre's home — which is the exact footgun `_state/host_registry.py` exists to keep out of sac. (On Spartan that path is also a symlink into GPFS, so even the right-looking guess is a lie.)

## Reuse, not reinvention

Everything dispatches through **`build_ssh_argv`**, sac's single remote choke point, so ProxyJump chains and the Lmod `env_preamble` apply for free and a call site added later cannot silently miss them. This repo has been bitten by a primitive fixed at one of six call sites; there is deliberately no second ssh path here.

## Live verification (measured, not inferred)

Run against the real fleet from this branch:

```
current   spartan  current with origin/develop
    checkout   /data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container
    module     /data/gpfs/.../src/scitex_agent_container/__init__.py
    ref        origin/develop @ 64418a4c4c8f
    head       64418a4c4c8f
    symbol     claim_port['agent_name', 'range_', 'explicit', 'db_path', 'explicit_is_pin']
```

The CI guard, live against the real GitHub API: `spartan → idle (3 runners idle, no queued/in-progress runs)`; `mba → not-applicable (no runner registered there)`.

**And on its very first `--check --all`, the detector found two real, previously-unannounced facts:**

- **mba** — `git fetch` fails there, so its code state is genuinely **UNKNOWN**. Correctly *not* reported as clean.
- **nas** — sac is installed as a **plain wheel**, not a checkout. There is no git tree to reconcile; it needs pip, not a pull. `sync` would never have worked there, and now it explains why instead of failing obscurely.

Neither was known before this PR existed. That is the detector doing its job on day one.

## Why this PR stops here

**Scope is code only.** No credential distribution — that proposal is explicitly marked "do not deploy" and is a separate decision. The verb is *designed* so credentials can ride the same one-way, precondition-guarded channel later (a sibling apply step behind the same probe/refuse/verify), rather than growing a second, unguarded path to the same hosts.

**Nothing is made read-only in this PR**, and the operator's own reasoning is why: guessing the writable boundary would stall the fleet. So we **measure first** — run `--check` for a week and let the data define where the boundary actually belongs. That is the cheap, reversible step, and it produces the evidence the next stage needs.

## One deliberate deviation, flagged for review

The brief permitted `--force` to override the AHEAD refusal *provided it printed the commits it would discard*. **I did not implement that.** Discarding ahead-commits requires `reset --hard`, which is the one operation the hard constraint forbids ("FF-ONLY … never merge/rebase/reset --hard"). Rather than ship a flag whose only function is destruction, `--force` overrides the **CI-idle guard only** — a *scheduling* guard, not a data-safety one — and the ahead-commits are printed **at the refusal**, alongside the exact command a human may run themselves to discard them.

Net effect: **there is no code path in sac that destroys a remote commit or an uncommitted remote edit.** That is precisely what makes this verb safe to run unattended from a cron, which is the whole point of the detector. Happy to add the destructive escape hatch if you disagree — but it should be a conscious choice, not a default.

## Tests

Real seams only — no mocks. Injected real callables and PATH-installed shims, per this repo's PA-306 rule.

**1301 green across every touched path** (`_hostsync` 91 · `_host_sync` CLI 9 · `host_group`/`host_crud`/`_drift` 102 · `_state` 1099). Local runs scoped to touched paths; CI runs the rest.

Covering, specifically: dirty-tree refusal · AHEAD refusal (and that the fast-forward is **never even attempted**) · diverged refusal · busy-runner refusal · **queued**-run refusal · `gh`-unavailable → UNKNOWN → refusal · truncated probe output → UNREACHABLE, not "clean" · **symbol-probe mismatch → FAILED, not success** · module-loaded-outside-the-synced-tree → FAILED · `--check` never runs a merge and never calls the CI API · `--check` exits non-zero on drift · no `~` is ever baked into a remote command.


ywatanabe1989:
### The CI guard just caught the race it was built for — live, on this PR

While **this PR's own `pytest-matrix` was executing on Spartan's runners**, I ran the guard against the real GitHub API:

```
state      : busy
may_mutate : False
busy       : ()            <-- every runner's `busy` flag was FALSE
active runs: 1             <-- but a workflow run WAS in progress
detail     : 1 workflow run(s) are queued or in progress ... 'spartan's runners are
             idle only for the moment — a queued job can start mid-sync.
```

Read that second line again. **The `busy` flag was false on every runner while CI was running on them.**

The original spec for this guard was "all `busy=false`". Implemented literally, it would have returned **IDLE** at that moment and **authorised a fast-forward of the very checkout the running job was auditing** — reproducing the 2026-07-14 corruption exactly (a `develop` run auditing the wrong tree, silently, for 40 minutes).

The second signal — queued / in-progress **runs**, not just the runner flag — is what refused. This is why the guard checks both, and it is not a hypothetical: *an idle runner is one queued job away from busy, which is less time than a merge takes to land.*

Measured, not argued. I could not have staged this test; the PR's own CI produced it.

ywatanabe1989:
### CI red here is NOT this PR — develop's audit gate is broken for everyone

`pytest-matrix` fails on all three legs with **`1 failed, 10418 passed`**. The one failure is:

```
tests/develop/test_audit.py::test_audit_all_clean
  tests/scitex_agent_container/_state/test__port_allocator_concurrency.py: no matching src file
```

That file arrived with **#685** and is on develop. It is not touched by this PR, and it fails identically on every open PR and on develop itself. Verified against the blobs (`git cat-file -e origin/develop:<path>` → present; introduced by `be00b755`).

Fixed separately in **#691** (a pure rename — one underscore; the file used the *private*-module convention `test__foo.py` ↔ `_foo.py` while its subject `_state/port_allocator.py` is *public*). Once #691 lands, this branch picks it up and re-runs.

For the record, this PR's own test files were checked against scitex-dev's actual `_test_has_src_match` predicate — all six mirror correctly:

```
OK  _hostsync/test__apply.py       OK  _hostsync/test__probe.py
OK  _hostsync/test__ci_guard.py    OK  _hostsync/test__sync.py
OK  _hostsync/test__model.py       OK  cli_pkg/test__host_sync.py
```

There is a nice symmetry in this: #685 is the PR that made the PR gate run the *same* audit as the release gate, ending the "green PR, ghost release" era. The first thing the newly-honest gate caught was #685's own test file — and the first thing it did to *this* PR was prove that a stale, unreconciled tree silently breaks a gate. Which is, more or less, the thesis.

ywatanabe1989:
## This rename is CORRECT but NOT SUFFICIENT — the audit gate is reading the wrong tree

CI on this PR still fails, with the orphan reported **at the old filename**. That is the decisive datum: the file is renamed *in this PR's tree*, and the auditor still sees the old name. **The audit is not reading this PR's code.**

### What it IS reading

Look at the path in the error:

```
ERRO: scitex-agent-container (/data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container): 1 error(s)
```

The PR's code is checked out at `/tmp/scitex-ci-runner-work/org-spartan-cpu-01/...`. `/data/gpfs/.../scitex-agent-container` is **Spartan's shared, persistent sac checkout**.

### The mechanism (scitex_dev `_resolve_repo_root`)

1. `find_spec("scitex_agent_container")` → in the SIF, `run-in-sif.sh` installs via `uv pip install --target="$TMPDIR/site" -e .`, so the package resolves under `$TMPDIR/site`, which has **no `pyproject.toml` above it**.
2. The src-layout and flat-layout walks both fail.
3. It then hits the fallback (`_discovery.py` L48-69): *"module is in site-packages… try common development checkout locations"* — it **guesses `~/proj/<distribution>`**.
4. On the Spartan runner `~/proj/scitex-agent-container` is a **symlink** → `/data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container`. Measured: `readlink -f` resolves exactly to the path in the error.

So the gate audits a shared mutable tree that no PR controls.

### Why this matters far beyond one filename

The audit's verdict on **any** PR reflects Spartan's checkout, not the PR. It is a false signal in **both** directions: it can fail a clean PR (this one), and it would pass a PR that introduces a violation. A gate that audits the wrong tree is a gate that is off.

This is the same failure that broke develop this morning — a CI run auditing a tree that wasn't the one it claimed to test — and the same `~`-guessing footgun that `_state/host_registry.py` exists to keep out of sac. It reproduces the symlink trap exactly.

### The deadlock

The orphan is on develop → Spartan tracks develop → the audit reads Spartan → red on every PR. This PR removes the orphan from develop, but **cannot merge**, because the gate it would fix is the gate blocking it.

### Breaking it

1. **Merge this PR with an admin override.** It is a pure rename, verified against scitex-dev's own `_test_has_src_match` predicate (`False` → `True`), with zero test-content change. This needs a human with the authority to bypass a required check — I have deliberately not done it.
2. Then reconcile the audited tree: `sac host sync spartan` (#690) — which is precisely the verb that makes a stale remote checkout impossible to have unnoticed.

### The durable fix (belongs to the audit gate / scitex-dev, not here)

The auditor must audit **the code under test**, never a checkout it guessed from `~/proj`. Either:

- have the gate pass an explicit `--path <repo root>` (`audit-all` supports it; the `audit_all_for_package` helper does not expose it), or
- delete the `~/proj` guessing fallback — it is unsound on any machine that has a development checkout, which is every machine in this fleet.

A fallback that silently substitutes a *different tree* for the one you asked about is worse than no answer at all.

ywatanabe1989:
## Confirmed by direct evidence: the audit gate audits Spartan's checkout, not the PR

Earlier I posted this diagnosis from inference. Here is the proof.

**#691's only change is renaming that one file.** Its CI run reports:

```
ERRO: scitex-agent-container (/data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container): 1 error(s)
        /data/gpfs/.../tests/scitex_agent_container/_state/test__port_allocator_concurrency.py
```

The PR's tree contains `test_port_allocator_concurrency.py` (renamed). The audit reports **`test__port_allocator_concurrency.py`** — the *old* name — at the **`/data/gpfs`** path. The PR's code is checked out at `/tmp/scitex-ci-runner-work/...`.

A tree cannot simultaneously hold the new name and report the old one. **The auditor is reading a different tree**: Spartan's shared, persistent sac checkout, reached via `_resolve_repo_root`'s `~/proj/<name>` fallback — and on that runner `~/proj/scitex-agent-container` is a symlink into `/data/gpfs/...`.

### The gate was measuring a stale tree

`sac host sync --check spartan`, run just now against the live peer:

```
DRIFTED   spartan  2 behind origin/develop (STALE CODE)
    checkout   /data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container
    ref        origin/develop @ 7db73752
    head       64418a4c
exit=1
```

That checkout — two commits stale, still holding the orphan — **is the tree the audit gate reads.** So the gate's verdict on every PR reflects Spartan's checkout, not the PR. False in *both* directions: it fails clean PRs (these two), and it would pass a PR that introduces a violation.

### The guard then refused to let me fix it unsafely

Running `sac host sync spartan` while #690's CI was executing:

```
REFUSED   spartan  2 behind origin/develop (STALE CODE)
    ci  busy — 3 runner(s) on 'spartan' are executing a job right now.
        Syncing would move the checkout under a running audit — exactly the
        corruption that broke develop's CI on 2026-07-14.
```

The runners were mid-audit **on that very checkout**. Fast-forwarding it would have moved the tree under a live job — reproducing the incident this guard exists to prevent. It stopped me. The sync runs once the runners go idle.

### The durable fix belongs upstream

Reconciling Spartan clears the symptom. The root cause is in the auditor: it must audit **the code under test**, never a checkout it guessed from `~/proj`. Either pass an explicit `--path <repo root>` (`audit-all` supports it; the `audit_all_for_package` helper does not expose it), or drop the `~/proj` fallback — it is unsound on every machine in this fleet, because they all have development checkouts.

A fallback that silently substitutes a *different tree* for the one you asked about is worse than no answer at all. That is the same lesson as the `~`-expansion trap in `_state/host_registry.py`, and the same lesson as this PR.

ywatanabe1989:
## Proof, and the fix: the audit gate was reading Spartan's checkout, not the branch under test

The cleanest artifact in this whole investigation is **develop's own post-merge run**:

```
run 29326570609
title  Merge pull request #691 from scitex-ai/fix/ps204-orphan-concurrency-test
branch develop
      /data/gpfs/.../tests/scitex_agent_container/_state/test__port_allocator_concurrency.py
1 failed, 10318 passed
```

Read it slowly. **develop had just merged the rename.** develop's tree contains `test_port_allocator_concurrency.py`. The audit nonetheless reported `test__port_allocator_concurrency.py` — the *old* name — at `/data/gpfs/...`.

A branch cannot hold the new name and report the old one. The auditor was never reading the branch. It was reading **Spartan's shared, persistent checkout**, reached through `_resolve_repo_root`'s `~/proj/<name>` fallback — which on that runner is a symlink into `/data/gpfs/...`.

So for as long as that checkout was stale, **the audit gate's verdict on every PR — and on develop — was a statement about a tree nobody was testing.**

## The fix, by the verb this PR adds

`sac host sync spartan`, run against the live peer:

```
attempt 1..4
REFUSED   spartan  2 behind origin/develop (STALE CODE)
    ci  busy — 3 runner(s) on 'spartan' are executing a job right now.
        Syncing would move the checkout under a running audit — exactly the
        corruption that broke develop's CI on 2026-07-14.

attempt 5
    ci      idle — 3 runner(s) idle; no queued or in-progress runs
    git: Updating 64418a4c..7db73752
    git: Fast-forward
    git:  ...__port_allocator_concurrency.py => test_port_allocator_concurrency.py} | 0
    head    7db737523a1b        (== the ref it aimed at)
    symbol  claim_port['agent_name', 'range_', 'explicit', 'db_path', 'explicit_is_pin']
=== SYNC OK ===
```

Four refusals, then one fast-forward. The guard would not touch the checkout while the runners were mid-audit **on that very tree** — the corruption it exists to prevent — and moved it the moment they idled. The single file it changed is the orphan rename. Verification is by loaded-module path + symbol, never a version string.

The audited tree, confirmed on the peer afterwards:

```
head=7db73752  Merge pull request #691 ...
(clean)
test_port_allocator_concurrency.py
test_port_allocator.py
```

The orphan is gone from the tree the gate actually reads.

## The durable fix is still open, and it is not in sac

Reconciling Spartan cleared the symptom. The root cause remains: **the auditor must audit the code under test.** Either pass an explicit `--path <repo root>` (`audit-all` supports it; the `audit_all_for_package` helper does not expose it), or delete the `~/proj` guessing fallback in `scitex_dev/_cli/audit/_project/_discovery.py` (L48-69). It is unsound on every machine in this fleet, because they all have development checkouts — and where that checkout is a symlink, it silently substitutes a *different tree* for the one you asked about.

That is the same failure this PR was written to eliminate, arriving from the other direction. A gate that audits the wrong tree is a gate that is off — false when it fails a clean branch, and false when it passes a dirty one.

## Also worth a look

`pytest-matrix` is evidently **not a required status check** — #691 merged with all three legs red. That is how #685's orphan reached develop in the first place.

PR #691, opened 2026-07-14
fix(tests): un-orphan the port-allocator concurrency test — develop's audit gate is red for every PR

ywatanabe1989:
## develop is red, and it blocks every open PR

Since #685 landed, `tests/develop/test_audit.py::test_audit_all_clean` fails on **every** PR and on develop itself:

```
=== audit-project ===
ERRO: scitex-agent-container: 1 error(s)
        tests/scitex_agent_container/_state/test__port_allocator_concurrency.py:
        no matching src file (orphan test)
```

Result: `1 failed, 10418 passed`. All three `pytest-matrix` legs (3.11 / 3.12 / 3.13) go red on this one file, so nothing can merge.

## There is a pleasing irony here

#685 is the PR that made **the PR gate run the same audit as the release gate** — the change that ended the era of "green PR, ghost release". The very first thing the newly-honest gate caught was **#685's own test file**.

The gate is not broken. It is working, on its first day, exactly as intended. This PR just clears the one thing it found.

## Root cause: one underscore

The file was named with the **private**-module convention (`test__foo.py` ↔ `_foo.py`), but its subject — `_state/port_allocator.py` — is **public** (no leading underscore).

`_test_has_src_match` resolves a test to its source by prefix, then strips trailing `_<descriptor>` segments, so one source file can host several themed test modules:

| test file | resolves to | |
|---|---|---|
| `test__port_allocator_concurrency.py` | `_port_allocator.py` | **missing → orphan** |
| `test_port_allocator_concurrency.py` | `port_allocator.py` | **exists → ok** |

This is the same mechanism that lets `test_host_config_resolve.py` mirror `host_config.py`.

## Verified by executing the rule, not by reading it

```
test__port_allocator_concurrency.py  -> src match: False
test_port_allocator_concurrency.py   -> src match: True
```

That is scitex-dev's own `_test_has_src_match` predicate run against the real source tree.

## Scope

**Pure rename. Zero test-content changes.** Those tests guard the TOCTOU race in `claim_port` that ghosted the v0.21.19 release — they are valuable and they stay exactly as they are. Only the filename changes, so it mirrors the public module it actually tests.

Found while shipping #690 (`sac host sync`), whose CI this unblocks — along with everyone else's.


ywatanabe1989:
## This rename is CORRECT but NOT SUFFICIENT — the audit gate is reading the wrong tree

CI on this PR still fails, with the orphan reported **at the old filename**. That is the decisive datum: the file is renamed *in this PR's tree*, and the auditor still sees the old name. **The audit is not reading this PR's code.**

### What it IS reading

Look at the path in the error:

```
ERRO: scitex-agent-container (/data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container): 1 error(s)
```

The PR's code is checked out at `/tmp/scitex-ci-runner-work/org-spartan-cpu-01/...`. `/data/gpfs/.../scitex-agent-container` is **Spartan's shared, persistent sac checkout**.

### The mechanism (scitex_dev `_resolve_repo_root`)

1. `find_spec("scitex_agent_container")` → in the SIF, `run-in-sif.sh` installs via `uv pip install --target="$TMPDIR/site" -e .`, so the package resolves under `$TMPDIR/site`, which has **no `pyproject.toml` above it**.
2. The src-layout and flat-layout walks both fail.
3. It then hits the fallback (`_discovery.py` L48-69): *"module is in site-packages… try common development checkout locations"* — it **guesses `~/proj/<distribution>`**.
4. On the Spartan runner `~/proj/scitex-agent-container` is a **symlink** → `/data/gpfs/projects/punim0264/ywatanabe/scitex-agent-container`. Measured: `readlink -f` resolves exactly to the path in the error.

So the gate audits a shared mutable tree that no PR controls.

### Why this matters far beyond one filename

The audit's verdict on **any** PR reflects Spartan's checkout, not the PR. It is a false signal in **both** directions: it can fail a clean PR (this one), and it would pass a PR that introduces a violation. A gate that audits the wrong tree is a gate that is off.

This is the same failure that broke develop this morning — a CI run auditing a tree that wasn't the one it claimed to test — and the same `~`-guessing footgun that `_state/host_registry.py` exists to keep out of sac. It reproduces the symlink trap exactly.

### The deadlock

The orphan is on develop → Spartan tracks develop → the audit reads Spartan → red on every PR. This PR removes the orphan from develop, but **cannot merge**, because the gate it would fix is the gate blocking it.

### Breaking it

1. **Merge this PR with an admin override.** It is a pure rename, verified against scitex-dev's own `_test_has_src_match` predicate (`False` → `True`), with zero test-content change. This needs a human with the authority to bypass a required check — I have deliberately not done it.
2. Then reconcile the audited tree: `sac host sync spartan` (#690) — which is precisely the verb that makes a stale remote checkout impossible to have unnoticed.

### The durable fix (belongs to the audit gate / scitex-dev, not here)

The auditor must audit **the code under test**, never a checkout it guessed from `~/proj`. Either:

- have the gate pass an explicit `--path <repo root>` (`audit-all` supports it; the `audit_all_for_package` helper does not expose it), or
- delete the `~/proj` guessing fallback — it is unsound on any machine that has a development checkout, which is every machine in this fleet.

A fallback that silently substitutes a *different tree* for the one you asked about is worse than no answer at all.

PR #729, opened 2026-07-17
feat(twin): --fork-session + deterministic --tag ids + boot identity gate (and: twin was inert)

ywatanabe1989:
Refines the **existing** `sac agents twin` (ADR-0019) on the three axes the operator directed 2026-07-17 — and fixes two defects found while doing so that made the feature **inert in production**.

Refs card `sac-agents-fork-context-inheriting-spawn-20260717`; new card `sac-twin-derive-spec-env-v3-invalid-20260717` for the defects. ADR-0019 amended.

---

## ⚠️ Headline: `sac agents twin` could not have worked end-to-end

Both defects are in `derive_twin_spec`, **demonstrated** against the real on-disk `scitex-tex` spec (which loads fine as a baseline) — not inferred from reading:

1. **Twin specs were unloadable.** Identity env was written to the **top-level `spec.env`**, which v3 validation rejects (`_V3_RELOCATED_FIELDS: env -> spec.apptainer.env`). Real parents have no top-level `spec.env`, so `derive_twin_spec` *created* an invalid one every time. Chain: derived spec fails validation → twin never starts → never receives `SAC_TWIN_PARENT` → `seed_twin_from_parent` (whose sole trigger is that var) **never runs** → no context inheritance at all.

   ```
   *** TWIN SPEC REJECTED ***: spec.env is no longer accepted at the top level;
       move it to spec.apptainer.env (v3 spec realignment §3).
   ```

2. **The parent's identity was re-injected via inherited `raw_args`.** Real specs pin identity as `raw_args: [--env, SCITEX_TODO_AGENT_ID=<parent>]`, and `build_run_argv` appends `raw_args` **after** every curated `--env`. Deep-copied into a twin, that re-asserts the parent's identity — arriving through the one channel the new boot gate *cannot see* (the gate reads host-side `config.env`, which correctly reports "twin", while the container's env would report "parent").

**Why nobody noticed:** the 29 twin tests passed throughout. Their fixture used a top-level `spec.env` — a shape no real spec has and the validator rejects — and the suite **never ran the validator**. Tests written to match the implementation's assumption cannot disagree with it. `test_derived_twin_spec_passes_v3_validation` is the test that can.

Fixed **in this PR** rather than deferred: all three assigned refinements are provably dead code without #1. Shipping them alone would have been shipping a lie.

---

### 1. Use the supported `--fork-session`

First boot now launches `claude --resume <parent> --fork-session --session-id <derived>` (SDK: the same three via `ClaudeAgentOptions`) instead of seeding the twin's live session to the parent's uuid.

- **No live collision is claimed.** Nobody demonstrated one, and I did not. The justification is *using the supported mechanism instead of hand-rolled state surgery*.
- **The transcript copy stays** and is **not** what the fork replaces — a point the brief conflated: parent and twin have **separate container homes** and `--resume` resolves from the **local** `~/.claude/projects`. The copy is cross-home transport; the fork only handles the id.
- The `session_id` marker still seeds to the parent's uuid: it is the in-container SDK runner's only "resume FROM" channel. The fork advances it after turn one, so it is transient, never the twin's identity.
- New validated `spec.claude.fork_session` / `session_id` reach **both** runtimes — 88/97 fleet specs are `runtime: tui`, 9 are `claude-agent-sdk`, and a twin inherits its parent's runtime, so both had to move together. SDK threading rides the documented `build_sdk_options(extra=)` seam, **first attempt only** (attempt>0 is the supervisor's resume-recovery; re-forking there would fight it).

### 2. `--tag` → deterministic ids

`<parent>-forked-<tag>`, session uuid `uuid5(<twin-id>)`. Re-running a tag targets the **same** twin instead of minting `-twin-2`, `-3`. Lands on **all four surfaces together** per the 掟: Python API, CLI, `agent_twin` MCP tool, skill 33 (incl. *when* to use it). Legacy default retained for back-compat.

### 3. Boot identity gate

`assert_twin_identity` **refuses to boot** a twin whose `SCITEX_TODO_AGENT_ID` is missing, is its parent's, or is not its own name. Runs on **every** boot (the transcript claims the parent's identity for the twin's whole life), before the first-boot early-return.

**Scope line kept (ADR-0019):** gates **agent identity** (env carries it). Does **not** gate **card ownership** — `add_task` has no env default for the owner, so there is nothing to assert with; the prompt-level `assignee=$SAC_TWIN_PARENT` rule stays load-bearing and untouched.

---

## Verified vs assumed

**Verified by observation:**

- `claude --help`: `--fork-session` ("When resuming, create a new session ID instead of reusing the original (use with --resume or --continue)"), `--session-id <uuid>` ("must be a valid UUID") — read directly.
- SDK `ClaudeAgentOptions` really has `fork_session` / `session_id` / `resume`.
- Emitted TUI argv: `claude --model haiku --resume <parent> --fork-session --session-id <derived>`; SDK argv carries both; a **fresh** session correctly omits them.
- uuid5 is deterministic, valid, round-trips — satisfies `--session-id`.
- Both defects reproduced against the real `scitex-tex` spec.
- `--tag` present on all four surfaces (introspected, not eyeballed).
- Tests: 58 twin + 14 validator + 56 argv green; ruff clean on every file touched here.

**Assumed / NOT verified (called out deliberately):**

- **A twin has not been booted live.** All evidence is spec-derivation + validator + argv level — which is where both bugs lived — but *"the spec now validates" is not "the twin runs"*. One live `sac agents twin <parent> --tag smoke` on the host is worth doing before trusting the feature. It cannot be done from in-container.
- Which duplicate `--env` apptainer honours (last-wins assumed). **Deliberately not relied on** — the fix *deletes* the inherited pairs, so precedence is moot.

---

## Refactor note

`_twin.py` exceeded the 512-line cap, so it splits along the halves its own docstring already named: `_twin_identity` (leaf — naming, gate, errors), `_twin_derive` (command-time), `_twin_seed` (boot-time), with `_twin.py` a thin **facade** re-exporting the public API so ADR-0019's single-import-surface promise holds. Tests mirror the split.

## The verb is untouched

Built on the `twin` substrate under the current name. No `fork` alias (a CLI-only alias would break the 1:1 掟), no rename — both are the operator's call and separate from this PR. All three refinements are identical either way.


ywatanabe1989:

**What this PR attempted:** feat(twin): --fork-session + deterministic --tag ids + boot identity gate (and: twin was inert)

**Why it is being closed:** conflicts with its base; 24 days old, so CI never returned a verdict on it in its current form.

**Nothing is deleted.** Branch `feat/twin-fork-session-tag-assert` still holds every commit, and this PR can be reopened at any time. If the change is still wanted, a fresh PR against today's default branch is the faster path.

PR #733, opened 2026-07-17
feat(binds): fleet-default cards bind for the S6 store migration

ywatanabe1989:
Adds `~/.scitex/cards:/home/agent/.scitex/cards:rw` to the fleet's default binds, unblocking scitex-cards' S6 store migration.

WHY THE EXISTING FULL-HOME BIND IS NOT ENOUGH — this is the whole point:
the store resolver keys off $HOME=/home/agent, NOT the host path. A
`/home/ywatanabe:/home/ywatanabe:rw` bind exposes the store at
/home/ywatanabe/..., which the resolver never consults. Measured 2026-07-16 in
the scitex-cards pilot: without this line `db import` SUCCEEDS into the
container overlay and lands nowhere real — a write that returns 0 and is lost.
(/home/agent/.scitex/todo is a bind, but /home/agent/.scitex itself is
overlay-local.)

NARROW ON PURPOSE: a `~/.scitex`-wide bind would also expose
`~/.scitex/agent-container/accounts`. This binds only the cards store.

EVIDENCE: scitex-cards has run this as a per-agent pilot for 12+ hours with
dev:inode identity confirmed FROM INSIDE the container (not from argv, which
proves nothing — apptainer's skip-if-missing makes a bad bind a SILENT no-op).

VERIFY THE SAME WAY: from inside a rebuilt agent, stat the inode of
/home/agent/.scitex/cards and compare it to the host's. Never verify a bind by
reading the argv.

Refs the scitex-cards S6 migration; their side is ready (0.16.0 on PyPI,
rehearse gate green on live data).

### DIFF
--- a/src/scitex_agent_container/runtimes/_p3a_default_binds.py
+++ b/src/scitex_agent_container/runtimes/_p3a_default_binds.py
@@ -60,6 +60,30 @@ _FLEET_DEFAULT_BINDS: tuple[str, ...] = (
     # P3a-2 — scitex-todo single shared store (operator directive
     # feedback_scitex_todo_single_shared_store).
     "~/.scitex/todo:/home/agent/.scitex/todo:rw",
+    # S6 store migration (scitex-todo -> scitex-cards). Same shape and same
+    # reason as the todo bind above, and the reason is NOT obvious: the store
+    # resolver keys off the AGENT's $HOME=/home/agent, so host-side reach is
+    # not enough. An agent whose spec binds the operator's ENTIRE home rw
+    # still cannot resolve ~/.scitex/cards — the data is present and
+    # unreachable at the same time, because the resolver never looks there.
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
Consumer endorsement from scitex-cards (the shared bot account cannot formally approve its own PR): the bind is the narrow shape we agreed — one store, one line; ~/.scitex-wide correctly rejected for the accounts exposure — and the comment carries the two non-obvious facts future readers need: the resolver keys off $HOME=/home/agent (host-side reach is not enough; my overlay-landing measurement cited), and skip-if-missing means rollout verification MUST be dev:inode from inside, never argv. Exactly how the pilot verified on my spec (dir 2096:5932244, cards.db 2096:3417791, identical both names). Ship it; hosts without ~/.scitex/cards keep silently skipping until their stores migrate, which is the intended S4 sequencing.

PR #741, opened 2026-07-17
fix(to_home): gate the telegrammer MCP on the spec, not on whatever the shell env carries

ywatanabe1989:
## The bug

The shared baseline `.mcp.json` carries a **claude-code-telegrammer MCP server for
every agent**, and the baseline-union merge faithfully preserves it.
`spec.claude.channels` gates the `--dangerously-load-development-channels` **flag** —
it never gated the **server**.

So an agent whose spec says nothing about Telegram still got a telegrammer MCP,
whose config expands `${CCT_BOT_TOKEN}` / `${CCT_AGENT_ID}` from the **shell
environment** at runtime. The spec says no; the agent runs one anyway.

## Why it detonated tonight

Invisible while those vars were empty. Then direnv began loading each project's
`.envrc` in the agent's shell. Three scitex-cards UI agents working in
`~/proj/scitex-todo` picked up that project's `CCT_BOT_TOKEN` **and its
`CCT_AGENT_ID`**, came up announcing themselves as `scitex-todo`, and fought over
the scitex-todo steward's bot — Telegram allows exactly one poller per token.

Measured by reading each poller process's own environment:

```
pid=2675173 owner=scitex-cards-chat    polling_as=scitex-todo
pid=2696432 owner=scitex-cards-gui     polling_as=scitex-todo
pid=2719398 owner=scitex-cards-mobile  polling_as=scitex-todo
pid=2909718 owner=scitex-cards-mobile  polling_as=scitex-todo
```

**The leaked token is the smaller half. The IDENTITY is the defect**: an ambient
file made whoever `cd`-ed into a directory *become* another agent.

That is **not the `.envrc`'s fault** — a project file may reasonably describe its
own bot. The fault is that the **spec was not the authority over what the agent
runs**.

## The fix

One call, at the last point where the agent's real MCP set is decided, keyed on
the **spec** (the SSoT) — never on whether an env var happens to be populated.

A config-less baseline pass is left untouched: it has no spec to consult and the
agent's own pass always follows. Stripping there would be a fallback, and a wrong
one.

## Evidence

- **4 new tests**, AAA, each name states the behaviour.
- **Mutation-proven**: with the gate line disabled,
  `test_agent_whose_spec_omits_the_channel_gets_no_telegrammer_mcp` goes RED, and
  its failure message shows the exact incident payload surviving into an agent
  that never asked for it. A guard nobody has watched fail is a hope with YAML
  around it.
- **Regression: 913 passed / 1 pre-existing failure.** That one
  (`test__sdk_channels.py::test_no_log_when_channel_not_requested`) is
  `SacBinaryNotFoundError` from `sac` not being on PATH in the test shell, and it
  fails **identically on pristine `origin/develop`** with this change absent —
  verified in a detached worktree rather than assumed.

## Sequencing (load-bearing)

The fleet rolling-restart is on hold behind this. Most agents currently run
*without* direnv active, so the trap is unarmed for them; a rolling restart
delivers direnv to all 96 specs and would arm it fleet-wide at once. **Fix → then
restart.** Killing the four pollers was a mitigation with a lifetime of "until the
next restart."

### DIFF
--- a/src/scitex_agent_container/runtimes/_to_home_deployers.py
+++ b/src/scitex_agent_container/runtimes/_to_home_deployers.py
@@ -30,6 +30,7 @@ from pathlib import Path
 from ..config import AgentConfig
 from ._mcp_merge import merge_mcp_json
 from ._mcp_reliability import inject_always_load
+from ._sdk_channels import _TELEGRAMMER_CHANNEL, _TELEGRAMMER_MCP_KEY
 from ._to_home_errors import WorkspaceMcpMergeError
 from ._to_home_text import (
     END_MARKER,
@@ -165,6 +166,7 @@ def _deploy_mcp_merge(
         merged = merge_mcp_json(base_doc, overlay_doc)
     else:
         merged = overlay_doc
+    _drop_unrequested_telegrammer(merged, config=config, dst=dst)
     # Cold-start race fix (fleet incident 2026-07-06): stamp ``alwaysLoad:true``
     # onto the critical stdio MCP servers so Claude Code BLOCKS on startup until
     # they connect (capped at MCP_TIMEOUT) instead of racing the slow fastmcp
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Closing: this is the wrong fix, and the operator is right.

**Why I'm closing rather than iterating.** This PR strips the telegrammer MCP from
agents whose spec didn't request it. That fights an intentional design. Operator:

> シェアードに書かれているエムシーピーは配られて起動される、当たり前じゃんそれでいいでしょ

Correct. A shared baseline MCP being distributed and started for every agent **is
the design**. The defect was never that the server starts — it is that it was
handed *someone else's identity* to start with.

**The actual root**, which the operator named after I'd produced three wrong diagnoses:

> 問題はアイデンティティをプロジェクトルートに紐づけてしまっているから一つのレポジトリから一つが立ち上がっていること

Identity is derived from the **directory**. Location decides who you are, so one
repo = one identity = one agent, *structurally* — and the second agent in a repo
doesn't merely collide, it **becomes** the first. Two offenders, same mistake, and
one is mine:

1. `~/proj/<x>/.envrc` exports `CCT_AGENT_ID` — a project-scoped file carrying an
   agent-scoped fact. (Operator: `cct の場合はエージェント単位なので .envrc に含めては
   いけなかった`.)
2. `_cct_token_pool._slot_candidates(name, workdir)` resolves from the **workdir
   basename first**. sac derives identity from location too. I spent the evening
   blaming the `.envrc` while shipping the same rule.

That single root explains all three of tonight's symptoms — bot contention, card-identity
forgery, fork collision. cct's framing: *"your four siblings were forks-by-cwd."*

**And the legitimate discomfort behind this PR has a better answer**, also the operator's:

> スペックにないものが動くのが気持ち悪いならばスペックに継承の仕組みを作らなきゃですね

Right. If "runs something not in its spec" feels wrong, the fix is to make the spec
**express its inheritance** — not to strip what it inherits. That is the same
`extends:` mechanism already being designed for fork deltas with dotfiles, arrived at
from the other direction. One mechanism, two callers: baseline inheritance and parent
inheritance. Then "not in the spec" stops existing as a state.

**Nothing is lost by closing.** This PR isn't needed to fix the incident: with the
direnv `eval` dropped (dotfiles, +1 given), no foreign token reaches the agent,
`${CCT_BOT_TOKEN}` expands empty, and no poller starts.

Tracked at `sac-cct-slot-collision-siblings-share-one-bot-20260717` (retitled to the
root). The tests here were mutation-proven and the technique carries over to the
inheritance work.

PR #754, opened 2026-07-18
feat(env): fleet-default env layer (spec.env wins) + finish the scitex-cards rename where it is visible

ywatanabe1989:
## Why

Two things land on the same fleet restart, so they ship together.

**The board is slow, and the fix cannot be deployed.** `get_board()` re-parses a ~9 MB tasks YAML at ~4.6 s, and every store write invalidates the cache — that is why the operator's board takes ~30 s. scitex-cards already has the SQLite-mirror fix and proved DB rows byte-identical to YAML across 1985 cards, but it refuses to serve while the mirror is stale, and the mirror goes stale within seconds unless **every** writer dual-writes. Every writer is every agent. So the flag only helps if it reaches the whole fleet at once — and there was no mechanism to say anything to the whole fleet.

**The rename is finished where it is invisible and unfinished where the operator looks.** 「いつまで移行中なんですか？」 The PyPI dist, console scripts and import name are all `scitex-cards`; his screen still says `scitex-todo`.

## Part A — fleet-default env layer

Before this PR the only path from config to container env was `spec.env`, rendered as `--env K=V` in `_apptainer_build_argv`. There was no baseline layer, so a fleet-wide variable meant editing N specs — and the specs live in the dotfiles repo, not here. That is why fleet-wide flags never happened.

New `runtimes/_fleet_env.py`. Precedence, lowest → highest:

1. `FLEET_DEFAULT_ENV` — sac's declared data
2. `config.yaml` → `spec.fleet_default_env` — operator, host scope
3. `spec.env` — per-agent, **always wins**

**The precedence rule and why it does not raise.** `_layer_merge.deep_merge_layers` raises `LayerMergeConflict` when two `to_home` layers assign one key two different scalars. That is correct there — those layers are peers, each key is owned by exactly one of them, and a collision is a mistake. It is wrong here: a default exists *in order to* be overridden, so a per-agent `spec.env` naming a fleet-default key is the feature. This follows the `.envrc` cascade idiom (`_envrc.eval_envrc_cascade`: "a later layer overrides an earlier one") instead. Overrides are logged at INFO so they are visible in the launch log rather than silently effective.

Implementation: `merge_fleet_env` — `src/scitex_agent_container/runtimes/_fleet_env.py:160`. Call site: `_apptainer_build_argv.py:296` (`effective_env(config)` replacing the bare `config.env` loop).

**Neutrality.** The defaults are DATA. No sac logic names a consumer — `merge_fleet_env` merges whatever mapping it is handed and behaves identically with an empty `FLEET_DEFAULT_ENV`. Adding or dropping a default is a one-line data edit. Per-agent opt-out needs no new mechanism: set the same key in `spec.env` (to another value, or `""` to neutralise).

Seeded with `SCITEX_CARDS_DUAL_WRITE=1` and `SCITEX_CARDS_READ_BACKEND=sqlite`.

> **Note on the brief:** it pointed at `config/_host.py` as already having a deep_merge/cascade concept. It does not — that module is purely hostname substitution. The real cascade helpers are `runtimes/_layer_merge.py` (to_home, conflict-raising) and `runtimes/_envrc.py` (env, override-wins). I followed the `_envrc` one because it has the semantics this needs. No pre-existing env-default mechanism existed; this is not a second one.

## Part B — rename residues (every location verified, not assumed)

**`.mcp.json` template: no such thing in this repo.** The brief asked me to flip the emitted server key. sac does not emit a `scitex-todo` server entry anywhere — `examples/agents/full-agent/to_home/.mcp.json` is `{"mcpServers": {}}`, `runtimes/onboarding.py` seeds an empty map, and the `apptainer-*.def` hits are comments. The real `.mcp.json` comes from the operator's `to_home` layers. sac's role is to *look servers up* in a file it does not own — which is exactly why the tolerance below is required rather than optional.

**Contracts with the deployed `.mcp.json` — dual-name tolerance kept:**

| Location | Change |
|---|---|
| `_mcp/_healthcheck.py:57` `CRITICAL_CAPABILITIES` | key → `scitex-cards` |
| `_mcp/_healthcheck.py:73` `SERVER_ALIASES` (new) | `scitex-cards` also matches a `scitex-todo` line |
| `_mcp/_healthcheck.py:125` `parse_mcp_status` | alias-aware; reports under the canonical key |
| `runtimes/_mcp_reliability.py:47` `CRITICAL_MCP_SERVERS` | carries **both** spellings |

**Where I kept tolerance and why.** A live fleet is rolled one agent at a time and the `.mcp.json` is not ours to flip. A hard rename would classify every not-yet-migrated agent's *healthy* board MCP as absent → `UNKNOWN` → alarm: a false alarm manufactured by the rename itself. For `_mcp_reliability` the tolerance is free — `inject_always_load` only stamps names present in the doc, so an unused name costs nothing. Drop the old entries once no deployed `.mcp.json` declares them.

Tolerance does not paper over real state: a failure on the legacy key still reports FAILED, and a genuinely absent server is still UNKNOWN (both asserted). One behaviour change: `run_healthcheck` now reports a legacy-key failure under the canonical name `scitex-cards` — an existing test was updated to match, with the reasoning in its docstring.

**Free-form display strings — straight flip, no tolerance needed** (these are provenance labels rendered into the `meta.source` bracket the operator reads; not lookup keys, not ACL-bearing):

- `_listen/_card_event_delivery.py:210` `from_agent`
- `_listen/_notify.py:76` `DEFAULT_NOTIFY_FROM`

**Whitelist** — `_baseline_assets/hpc_login_hooks/hpc_login_whitelist_policy.py:63` now lists both; safe because both console scripts exist on PyPI and 0.16.2 made the old-name install upgrade-proof.

**Deliberately NOT changed:** `config/_parsers/_skills_defaults.py` `BASE_REQUIRED_SKILLS`. This names a skill *directory*, and the only one on disk is `~/.claude/skills/scitex/scitex-todo` — there is no `scitex-cards` skill. Flipping it would put a nonexistent skill into every agent's CLAUDE.md, which is worse than the stale name. Blocked on the dotfiles skill rename; the sac side is a one-line change once that lands.

## Part C — MCP panel cleanup (added scope)

The operator's panel showed two failures, one apparent duplicate and one stale name.

**1. `plugin:telegram:telegram` — not ours.** sac only *reads* plugin state (`_account/credentials.py` `_read_installed_plugins`, `_state/_meta/quota.py`); it never writes `enabledPlugins` or installs a marketplace plugin. That entry comes from Claude Code's plugin system via the operator's `~/.claude` settings / installed-plugins, which live in the dotfiles layers. **Reporting, not guessing — no sac change can remove it.** Removing it is a one-line dotfiles edit.

**2. `sac` vs `scitex-agent-container` — both intentional, kept.** Not a duplicate. `scitex-agent-container` is `sac mcp start`, the full control-plane MCP (36 tools). `sac` is a *different* server: the `sac mcp channel --name <agent>` stdio channel subscriber, registered only when a spec declares the `server:sac` channel (`runtimes/_apptainer_inner_argv_tui.py:354`, built by `_sac_channel_mcp_server`). Different command, different purpose, and the 5-vs-36 tool counts are exactly what that difference predicts. Removing either would break a live rail, so both stay.

**3. `claude-code-telegrammer` FAILED — fixed.** This was the tokenless-boot case, and it implements the deferred card `sac-omit-telegram-mcp-when-no-cct-bot-token-20260702`. The shared baseline `.mcp.json` declares the telegrammer for every agent with `"CCT_BOT_TOKEN": "${CCT_BOT_TOKEN}"`; an agent with no bot launches it with an empty token, cct correctly refuses, and the panel carries a permanent failed row. That fail-loud is right for a *misconfigured* agent and wrong for a deliberately bot-less one, and once the entry exists the two are indistinguishable — so the fix is to not emit it.

`prune_tokenless_telegrammer_mcp` (`runtimes/_cct_token_pool.py:373`), wired into `_to_home.deploy_to_home:300`. **Ordering is load-bearing:** it must run after the walk materialises `.mcp.json` and after `ensure_cct_bot_token` resolves the pool token. Verified by reading the call site, and asserted end-to-end through `deploy_to_home` — a unit test alone would pass even with the call wired at the wrong point.

## Tests

| Suite | Result |
|---|---|
| `test__fleet_env.py` (new, 22 tests) | 22 passed |
| `test__apptainer_build_argv.py` (+3 fleet-env argv tests) | 51 passed, 2 skipped |
| `test__healthcheck.py` (+6 tolerance tests) | 23 passed |
| `test__cct_token_pool.py` (+8 prune tests) | 40 passed |
| `test__to_home.py` (+3 e2e ordering tests) | 84 passed |
| `test__mcp_reliability.py` | 12 passed |

The wider `runtimes/ _mcp/ _listen/ config/` sweep (3501 tests) was still running when this PR was opened; every suite this PR touches is listed above and green. I will post the sweep result as a comment.

Key names: `test_fleet_default_env_reaches_the_built_argv`, `test_spec_env_overrides_fleet_default_in_argv`, `test_legacy_server_key_still_resolves_under_the_new_name`, `test_genuinely_absent_board_server_is_still_unknown`, `test_tokenless_agent_gets_no_telegrammer_entry`, `test_agent_with_a_token_keeps_the_telegrammer_entry`.

### Mutation proof

Reversed the precedence in `merge_fleet_env` (defaults applied *after* `spec.env`) and re-ran:

```
FAILED test__fleet_env.py::test_spec_env_overrides_a_fleet_default_of_the_same_name
FAILED test__fleet_env.py::test_spec_env_can_neutralise_a_fleet_default_with_an_empty_value
FAILED test__fleet_env.py::test_a_same_key_collision_does_not_raise
FAILED test__fleet_env.py::test_effective_env_reads_config_env
FAILED test__apptainer_build_argv.py::test_spec_env_overrides_fleet_default_in_argv
5 failed, 68 passed, 2 skipped
```

The argv-level test fails too, so the guard covers the layer that actually reaches the container. Restored and re-verified green (73 passed, 2 skipped).

## Incidental

`_apptainer_build_argv.py` was already **at** the 512-line cap, so any edit was blocked. Extracted `runtimes/_apptainer_quota_cache.py` (the constants + resolver — a lookup, not argv assembly, and the only reason that module imported `os`). Both historical import paths still resolve, verified by probe; file is now 508 lines.


ywatanabe1989:
## Wider sweep result

`runtimes/ _mcp/ _listen/ config/` — **3480 passed, 1 failed, 20 skipped** (27m07s).

The one failure is **pre-existing and environmental, not from this PR**:

```
FAILED tests/scitex_agent_container/runtimes/test__sdk_channels.py::TestTelegrammerWakeDiagnostics::test_no_log_when_channel_not_requested
SacBinaryNotFoundError: Cannot resolve the `sac` console script: not on PATH
and not at any known candidate location (/opt/venv-agent/bin/sac).
```

Evidence it is not mine, rather than an assumption:

1. `_sdk_channels.py` is **not in this PR's diff** (`git diff --name-only HEAD~1 HEAD` has no match). This PR only *imports* an existing constant from it, which executes no new code.
2. Reproduced on a **detached worktree at a pre-PR commit** (`87a93f96`), same interpreter: `1 failed, 43 passed` — identical error.
3. The cause is the test host lacking the `sac` console script on PATH (`/opt/venv-agent/bin/sac` absent in this container), which is an environment gap, not a code defect.

Every suite this PR touches is green — see the table in the PR description.

PR #756, opened 2026-07-18
fix(lifecycle): never fall back to host execution silently; report a no-op restart honestly

ywatanabe1989:
## Summary

Both reported defects share one shape: **the system reports success while the thing did not happen.** They turned out to be the *same* mechanism, and the real evidence pins it exactly.

One correction to the brief up front, because it changes what got fixed: **there is no silent apptainer→host fallback in sac.** See "Premise correction" below. What does exist — and what actually produced every symptom in the report — is a `--force` that was silently dropped at the container boundary, turning a RESTART into a no-op that still announced success.

## Defect 1 — an in-SIF restart silently degraded into a no-op that reported success

`agent_restart` calls `agent_start(force=True)` precisely because a restart must REPLACE the process. But the in-SIF broker fires at `_lifecycle/_start.py:156` — **before** the local `force` branch at `_start.py:296` is ever consulted — and `maybe_broker_in_sif_spawn` had no `force` parameter at all. The flag was dropped on the floor.

The host then ran a plain, unforced `sac agents start <name>`, hit the idempotent "already running → no-op" branch, printed `SUCC` and exited 0.

Real evidence, `scitex-storage` runtime dir, `STARTUP_FAILED` 2026-07-12 (`phase=post_ack_liveness`, `kind=post_ack_no_apptainer_pid`, `exit_code=0`) — the child's own stderr, verbatim and in this order:

```
Agent 'scitex-storage' is already running. No-op. Use --force to restart.
SUCC: scitex-storage started (ywata-note-win@/home/ywatanabe/proj/scitex-storage:/work)

[listen post-ack liveness probe] post_ack_no_apptainer_pid: `sac agents start`
returned rc=0 but no apptainer_pid file appeared in
/home/ywatanabe/.scitex/agent-container/runtime/scitex-storage within 5.0s.
```

Those four lines are the whole bug: it declined to touch the agent, then said it started it. And because **no new container was launched, no `apptainer_pid` was ever written** — which is exactly what tripped the post-ack probe. The `post_ack_no_apptainer_pid` alarm was not a separate fault; it was this fault's shadow.

The dropped-`force` chain, file:line:

| Hop | Location | Was |
|---|---|---|
| caller | `_lifecycle/_stop.py:423` | `agent_start(..., force=True)` ✓ |
| broker call | `_lifecycle/_start.py:156` | `force` **not passed** |
| chokepoint | `_lifecycle/_in_sif_broker.py:242` | no `force` parameter |
| wire | `_lifecycle/_spawn_client.py:245` | no `force` field |
| host handler | `_listen/_agent_exec.py:219` | never appended `--force` |
| host start | `_lifecycle/_start.py:337` | no-op → `return True` |

## Defect 2 — the no-op reported itself as a successful restart

`agent_start` returned a bare `True` from **both** the branch that really launched a container and the branch that found the agent already alive and did nothing. Those two returns were byte-identical, so no caller could distinguish them. `cli_pkg/lifecycle/_restart.py:173` read `agent_restart(name) is not False` → `True` → `{"name":"scitex-storage","restarted":true,"dispatched":false}`, rc=0, over a process whose pid never changed.

## What changed

1. **`force` propagates end to end** — `agent_start` → `maybe_broker_in_sif_spawn` → `broker_start_to_host` → `request_spawn` (wire field `force: true`) → the host `/agents` handler, which appends `--force` to its inner argv. Emitted only when truthy, matching the existing `foreground`/`one_shot`/`assume_yes` precedent, so a pre-fix host ignores the absent field and an ordinary brokered start keeps its idempotent behaviour. Non-boolean `force` is rejected with 400 rather than coerced.

2. **The no-op names itself** — new `_lifecycle/_start_outcome.py`; the no-op branch returns `NOOP_ALREADY_RUNNING` instead of bare `True`. It stays **truthy** on purpose: re-running `sac agents start` over a live agent IS a legitimate success, and downgrading it to `False` would invent the mirror-image lie. It just now carries WHY.

3. **`sac agents restart` reports honestly** — a start leg that no-op'd yields `restarted: false`, `reason: "already-running"`, and a `hint` naming the `--force` recovery, plus a non-zero exit code. Human output gets its own branch so the operator is not sent to the wrong recovery (`kill-session` + `--fresh`) for an agent that is perfectly healthy and simply never cycled.

### Backward compatibility

`agent_start` is declared `-> bool` and its result is consumed by a long tail of callers. `bool` cannot be subclassed, but `bool` IS an `int` subclass, so `StartOutcome(int)` keeps every existing form working — verified explicitly in the tests:

- `bool(NOOP_ALREADY_RUNNING)` → `True` (`if result:` unchanged)
- `NOOP_ALREADY_RUNNING is not False` → `True` (the restart CLI's deliberate identity check, which exists so a runtime returning `None` isn't read as failure, keeps its meaning)
- `NOOP_ALREADY_RUNNING == True` → `True`

`outcome_kind` returns `None` for plain `True`/`False`/`None`, so it is safe on any older or hand-rolled start result.

**One idiom does NOT survive, and CI caught it — correcting my earlier claim.** `bool` cannot be subclassed, so no object can ever BE the `True` singleton: `result is True` fails. My first pass asserted the change was zero-ripple; that was wrong, and all three pytest-matrix jobs failed deterministically on exactly two tests:

```
test_lifecycle.py::test_agent_start_idempotent_when_already_running
test__start_verdict.py::test_an_alive_agent_no_op_returns_success
  AssertionError: assert StartOutcome(1, kind='already-running') is True
```

No production code uses the identity form on a start result — the blast radius is those two tests. Both assert **success**, not object identity (one is named `..._no_op_returns_success`; the other's comment reads "returns success and never calls start"), so both were rewritten to `bool(ok) is True` **and** `outcome_kind(ok) == KIND_ALREADY_RUNNING`.

That is **stricter** than before, not looser: they now also pin *which* branch produced the success — the distinction a bare `True` made impossible, which is the whole bug. Mutation-proved: reverting `_start.py` to the original `return True` turns **both** RED, whereas the original `is True` versions passed against exactly that buggy behaviour. The limitation is now pinned by its own test (`test_no_op_is_NOT_identical_to_the_True_singleton`) so the next reader meets it in the suite, not in a red pipeline.

**Response-shape callers.** The only production reader of `restarted` is `cli_pkg/lifecycle/_restart.py:128` (the cross-host peer verdict), unaffected. `reason`/`hint` are **added** keys on the no-op path only; no existing key changed meaning on any path that previously succeeded. No shell script, doc, or YAML reads either key — every repo-wide hit is English prose. One existing test, `test__restart.py::test_local_restart_json_envelope_marks_not_dispatched`, asserts `restarted is True` for the local path; it stubs `agent_restart` to return plain truthy, so it still passes (it pinned the incident payload as correct precisely because it also could not tell a no-op from a real restart).

## Premise correction — no host fallback exists

The brief expected a failed in-SIF boot to silently run the agent host-side. I traced every start path and **that path does not exist**; I am not inventing a fix to match the hypothesis.

- `_lifecycle/_runtime_select.py:57-70` returns only `TuiSessionRuntime` or `ClaudeSessionRuntime` — both container runtimes — and **raises** `ValueError` on an unknown value. No host arm.
- Every apptainer-absent branch is fail-closed: `runtimes/_apptainer_runtime.py:161` **raises** when `apptainer` is not on `$PATH`; `:140` returns `None` → `start` returns `False`; `runtimes/tui_session.py:273-284` raises `TuiAuthStageError`. `argv[0]` is the literal `"apptainer"` unconditionally (`_apptainer_build_argv.py:82`).
- The one host-fallback-shaped function — `_runners/_tmux/claude_code.py:598-651` (`ClaudeCodeRuntime`, an `if/elif/elif` with no `else` falling through to a host tmux launch with host-venv activation) — is **unreachable**: nothing constructs it, and its three imports (`.apptainer`, `.docker`, `.podman`) name modules that **do not exist**, so reaching it would `ImportError`, not silently host-run.

`scitex-storage` was therefore never moved to the host by sac. What happened is what the log says: the restart no-op'd and announced success, so it stayed whatever it already was.

Two other brief items were **already implemented** and needed no change:

- **`post_ack_no_apptainer_pid` already fails loud** — `_listen/_agent_exec.py:363-390` writes a `STARTUP_FAILED` marker **and** returns HTTP 502. It is an alarm, not a logged note.
- **Booting does not require a Python workdir.** The `uv pip install -e .` error came from the agent's **own** `spec.yaml` `startup_commands` (`~/.scitex/agent-container/agents/scitex-storage/spec.yaml:71`), not from sac source — sac only sets `UV_PROJECT_ENVIRONMENT` (`_apptainer_listen_env.py:113`) and never generates that bootstrap. The `sac agents create` template already guards its install with `|| true` (`cli_pkg/_create.py:228`). The failing spec line is guarded only on the venv's existence, not on the target being a Python project. That is per-agent spec data and is fixed by adding a guard to the spec (`[ -f pyproject.toml ] && uv pip install -e . || true`) — flagging it rather than silently rewriting operator specs from here, since ~25 live specs share the pattern.

## Tests — 26 new, every one mutation-proved

| File | Covers |
|---|---|
| `_lifecycle/test__start_outcome_no_silent_noop.py` (10) | no-op is truthy, survives `is not False` / `== True`, carries its kind, is distinguishable from a real start, safe on legacy returns |
| `_lifecycle/test__in_sif_broker_force_propagation.py` (8) | `force` on the wire, absent by default, broker signatures, `agent_start` forwards it; **end-to-end** host-handler argv via `TestClient` + a real fake `sac` on `$PATH` (no mocks), and 400 on non-boolean |
| `cli_pkg/lifecycle/test__restart_noop_reports_false.py` (8) | no-op → `restarted: false` + `reason` + `hint` + non-zero exit; real restart still `true`; `None` not read as failure; explicit `False` still fails |

**Mutation proof** — each fix reverted individually, confirming the tests detect the old behaviour, then restored:

```
M1: agent_start no-op returns bare True again (the original bug)  -> RED
M2: broker call drops force again (the root cause)                -> RED
M3: spawn client stops putting force on the wire                  -> RED
M4: host handler stops appending --force to the inner argv        -> RED
M5: restart CLI stops downgrading the no-op to restarted=false    -> RED
RESTORED baseline                                                 -> GREEN (26 passed)
```

M4 is worth calling out: it first came back **GREEN**. My original assertion checked that `'inner_argv.append("--force")'` appeared in the module source, so mutating the guard to `if False:` left the string in place and the test passed over dead code — an assertion that could not disagree with itself. It was rewritten to drive the real handler over HTTP and read back the argv a real fake `sac` binary recorded, which turns it RED. The weak version is called out in that class's docstring so it does not come back.

## Verification notes

- Run with an absolute interpreter (`/opt/venv-sac/bin/python -m pytest`); the repo's `.venv` is a broken 3.11 symlink. `pyproject.toml` `pythonpath = ["src"]` plus `tests/conftest.py::pytest_sessionstart` guarantee the checkout — not the stale `/opt/venv-sac` copy — is the package under test.
- `tests/scitex_agent_container/_lifecycle/test__broker_self.py` hangs in this container. **Pre-existing and unrelated**: it hangs identically on a clean tree with my changes stashed, it spawns a real `sac listen` subprocess, and it references none of the symbols this PR touches (grep: 0 matches).

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_in_sif_broker.py
+++ b/src/scitex_agent_container/_lifecycle/_in_sif_broker.py
@@ -127,6 +127,7 @@ def broker_start_to_host(
     foreground: bool = False,
     one_shot: bool = False,
     assume_yes: bool = False,
+    force: bool = False,
 ) -> dict:
     """POST a spawn request to the host-side ``sac listen``; FAIL LOUD on error.
 
@@ -216,6 +217,7 @@ def broker_start_to_host(
             foreground=foreground,
             one_shot=one_shot,
             assume_yes=assume_yes,
+            force=force,
         )
     except SpawnRequestError as exc:
         # Re-throw under the broker's own error type so the integration
@@ -240,6 +242,7 @@ def maybe_broker_in_sif_spawn(
     foreground: bool = False,
     one_shot: bool = False,
     assume_yes: bool = False,
+    force: bool = False,
 ) -> bool:
     """Single-call broker chokepoint for the in-SIF redirect in agent_start.
 
@@ -282,6 +285,25 @@ def maybe_broker_in_sif_spawn(
     still-alive Popen pid, returning SUCC while the capsule dies
     silently later.
 
+    ``force``: propagate the caller's own ``--force`` across the broker
+    boundary so the host's ``/agents`` handler appends ``--force`` to its
+    inner ``sac agents start`` argv.
+
+    THIS FIELD IS LOAD-BEARING FOR RESTART CORRECTNESS (incident
+    2026-07-12, scitex-storage). ``agent_restart`` calls
+    ``agent_start(force=True)`` precisely because a restart must REPLACE
+    the process; but this broker fires BEFORE that force is ever consulted
+    locally, so before this parameter existed the flag was silently
+    dropped here. The host then ran a plain, unforced ``sac agents start
+    <name>``, hit the idempotent "already running → no-op" branch
+    (``_start.py``), printed ``SUCC: <name> started`` and exited 0. The
+    restart reported success over an agent that never cycled — same
+    process, same pid, same stale credentials — which is the precise class
+    of lie ``_stop.py``'s force/gate comments already fight on the LOCAL
+    path. It also explains the ``post_ack_no_apptainer_pid`` that followed:
+    no new container was launched, so no ``apptainer_pid`` was ever
+    written.
+
     ``assume_yes``: propagate the caller's own ``-y``/``--yes`` consent
     through to the host's ``/agents`` handler (bug fix 2026-07-05,
     paper-scitex-clew report). The host handler shells a fresh
@@ -303,6 +325,7 @@ def maybe_broker_in_sif_spawn(
         foreground=foreground,
         one_shot=one_shot,
         assume_yes=assume_yes,
+        force=force,
     )
     rc = result.get("returncode") if isinstance(result, dict) else None
     if rc != 0:
--- a/src/scitex_agent_container/_lifecycle/_spawn_client.py
+++ b/src/scitex_agent_container/_lifecycle/_spawn_client.py
@@ -228,8 +228,7 @@ def _http_error_message(child_name: str, status: int, parsed: Any) -> str:
             f"check_spawn ACL denied this caller. Server said: {parsed!r}"
         )
     return (
-        f"spawn of {child_name!r} rejected: listen returned HTTP "
-        f"{status} ({parsed!r})"
+        f"spawn of {child_name!r} rejected: listen returned HTTP {status} ({parsed!r})"
     )
 
 
@@ -246,6 +245,7 @@ def request_spawn(
     foreground: bool = False,
     one_shot: bool = False,
     assume_yes: bool = False,
+    force: bool = False,
 ) -> dict:
     """POST a spawn request to the host listen server; FAIL LOUD on error.
 
@@ -311,6 +311,26 @@ def request_spawn(
         at the top of the call chain. This does NOT weaken the
         human-at-a-TTY default-refuse safety net — it only lets the
         brokered/automated path assert consent that was already given.
+    force
+        Forwarded as ``force: true`` in the POST body when set. The host
+        listen's ``/agents`` handler appends ``--force`` to its inner
+        ``sac agents start`` argv, so a still-running agent is TORN DOWN
+        and replaced instead of hitting the idempotent "already running →
+        no-op" branch.
+
+        Silent-degradation fix (incident 2026-07-12, scitex-storage): a
+        RESTART issued from inside a SIF reaches ``agent_start(force=True)``,
+        which brokers to the host — and before this field existed, ``force``
+        was DROPPED at that boundary. The host then ran a plain
+        ``sac agents start <name>``, saw the agent already up, no-op'd,
+        printed "SUCC: <name> started" and exited 0. The restart reported
+        success while nothing whatsoever cycled: same process, same pid,
+        same stale credentials. Because no NEW container was launched, no
+        ``apptainer_pid`` file appeared either, which is what tripped the
+        listen's ``post_ack_no_apptainer_pid`` probe.
+
+        Back-compat: only emitted when truthy, so a pre-fix host simply
+        ignores the absent field and behaves exactly as before.
 
     Returns
     -------
@@ -353,6 +373,13 @@ def request_spawn(
     # one_shot above — pre-fix brokers simply ignore an absent field.
     if assume_yes:
         body["assume_yes"] = True
+    # Silent-degradation fix (incident 2026-07-12): carry the caller's
+    # ``force`` across the broker boundary. Dropping it turned an in-SIF
+    # RESTART into a plain host-side start that no-op'd over the live agent
+    # and still reported success. Same truthy-only back-compat rationale as
+    # the fields above.
+    if force:
+        body["force"] = True
 
     payload = json.dumps(body).encode("utf-8")
     url = f"{base}/agents"
--- a/src/scitex_agent_container/_lifecycle/_start.py
+++ b/src/scitex_agent_container/_lifecycle/_start.py
@@ -26,6 +26,7 @@ from ._spawn_gate import enforce_spawn_gate, persist_acl_policy
 # Re-exported from _start_failure_diag for back-compat: this helper lived here
 # before the 512-line-cap split.
 from ._start_failure_diag import _format_boot_stderr_section  # noqa: F401
+from ._start_outcome import NOOP_ALREADY_RUNNING
 
 # Re-exported from _start_preflight for back-compat: callers/tests import these
 # pre-flight helpers from _start. ``_verify_real_liveness`` is no longer the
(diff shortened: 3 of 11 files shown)


ywatanabe1989:
### CI is green — plus one flaky test worth recording (not caused by this PR, but not papered over either)

**Green now:** all three required jobs pass on `11a5f12e` — `pytest-matrix-on-ubuntu-py3.11` (8m35s), `py3.12` (8m25s), `py3.13` (8m48s), alongside ruff / import-smoke / rtd-sphinx / quality-audit / no-hosted-runners-guard / CLAssistant.

**The real defect this PR had, and its fix.** The first CI round failed on all three Pythons, deterministically, and it was mine: my backward-compatibility analysis for the `StartOutcome` sentinel covered `if x:`, `x is not False` and `x == True` but missed `x is True`. `bool` cannot be subclassed, so no object can ever BE the `True` singleton. Two tests used that identity form:

```
test_lifecycle.py::test_agent_start_idempotent_when_already_running
test__start_verdict.py::test_an_alive_agent_no_op_returns_success
```

Both assert **success**, not object identity, so both were rewritten to `bool(ok) is True` **and** `outcome_kind(ok) == KIND_ALREADY_RUNNING` — strictly stronger than before, because they now also pin *which* branch produced the success, the distinction a bare `True` made impossible and the absence of which is the whole bug. Mutation-proved: reverting `_start.py` to `return True` turns **both** RED, whereas the original `is True` versions passed against exactly that buggy behaviour. No production code uses the identity form on a start result.

**The separate flake.** After that fix, py3.11 and py3.12 went green and py3.13 failed on something unrelated:

```
tests/scitex_agent_container/cli_pkg/_helpers/test__agent_list_perf.py::test_ports_come_from_a_single_list_claims_call
[gw1] assert calls == {"list_claims": 1, "get_port": 0}
  Differing items: {'get_port': 1} != {'get_port': 0}
1 failed, 11394 passed, 78 skipped
```

Evidence that this is nondeterministic rather than a defect of this branch:

1. **A re-run of the identical commit passed.** Same code in, different result out — that is nondeterminism by definition.
2. **It failed on 1 of 3 Pythons.** The genuine breakage above failed 3 of 3, deterministically. A deterministic code fault does not pick one interpreter.
3. **The test references none of the symbols this PR touches** (`force`, `agent_start`, `agent_restart`, `request_spawn`, broker, outcome) — 0 matches.
4. **It passes locally when run immediately after this PR's new tests in one process** (18 passed), so there is no deterministic leak from the new `TestClient` tests into it.
5. **`develop` is green** on all recent runs (#757 release, #754, #753).

**Why it is fragile, for whoever picks it up.** The test swaps `port_allocator.get_port` as a **module global** and then asserts an **exact call count of 0**. Any concurrently-live code in that worker process that reaches `get_port` through module-attribute access — `_listen/_forward.py:39`, `_listen/_registry_endpoints.py:81` — lands in the counter. An exact-count assertion over a shared mutable global is order- and timing-sensitive by construction, which is the same shape as the leaked-daemon-thread pollution already fixed once in this repo (`7ef212a7`, "a leaked daemon thread was polluting other tests' store"). Adding tests re-rolls the xdist distribution, so a new PR can surface it without having caused it.

I have not touched that test — bending it to fit would be exactly the wrong move, and it is not this PR's to fix. Flagging it so it gets a real fix (worker-local counter, or membership rather than exact count) instead of being buried under a green tick.

PR #759, opened 2026-07-18
feat(pr): card-per-open-PR timer with a tri-state that cannot report success while blind

ywatanabe1989:
Makes the operator's PR-freshness policy a **mechanism** instead of a rule someone has to remember.

35 open PRs accumulated untracked until 31 were force-closed **by hand** on 2026-07-18. The operator's diagnosis named three failures — 「カードになってないか追跡できてない、催促できてないのも悪い」 — and the **first** is the one nothing owned.

## Job 1 (ships): `sac.sync-pr-cards`

Upserts **one** scitex-todo card per open PR (title, author, age, draft state, CI status), keyed by PR number so it is idempotent. Completes the card when the PR merges or closes. Federated as a 30-minute systemd timer, so it runs without anyone remembering it.

It builds **no nudge logic**, deliberately. scitex-todo's stale-active sweep already nudges the owner of any untouched open card (it nudged one agent 24 then 26 cards in a single night, unprompted). A second nudger with independent state would race it — the double-supervisor class. **sac owns PR facts; scitex-todo owns the reminder.** The gap was that an un-carded PR is invisible to that rail entirely.

New surface: `sac pr sync-cards [--check|--apply] [--repo owner/name] [--json]`.

## Job 2 (deliberately NOT shipped): the 3-day expiry

The shelf life is a **fleet-wide** rule — 「3日ルールは全てのレポジトリで共通です」 — that scitex-dev owns as a shared primitive, under the standing rule that dev holds primitives and leaves consume them. sac implementing its own staleness decision would **fork** that primitive, and two repos hand-rolling "what does stale mean" makes the fleet's answer depend on which tool you asked.

**Measured, not assumed:** scitex-dev 0.31.1 has no PR-staleness primitive (it has `ecosystem prune-merged`, `cron ci-watch`, `branch-protection` — nothing that adjudicates shelf life). So sac ships **nothing** for expiry: no verb, no JobSpec, no timer, **and no placeholder**. A scheduled job calling a primitive that does not exist is a red timer, and "we'll swap it later" is how this host ended up with five differently-versioned sac installs.

An earlier draft of this branch *did* implement it locally. It was removed rather than kept. The design is preserved in `_prlifecycle/_expiry_seam.py` and on card `sac-pr-expiry-consume-dev-primitive` (blocked on dev), including the four requirements sac will hold the primitive to:

1. Dry-run by default; explicit `--apply`.
2. **The intent-registry write must SUCCEED before any close proceeds.** A close whose record failed is an unrecoverable deletion. The 2026-07-18 hand pass wrote intent to a manifest card before closing 31 PRs but never made the close *conditional* on that write succeeding — the ordering was luck, not design.
3. Scheduled materialisation, never anything depending on someone remembering.
4. Tri-state verdict.

Also proposed there: measure from `updatedAt`, not `createdAt`. A long-lived PR pushed to an hour ago is fresh work; a 4-day-old PR nobody touched is the rot. Measuring from `createdAt` closes the former and spares the latter — the inverse of the ruling.

## The tri-state — the point of the exercise

`0` clean · `1` action needed · `2` **could not determine**.

An empty open-PR list and an unreadable one are **different facts**. The bug refused here is `_authheal/_pass.py::PassOutcome.exit_code()`, which ends in a bare `return 0`: detection failing produces no reports, no reports fall through to that return, and five systemd timers reported success every ten minutes while an agent sat login-expired for hours.

Rendering an unreadable backlog as an empty one would show a 35-PR pile-up as a clean board — and worse, would **complete every card**, since completion is inferred from a PR's absence.

Structural guards, not just a check:

- The `gh` seam returns `GhInvocation`, **not `str`**. A string cannot carry "I failed", which is precisely how the collapse gets built. (The pre-existing `_lifecycle/_github_ci.py::list_open_prs` has this defect: every failure returns `[]`.)
- `PRFetch.readable` is a **whitelist of exactly one state**, so any `FetchState` member added later defaults to UNKNOWN — never to clean. This is the inverse of a fallthrough.
- `gh` exit 0 with **blank** stdout is `UNPARSEABLE`, not empty — a genuinely empty repo prints `[]`.
- Card completion is gated on a proven-readable fetch **and** a readable board.
- Sweeping zero repos is UNKNOWN: "I examined nothing" is not "nothing is wrong".

### Mutation-proven

Replacing `exit_code()` with the naive form turns **3 tests RED**, including:

```
E   assert {0, 1} == {0, 1, 2}
E     Extra items in the right set: 2
```

— the tri-state visibly collapsing to a bi-state. The suite also carries `_NaiveOutcome`, the incident's logic reconstructed, and asserts this suite's own criteria reject it, so the gate stays honest if someone weakens the assertions later.

Verified **live** end-to-end, not only in unit tests:
- `sac pr sync-cards --check` → exit **1**, 6 real PRs read (ages 0.0d–2.6d)
- same command with `gh` removed from `PATH` → exit **2**, `UNKNOWN [no-client]`, no card touched

## Absolute `sac` path in the JobSpec

Five sac installs were measured on this host (0.21.24 / 0.21.22 / 0.21.21 / 0.21.11 / none), and which one a bare `sac` resolves to depends on the invocation form. systemd units get a minimal PATH that is not the operator's login PATH, so a bare `sac` can silently run months-old code while reporting the same success. `sac_bin()` resolves `$SAC_BIN` → the release path → `which sac` → `"sac"`, per call.

The five older JobSpecs still spell a bare `sac`. That is a deliberate non-drive-by: changing running timers' commands is a behaviour change deserving its own PR and verification.

## Tests

91 new, **no mocks**. Real temp scitex-todo store written and read back through the real API; a recorded 19KB `gh pr list` response committed under `fixtures/`. The recording immediately corrected a wrong belief: an `IN_PROGRESS` CheckRun carries an **empty** `conclusion`, so conclusion alone cannot distinguish pending from success — a hand-built fixture would have encoded the mistake.

## Notes for the reviewer

- One flaw found and fixed during live verification: a *missing* store logged "BOARD could not be read", conflating "no cards yet" with "unreadable" — the same conflation this PR exists to prevent. Now `FileNotFoundError` → `{}` (genuine first run), other failures → `None` (unknown).
- Pre-existing and **not** touched here: `scitex_todo` is deprecated in favour of `scitex_cards` (8+ sac modules use the old alias); `ruff check src/` reports 66 pre-existing errors outside these files. Both deserve their own sweeps.


ywatanabe1989:
Closing under the freshness rule — the work is sound, the reconciliation cost is not, and there is a structural prerequisite in front of it.

## Why

develop has moved ~10 merges since this branch was cut (gitlink removal, rescue removal, 0.22.0, tags abolition, the bake stdin fix, auth-audit, agentstate). Re-merging produced conflicts in THREE files with six hunks — `_jobs_plugin.py` (2), `cli_pkg/_main.py` (1), `tests/.../test__jobs_plugin.py` (3). All of them are the same shape: this branch and develop each APPEND a different entry to the same list. Individually trivial, collectively a tax that will be re-levied on every future merge while develop keeps moving.

## The actual blocker, and it outlives this PR

Resolving the merge pushes `src/scitex_agent_container/_jobs_plugin.py` to **524 lines against the repo cap of 512**. The pre-edit hook blocks it, correctly. So this is not "one more conflict" — **no new job can be added to that plugin until it is split**, and this PR happens to be the first to hit the wall.

That is worth fixing properly rather than suspending the limit to squeeze one job through. The module is now eight jobs, each carrying a substantial rationale comment (deliberately — those comments are why the jobs are understandable), so it is a natural extraction: one module per job, or a `_jobs/` package with the plugin left as a thin aggregator that preserves the `scitex_dev.jobs` entry point.

## Nothing here is lost

The branch `feat/pr-lifecycle-timers` remains. Its content that still matters:
- `sac.sync-pr-cards` — one card per OPEN PR, keyed by PR number so it is idempotent, completed when the PR merges or closes.
- Its TRI-STATE discipline, which is the good part: an unreadable PR list reports UNKNOWN and exits 2 rather than seeing an empty list and reporting a clean board — and it refuses to COMPLETE any card on such a pass, since completion is inferred from absence and a blind pass would otherwise complete every card at once.
- The deliberate NON-implementation of `sac.close-expired-prs`, with its reasoning: the 3-day PR shelf life is a fleet-wide rule scitex-dev owns as a primitive, and sac hand-rolling its own staleness decision would fork it.

## Order to re-do it in

1. Split `_jobs_plugin.py` (carded separately). 
2. Re-cut this job on top — it is a single `JobSpec` plus its command, so it is cheap once there is room for it.

Also worth noting the problem it solves is less acute than when it was written: it cites 35 open PRs accumulating until 31 were force-closed by hand. The queue is currently down to a handful, so this is no longer urgent — which is exactly why it should wait for the clean foundation instead of being forced in.

PR #761, opened 2026-07-18
fix(security): make the creds-watch change key see a same-size token rotation

ywatanabe1989:
## The bug

`_account/creds_watch.py:96` returned `(st.st_mtime, st.st_size)` — a change-detection **equality** key built from a float-second mtime. Two writes inside one timestamp granule with the same length compare equal, so the watcher concludes nothing changed.

That is not hypothetical for this file. **A rotated token bundle is the same length as its predecessor** — measured on this fleet:

| file | bytes |
|---|---|
| `~/.claude/.credentials.json` | 1102 |
| `.credentials.json.bak` (its immediate predecessor) | 1102 |
| `.credentials.json.bak-expired-20260714-025724` | 1102 |

So `st_size` cannot see a rotation at all, leaving the timestamp as the only real term. A shared-account `refreshToken` is single-use — one rotation invalidates every co-tenant's in-memory token — so a watcher that can miss a rotation is a watcher that cannot warn anyone. `inotifywait` is absent in the container, so `watch_poll` (and this key) is the *active* detection path here, not a fallback corner case.

## The fix

`(st_mtime_ns, st_size, st_ino)`. Each term is load-bearing:

- **`st_mtime_ns`** — `st_mtime` is a lossy conversion. At a 2026 epoch a double's ulp is ~477 ns, so writes closer than that collapse to one value.
- **`st_size`** — the only discriminator left on a coarse-timestamp filesystem, where `st_mtime_ns` is nanosecond-*typed* but not nanosecond-*accurate* and its low digits are zeros.
- **`st_ino`** — `(st_mtime_ns, st_size)` alone was **not** enough. Both credential writers we control (`active_login_write.sync_active_login`, `creds_sync._atomic_copy`) write tmp + `os.replace`, so the inode moves on exactly the atomic rotation where the other two terms are least helpful.

**Residual, documented in the docstring rather than papered over:** inode numbers are recycled — on ext4 a freed inode goes straight back to the next allocation (measured: `os.replace` cycled between two inode numbers). This narrows the hole; it does not close it.

## Mutation evidence

Each term has a test that goes RED when only that term is dropped.

**Against the original shipped key `(st_mtime, st_size)`:**

```
tests/scitex_agent_container/_account/test_creds_watch.py .FF.           [100%]
E   assert (1784352999.0, 1102) != (1784352999.0, 1102)   # mtime test
E   assert (1784352999.0, 1102) != (1784352999.0, 1102)   # inode test
================== 2 failed, 2 passed, 9 deselected in 1.60s ===================
```

**Against `(st_mtime_ns, st_size)` — i.e. the ns fix without the inode:**

```
E   assert (1784352999000000000, 1102) != (1784352999000000000, 1102)
================== 1 failed, 3 passed, 9 deselected in 0.27s ===================
```

**Against `(st_mtime_ns, st_ino)` — i.e. dropping size:**

```
E   assert (1784352999000000000, 14656241) != (1784352999000000000, 14656241)
================== 1 failed, 3 passed, 9 deselected in 0.20s ===================
```

**With the real fix:** `13 passed in 0.69s`.

## On the test method (why the obvious test is not written here)

A test that writes the file twice with a sleep between them **passes under the buggy code**. This filesystem (ext4) reports *true* nanosecond mtimes — four back-to-back same-size writes gave 4/4 distinct `st_mtime_ns` **and** 4/4 distinct `st_mtime` floats — so both the old and new keys distinguish two ordinary writes and the naive test proves nothing. That is how this defect survived.

The tests therefore pin timestamps with `os.utime(ns=...)` and state the intended collision directly: deterministic, no sleep, no flake. The 1 ns delta is provably below the ~238 ns half-ulp, so the float collision is guaranteed rather than hoped for.

## Not persisted

The key is held only in `watch_poll`'s local `last`; nothing serializes it. On restart the loop runs an unconditional `run_sync_once` and then establishes a fresh baseline, so there is no cross-restart comparison and no old-float-vs-new-ns format mismatch to guard against.

## Scope

Only this equality key. The ~20 other bare `st_mtime` uses in sac are sorting (`key=lambda p: p.stat().st_mtime`) and age arithmetic (`_now() - mtime`), where float granularity is harmless — only equality turns granularity into correctness. `_liveness_tick_resolve.py` and `_verdict_state.py` are deliberately untouched.

Found by sweeping sac after scitex-cards hit the identical defect in their board cache.

### DIFF
--- a/src/scitex_agent_container/_account/creds_watch.py
+++ b/src/scitex_agent_container/_account/creds_watch.py
@@ -10,8 +10,9 @@ Two watch backends, picked at runtime:
 
 * **inotify** via the ``inotifywait`` binary when it is on ``$PATH`` —
   event-driven, zero idle CPU.
-* **poll** fallback otherwise — stat the file's ``mtime``/``size`` on a
-  short interval.
+* **poll** fallback otherwise — stat the file on a short interval and
+  compare a ``(mtime_ns, size, ino)`` change key (see
+  :func:`_signature` for why all three terms are load-bearing).
 
 Both call the same engine, so behaviour is identical; only the
 wake-up mechanism differs. Each sync attempt is logged (stderr by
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
## Production evidence, observed after this PR was opened

While this PR was in flight the fleet took a simultaneous 401 from a shared-account token rotation, and my own session was restarted by auth-heal. That meant I measured the same credential file on both sides of a **real** rotation, by accident:

| | inode | size | mtime |
|---|---|---|---|
| before | 14667439 | **1102** | 2026-07-18 13:26:26 |
| after | 14915160 | **1102** | 2026-07-18 19:31:28 |

A real production rotation: **the size did not move, the inode did.** That is the argument for both `st_size` being insufficient and `st_ino` being load-bearing, confirmed in the wild rather than in a fixture.

The storm shape is also visible. Two account stores rotated in the same second:

```
ywata1989-gmail-com     19:31:28.306565439   size=1102
ywatanabe-scitex-ai     19:31:28.845966627   size=1102
```

539 ms apart, both 1102 bytes. All three account snapshots on this host are 1102 bytes. These are two different files, so they are not the same watcher key — but it establishes that rotation writes genuinely arrive sub-second and same-size, which is the shape the old float-second key could not represent.

One thing this does **not** show: I never observed the old key actually miss a rotation in production. The mtime moved by hours in the instance above, so the old key would have caught that one. The claim this PR makes is narrower and is the one the mutation tests prove — the key must be able to represent a same-size write inside one timestamp granule, and the old one could not.

PR #776, opened 2026-07-18
refactor(lifecycle)!: abolish the pre-stop rescue autosave entirely

ywatanabe1989:
Operator ruling 2026-07-19: 「rescue 一切やめましょう」. The fleet-default
pre-stop rescue is removed entirely.

## What it was

Before stopping an agent, `_lifecycle/_stop.agent_stop` called
`run_pre_stop_rescue`, which walked the agent's worktrees and committed any
dirty tree (`git add -A` + commit, message `rescue: pre-stop autosave
<agent>@<UTC>`), so uncommitted work survived a restart.

## Why it had to go, not be patched

**Its central contract was enforced against the wrong verb.**
`_pre_stop_rescue_git.py`'s docstring stated: *"There is deliberately NO push
primitive here… the whole point is that no code path can publish on the
agent's behalf."* Yet rescue commits are reachable from `origin/develop`
right now.

The leak was never a push, which is exactly why the no-push guard never
caught it. Verified with `git branch -a --contains`:

| commit | branch it landed on | |
|---|---|---|
| `8e70ca8b` | `rescue/scitex-agent-container-20260718T112001Z` | correct — side branch |
| `5340014c` | `fix/restart-preflight-auth-before-stop` | **feature branch** |
| `1042139e` | `feat/a2a-default-communicate-and-role-visibility` | **feature branch** |

When the agent sat on a NON-protected topic branch, `rescue_worktree`
committed **in place**. That branch later became a PR and was merged
normally, carrying the rescue commit into `develop`. A guard on `push`
cannot stop a commit riding a legitimate `merge`.

**The damage was real and expensive.** Rescue commit `37d83977`
(2026-07-01) — an ancestor of BOTH `develop` and `main` — committed nine
`mode 160000` gitlinks under `.tmp-audit/` with no `.gitmodules`, breaking
`actions/checkout` on every run of every workflow until PR #769 removed
them. That is the generic failure mode, not bad luck: a broad `git add -A`
over an agent's dirty worktree sweeps up whatever happens to be sitting
there.

**Why a targeted fix (always use a side branch) was not preferred.** It
would close this leak, but the module's history is four hardening patches
each closing the previous incident's specific shape — #369 shipped it with a
push; #578 added protected-branch routing; #743 removed the push after it
force-pushed a peer's branch; then an ownership marker after it committed
another agent's worktree under the wrong identity. Each fix was correct and
each was followed by a new leak, because the underlying operation — a broad
`git add -A` over a tree nobody reviewed — is unsafe in a way branch
selection does not address. Also, by its own docstring the rescue mattered
"mainly for CRASHES", which is precisely the case a pre-stop hook cannot
cover: it only runs on a graceful stop.

## What the stop path does now with a dirty worktree

Nothing — the tree is left exactly as dirty as it was found. Verified that
nothing downstream assumed a clean tree:

- `_state/worktree_safety.is_safe_to_reap` (the janitor's prune gate)
  independently returns `False` for any worktree whose `git status
  --porcelain` is non-empty. The lead-learnings/19 prune-destruction window
  stays closed without the rescue; the rescue only ever changed whether the
  work was already committed, never whether the janitor could destroy it.
- Uncommitted work still survives a restart on its own: `workdir` is a host
  bind mount.
- The remaining `_stop.py` sequence (handover snapshot → `pre_stop` hooks →
  `runtime.stop` → `post_stop` hooks → registry teardown) makes no
  clean-tree assumption.

## Changes

- **Archived** to `.old/20260718T232645Z/` (constitution §3, not deleted):
  `_lifecycle/_pre_stop_rescue.py`, `_lifecycle/_pre_stop_rescue_git.py`,
  `tests/…/test__pre_stop_rescue.py`.
- **Call site removed**: `_lifecycle/_stop.py` — the `from ._pre_stop_rescue
  import run_pre_stop_rescue` + `run_pre_stop_rescue(config)` block, replaced
  by a comment recording why it must not come back.
- **Docstrings corrected** where they described the now-absent mechanism:
  `_lifecycle/_orphan_mcp_cleanup.py`, `_state/worktree_safety.py`,
  `_baseline_assets/claude_worktree_hooks/worktree_create.py`.

## Tests — mutation-proved, RED confirmed first

New `tests/scitex_agent_container/_lifecycle/test__stop_no_rescue.py`, driven
against a REAL `tmp_path` git repo on a topic branch (the case the old code
got wrong — a test on `develop` would have passed against the buggy code and
proved nothing).

RED against the pre-removal code:

```
>       assert _git_out(["rev-parse", "HEAD"], cwd=workdir) == head_before
E       AssertionError: assert 'f579e2a341ac...022b0e92ac7a8' == '6d58b6710bb5...bc7ef56499597'
E         - 6d58b6710bb5dc9a04f583e6cc2bc7ef56499597
E         + f579e2a341acedbaf585ef49ccd022b0e92ac7a8
```

HEAD moved — the rescue committing on a topic branch, reproduced.

Guards: HEAD unmoved across the stop; the change is still there and still
dirty (HEAD could also stay put if the work were discarded); no `rescue/`
branch created. Controls: the runtime is still torn down, and `agent_stop`
still returns `True` — without these, "no commit" could pass by breaking
stop entirely.

## Open questions for the operator — deliberately NOT actioned here

1. **The rescue commits already in `develop` are untouched.** They are
   ancestors of published branches; rewriting history is destructive and is
   your call, not ours. The `.tmp-audit/` gitlinks that actually broke CI
   were already removed by PR #769, so nothing is currently broken — what
   remains is cosmetic history.

2. **Seven other repos carry the same pollution.** Not touched — not ours to
   edit. Measured two ways, since commits and polluted paths are different
   units:

   | repo | rescue commits (all refs) | reachable from `origin/develop` |
   |---|---|---|
   | scitex-python | 6 | 0 |
   | scitex-code | 6 | 0 |
   | scitex-io | 4 | 1 |
   | scitex-dataset | 7 | 1 |
   | scitex-ui | 7 | 4 |
   | scitex-audio | 2 | 0 |
   | scitex-cv | 4 | 0 |

   Only 6 of 36 actually reached a shared branch; the rest sit on side
   branches and are harmless. The gitlink paths are all under `.worktrees/`
   (e.g. scitex-python: 8 paths, scitex-ui: 3 paths — matching the
   path-counts in the brief).

3. **The `<git-dir>/sac-owner` stamp now has no consumer.** Its only reader
   was the rescue's ownership gate. Retained rather than removed: the write
   is one cheap out-of-tree file, and the `WorktreeCreate` hook is a baked
   baseline asset, so changing it means re-baking every agent home — wider
   blast radius than this PR should carry. Its docstring now says plainly
   that no ownership gate is enforced anywhere, so nobody infers one. Say the
   word if you want it pulled too.

Nothing under `.worktrees/` or `.tmp-audit/` was deleted on disk — other
agents may be standing on those.


ywatanabe1989:
Superseded by #777. This PR was branched from pre-#772 develop; #772 landed the call-site removal under operator escalation, so #777 is rebased onto current develop and carries only the remaining work (archiving the orphaned modules, the no-second-caller audit, doc corrections, and the regression guard).

PR #780, opened 2026-07-19
fix(env): set the canonical board identity, and reject an unexpanded substitution

ywatanabe1989:
## The bug, in one line

A value meaning *"I could not resolve this"* was accepted and written to the board as if it were an answer.

## What happened

scitex-cards renamed its identity variable `SCITEX_TODO_AGENT_ID` → `SCITEX_CARDS_AGENT_ID`. sac injected only the **old** name. The shared baseline `.mcp.json` had already migrated to the **new** name, so it expanded that reference against an environment where the variable did not exist — and the literal placeholder text survived all the way into the card store.

Cards in the live DB record the variable *name*, unexpanded, as their author. **7 rows when first reported, 15 a few hours later** — the count climbed because every new card written by an affected agent added one. The board cannot say who wrote them.

Nothing errored. That is the whole problem.

### 1. The missing name — `apply_board_identity_alias()`

Mirrors the identity onto **both** spellings with the same value.

Deliberately a mirror and not a swap: installed scitex-cards versions differ across the fleet, so an agent still on a pre-rename build reads only the legacy name. Dropping it would break that agent's writes *in exactly the way the missing new name broke these*.

The effective identity is resolved the way apptainer itself resolves it — `raw_args` are appended **after** the flags rendered from `spec.env`, and `--env` is last-wins — so the mirror cannot disagree with what the container actually receives. For most of the fleet the identity is declared only in `raw_args`, so an alias derived from `spec.env` alone would mirror a value the agent never runs with.

The module docstring states the two conditions that must **both** hold before the legacy name is removed, so the next person does not delete it early and recreate this incident with the names swapped.

### 2. The non-answer stored as an answer — `reject_unexpanded_env()`

The deeper bug is not the rename. Any value that still looks like an unexpanded shell substitution is now rejected **at the boundary where sac sets it**, naming three things the reader needs in order to act: the variable, the offending value, and the likeliest cause.

It is never silently replaced by a default. A silent default would put a *wrong* author on the card — worse than a launch that stops and says why.

Note the deliberate asymmetry with `_to_home_text.interpolate_env`, which **leaves** such literals in materialised *files* for expansion later inside the container. That is correct there — the file is expanded downstream by the agent's own environment. It is exactly what must never happen here: this code renders the container's `--env` flags, which are the end of the line. Nothing downstream expands them, so a substitution reaching this point has already failed.

## Verification

- New unit tests cover both functions, including the split (`--env K=V`) and glued (`--env=K=V`) spellings that both occur in real specs, and the raw_args-wins-over-spec.env precedence.
- `test__apptainer_jail.py` fails **4 of 25 identically on this branch and on develop** — those failures are pre-existing and unrelated. Worth its own issue: develop is currently red.

## Scope note

This fixes the *injection* side. The already-corrupted rows are scitex-cards' data and are left for their migration to normalise — raised with that agent rather than repaired unilaterally.

### DIFF
--- a/src/scitex_agent_container/runtimes/_board_identity_env.py
+++ b/src/scitex_agent_container/runtimes/_board_identity_env.py
@@ -180,21 +180,40 @@ def apply_board_identity_alias(
     *,
     raw_args: Iterable[Any] | None = None,
 ) -> dict[str, str]:
-    """Mirror the board identity onto BOTH env names. Returns a NEW dict.
+    """Ensure the CANONICAL board identity is set. Returns a NEW dict.
 
-    The EFFECTIVE identity is resolved the way apptainer resolves it, so the
-    mirror can never disagree with what the container actually receives:
+    ASYMMETRIC BY DESIGN — reads both names, writes only the new one.
+
+    The identity is READ from either spelling, because specs are mid-migration
+    and many still declare only the legacy name. It is WRITTEN only as
+    :data:`BOARD_ID_ENV`. sac must never re-introduce the legacy name: the
+    operator is removing ``SCITEX_TODO_*`` from the specs entirely
+    (2026-07-19), and an injector that helpfully mirrored it back would
+    silently undo that migration one launch at a time — the deleted variable
+    would keep reappearing and nobody would understand why.
+
+    So this function is also the MIGRATION BRIDGE: a spec that still declares
+    only the legacy name still gets a working canonical identity, which is
+    what lets the legacy name be deleted from specs incrementally instead of
+    in one flag-day sweep.
+
+    Verified before narrowing (2026-07-19): scitex-cards 0.17.0 resolves from
+    :data:`BOARD_ID_ENV` alone and emits no deprecation warning, so dropping
+    the legacy write costs nothing against the installed fleet.
+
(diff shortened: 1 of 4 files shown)


ywatanabe1989:
Direction corrected mid-PR, per the operator (2026-07-19): `SCITEX_TODO_*` is being removed from the specs entirely (「SCITEX_TODO_ 系は全て削除してくれ」/「新名だけでいきましょう」).

The first commit mirrored the identity onto **both** names. That was wrong for this direction — an injector that helpfully writes the legacy name back would undo the migration one launch at a time, and the deleted variable would keep reappearing with no visible cause.

Second commit narrows it: the function now **reads either spelling and writes only the canonical one**. That asymmetry is deliberate and is what makes it a migration bridge — a spec still declaring only the legacy name keeps a working identity, so the old name can be deleted from specs incrementally rather than in one flag-day sweep.

Verified before narrowing, with `env -i` carrying only the new name: scitex-cards 0.17.0 resolves from `SCITEX_CARDS_AGENT_ID` alone and emits no deprecation warning. So dropping the legacy write costs nothing against the installed fleet.

62 passed. The 4 `test__apptainer_jail.py` failures are pre-existing on develop (tracked separately as a P1).

PR #785, opened 2026-07-19
fix(containers): the SIF gates assert a shim whose removal would break every bake

ywatanabe1989:
## The problem

All four container symbol gates imported the board package's **transition shim** and asserted it aliased the real module:

```python
import scitex_cards
import scitex_todo  # the transition shim — its absence breaks the fleet
if scitex_todo is not scitex_cards:
    print("FATAL: ... two module trees baked"); sys.exit(1)
```

`scitex_todo` is an alias the `scitex-cards` wheel ships for the 2026-07-16 rename. Its lifetime is **scitex-cards' call, not ours** — and the day they drop it, *every SIF bake dies at the gate*, broken by a module the image does not need.

The comment asserting "its absence breaks the fleet" is what made this look load-bearing. It isn't, and most of this diff is the argument for why.

## Why removing it does not weaken the gate

Each gate keeps **both** assertions that actually do work:

| assertion | catches |
|---|---|
| `from scitex_cards._throughput import WIP_STATUSES` | a **stale bake** — ImportError, unhandled, non-zero exit, dead build |
| `"in_progress" not in WIP_STATUSES` | a **regressed** WIP gate, which imports perfectly fine |

Dropping the second is how the 2026-07-12 fleet-drift incident shipped: a gate that can only pass. Both are now pinned by tests so a future edit cannot quietly reduce the gate to a no-op.

## Scope — and what this is *not*

This covers the package the **image installs** and the server **agents talk to**. sac's own Python imports of the card package are being deleted wholesale in the separate decoupling PR and are **untouched here**:

```
$ git diff --name-only <base> | rg '^src/scitex_agent_container/' | rg -v '^src/scitex_agent_container/containers/'
(empty)
```

Zero overlap — this can land before or after the decoupling, in either order.

### Three sites checked and deliberately left alone

Each is correct as written; changing it breaks the fleet.

1. **`_mcp/_healthcheck.py` SERVER_ALIASES / `_mcp_reliability.py` CRITICAL_MCP_SERVERS** already *report* under `scitex-cards` and merely *tolerate* the old `.mcp.json` key. sac does not emit that file and the fleet rolls one agent at a time, so a hard flip would classify a not-yet-migrated agent's **healthy** board MCP as absent — and absent means UNKNOWN, which alarms. A false alarm manufactured by the rename itself.

2. **`BASE_REQUIRED_SKILLS = ("scitex-todo",)`** is a *skill id*, and scitex-cards 0.17.0 still ships it as `scitex_cards/_skills/scitex-todo/` (verified on disk). Renaming points every agent at a skill that does not exist. **Blocked upstream, not by us** — it unblocks when scitex-cards renames the directory.

3. **`SCITEX_TODO_*` env vars** — `scitex_cards._store.ENV_AGENT` is literally `"SCITEX_TODO_AGENT_ID"` in 0.17.x. Not ours to rename.

## Verification

**Run, not read:**

- **36 passed** — the new guard + `test_sif_symbol_probe.py` + the renamed def-contract suite.
- **Mutation-proven three ways**, each reverting to green (17 passed):

  | mutation | result |
  |---|---|
  | M1 — reintroduce `import scitex_todo` into `apptainer-base.def` (*the exact revert*) | RED on `test_gate_does_not_import_the_legacy_shim` |
  | M2 — delete the `WIP_STATUSES` import from the probe | RED on `test_gate_still_imports_the_wip_statuses_symbol` |
  | M3 — replace the `in_progress` check with `if False:` | RED on `test_gate_still_checks_in_progress_membership` |

  M2 and M3 also tripped the bake-script lockstep test — intended: `spartan-sif-bake.sh` embeds the probe verbatim.

- Audit: `PS-140` (which I introduced mid-work and then reverted) is **absent**. The 5 remaining `audit-all` errors are pre-existing on the base branch — `PS-204` traces to base commit `e70ce900`; `§4b`, `§6a`, `§10w`, `§13` are repo-wide and untouched here.

**Read, not run:** the SIF bake itself. I did not build an image — the gates are asserted statically, so the claim "the bake still fails on a stale image" rests on the unchanged `WIP_STATUSES` import, not on an observed build.

## Also

`test_apptainer_base_def_scitex_todo.py` → `..._scitex_cards.py`. Its 13 assertions already targeted `scitex-cards`; only the filename, four identifiers and the prose moved — **no assertion changed**. `apptainer-base.def` carried a comment citing one of those test names, which the rename would have silently falsified; the guard surfaced it and it is corrected in the same pass.


ywatanabe1989:
**Collision note before anyone rebases this — its floor hunks are SUPERSEDED, and re-applying them would be a silent regression.**

This PR raises `scitex-cards[mcp]` `>=0.16.0` -> `>=0.17.2` in `apptainer-base.def`, and replaces the two git-URL pins with `>=0.17.2` in `apptainer-scitex.def`.

Since this PR was opened (2026-07-20), two things happened:

1. **`develop` already carries `>=0.17.2`** at all three sites — the raise reached it by another route, so those hunks are now no-ops against current `develop`. This PR reads `CONFLICTING` for that reason.
2. **PR #827 raises those same three sites to `>=0.17.9`.** Reason: scitex-cards published 0.17.8 to PyPI and then installed further code onto the host under that same version, so an agent reading `0.17.8` in its container got the OLD inbox renderer — the one that collapsed a notification to a bare count and let an operator DM sit unread for two hours. 0.17.9 is the first version where the number and the code agree.

**So if this branch is rebased and merged as-is, it will set those three lines back to `>=0.17.2` and silently undo #827.** The floor would still *resolve* to the newest release today, so nothing would break loudly — it would simply remove the guard that stops a pre-publish bake from quietly taking a stale version, which is the exact failure it was added to prevent.

**Recommendation for whoever picks this up:** keep everything else here (the SIF symbol gates, `sif_symbol_probe.py`, the container-gate tests, the lint and docs changes — none of which are in `develop`), and DROP the three `scitex-cards[mcp]` version hunks during the rebase. They are already applied upstream at a higher floor.

Found by checking open PRs against in-flight work rather than branching from `develop` and assuming it was current — an open PR is invisible from the branch you build on. Credit to scitex-cards for naming that shape today.

ywatanabe1989:

**What this PR attempted:** fix(containers): the SIF gates assert a shim whose removal would break every bake

**Why it is being closed:** conflicts with its base; 22 days old, so CI never returned a verdict on it in its current form.

**Nothing is deleted.** Branch `refactor/scitex-cards-name-only` still holds every commit, and this PR can be reopened at any time. If the change is still wanted, a fresh PR against today's default branch is the faster path.

PR #788, opened 2026-07-20
fix(jobs): a JobSpec must not declare a bare `sac` — and ship the detector

ywatanabe1989:
## Part 1 — the fix

`sac.accounts-refresh` declared `command="sac accounts refresh …"`. A bare head.

**What the live host actually showed** (measured, not assumed):

```
unit file : ExecStart=/usr/bin/env sac accounts refresh --all --include-active --sync-active-login
override  : ExecStart=/home/ywatanabe/.env-3.11/bin/sac accounts refresh …   (override.conf, 2026-07-10)
resolved  : /home/ywatanabe/.env-3.11/bin/sac accounts refresh …
```

The brief said the renderer resolves a bare head "through PATH". It is slightly
worse than that. `resolve_execstart` *tries* to absolutise, in order: a `venv`
pin, then **`sys.executable`'s sibling bin**, then PATH, then `/usr/bin/env` as
a rule-3 last resort with a loud warning. So `/usr/bin/env sac` in the unit file
is not the normal path — it is the resolution having **missed entirely**, and a
systemd `--user` unit's minimal PATH fails that at `status=127`. The job has
worked since only because the hand-added `override.conf` pins the absolute path.

That is the defect the brief identified, and it holds: the safety lives in host
state no PR can see. Delete the drop-in or regenerate the unit and the hazard
silently returns.

It also means the hazard has a second axis. Because rule 1 resolves against
whichever interpreter ran `ecosystem up`, the version baked into a unit is
decided by invocation form — and this host carries **seven `sac` installs at
five versions**:

| path | version | kind |
|---|---|---|
| `/home/ywatanabe/.env-3.11/bin/sac` | 0.22.1 | wheel |
| `proj/scitex-cards/.venv/bin/sac` | 0.22.0 | editable |
| `proj/scitex-dev/.venv/bin/sac` | 0.22.0 | editable |
| `proj/scitex-agent-container/.venv/bin/sac` | 0.22.0 | editable |
| `proj/scitex-python/.venv/bin/sac` | 0.21.2 | editable |
| `proj/scitex-mnist-demo/.venv/bin/sac` | 0.21.11 | editable |
| `~/.local/bin/sac` | 0.21.11 | editable, `#!/usr/bin/python3.10` |

### The measured path, and why it is the right one

`/home/ywatanabe/.env-3.11/bin/sac`, because:

1. it is what the live `override.conf` already pins — so this PR makes the
   source agree with reality rather than changing behaviour;
2. it is what **every other installed `sac.*` unit** resolved to;
3. it is the **only non-editable (wheel) install** — the editable ones execute
   whatever is in a working checkout at the time the timer fires;
4. its shebang is the 3.11 venv the rest of the fleet runs on.

Only the string we declare changed. Nothing is asked of the JobSpec/rendering
contract. This is the same rule `sac.heal-agent-auth` already followed:
`resolve_execstart` passes an absolute head through **verbatim**, so the
declaration depends on neither the ambient PATH nor the generating interpreter.

**`sac listen` is untouched** — still deliberately not federated.

## Part 2 — the detector

`sac dev audit-execstart`, backed by `_execstart_audit`. It asks systemd what
each sac unit actually runs and compares it to what the generator intends —
obtained by **calling `resolve_execstart`**, never by re-implementing it (a
checker that states its own opinion of the generator drifts silently, which is
the disease `_jobs_audit` exists to detect).

### Contract

| verdict | meaning |
|---|---|
| `MATCH` | the unit runs exactly what the JobSpec declares |
| `DIVERGED` | it does not — an unmanaged override, or a generator bug. Carries **both** sides (construction fails otherwise: a one-sided divergence is unactionable) |
| `NOT_INSTALLED` | declared, nothing deployed. **Not** a divergence — several sac jobs sit behind a deliberate deploy gate, and flagging those would train the reader to ignore the report |
| `UNVERIFIABLE` | the declared head is not absolute, so the intent is **not reproducible** in the checking interpreter. The check refuses to compare rather than emit a verdict that proves nothing — and names the absolute-head fix |
| `UNKNOWN` | systemd could not be asked at all |

**The UNKNOWN case.** `systemctl` missing (every container), a non-zero exit, no
user bus, a timeout, or an rc=0 shape we cannot parse — each becomes an explicit
UNKNOWN carrying the reason. UNKNOWN is never folded into MATCH, and every
"could not tell" branch is evaluated **before** the comparison so a missing
input cannot fall through into a pass. UNKNOWN does *not* set a failing exit
code — "could not ask" is not "found a problem", and making it red would mute
the check everywhere systemd is absent — but it is always rendered loudly, so
`ok=True` never silently means "nothing was checked".

Two design notes worth flagging:

- **`LoadState`, not the exit code.** `systemctl --user show` exits **0 for a
  unit that does not exist**, printing only `LoadState=not-found` with empty
  stdout and empty stderr. The exit code cannot discriminate; `LoadState` can.
- **Only `argv[]` is compared.** The `ExecStart` record wraps it in `path`,
  `pid`, `status` and timestamps, all of which differ between two identical
  runs. Comparing the record would report a divergence on every restart.

**No `2>/dev/null`.** stderr is captured and carried, including on rc=0. A
file-only test asserts the pattern (and `DEVNULL`) stays out of the detector's
own code; it parses with `ast` and strips docstrings, so it flags *code* rather
than the prose that discusses the hazard. Another file-only test asserts the
package has no write path — it **reports, never repairs**, because the
`override.conf` above is a live example of a hand-patch that was *correct*, and
an auto-repair would have reverted it.

It is a CLI verb rather than a CI test on purpose: CI has no access to the fleet
host's `systemctl --user`, so a checker wired there could only ever report
"could not ask". `_jobs_audit`'s docstring already scopes this axis out for
exactly that reason; this is its sibling, not a replacement.

## Mutation proof

**Live, against real systemd on the fleet host** (temporarily appending
`--MUTATION-PROBE` to the declared command — my source only, never host state):

```
RED   ok = False
      1 UNIT(S) DO NOT RUN WHAT THE SOURCE DECLARES
        sac.accounts-refresh  (sac.accounts-refresh.service)
            intended: …/bin/sac accounts refresh --all --include-active --sync-active-login --MUTATION-PROBE
            resolved: …/bin/sac accounts refresh --all --include-active --sync-active-login
GREEN ok = True     accounts-refresh verdict: ['match']   (probe removed)
```

**Suite mutations** (proving the tests are not vacuous):

- `commands_equal` forced to `return True` (detector blinded) → **4 tests red**,
  incl. both end-to-end divergence tests. Restored → green.
- the no-`systemctl` branch made to return a silent pass instead of the UNKNOWN
  carrier → **1 test red**
  (`test_query_unit_without_systemctl_is_unknown_never_a_pass`). The end-to-end
  UNKNOWN test stayed green because a second guard also catches it — defence in
  depth, not a gap. Restored → green.

No mutation residue remains (`rg MUTATION-PROBE src/` is empty).

Both halves are pinned permanently in the suite, so the proof is a test rather
than a one-off I performed by hand.

## Fixtures are real

No mocks, no monkeypatch. Where systemd is needed the suite writes a **real
executable `systemctl` script** into a real `tmp_path` and runs it through the
real `subprocess.run`, injected via the `which`/`runner` fake-callable seams
(the same convention scitex-dev's own `resolve_execstart` documents). The canned
outputs are **verbatim captures from the fleet host** (systemd 249), not
invented shapes.

One fixture flaw worth recording, because it was a real near-miss: the first
version answered with the same record for *every* unit, which made
`sac.heal-agent-auth` look divergent. Rather than weaken the assertion I made
the fake systemctl **unit-aware** — it now dispatches per unit like the real one.

## Live result on the fleet host

```
match           sac.accounts-refresh
unverifiable    sac.host-sync-check
not-installed   sac.worktree-gc
unverifiable    sac.spartan-sif-bake
not-installed   sac.freshness-refresh
unverifiable    sac.fleet-reconcile
not-installed   sac.restart-login-expired-agents
not-installed   sac.heal-agent-auth
```

`accounts-refresh` is now `MATCH`: the source states what the unit runs, and the
`override.conf` becomes redundant rather than load-bearing. **I have not touched
it** — report, don't repair. It can be removed once this ships, but that is host
state and an operator call.

## What I did not do, and could not verify

- **Five other jobs still declare a bare `sac`** (`host-sync-check`,
  `worktree-gc`, `spartan-sif-bake`, `freshness-refresh`, `fleet-reconcile`).
  The detector surfaces three of them as `UNVERIFIABLE` above — I deliberately
  left them: their units currently carry the right absolute path with no
  override masking anything, so they are latent rather than live, and
  hardcoding an operator-specific path into five more shipped declarations
  would trade today's *warned* failure for a silent `203/EXEC` on any other
  host. Worth a follow-up decision, not a silent widening of this PR.
- **`resolve_execstart` is not publicly exported** from `scitex_dev.jobs` as of
  scitex-dev 0.31.1 — only defined in `scitex_dev.jobs._systemd`. I try the
  public path first and fall back to the defining module, contained to one
  function, with both failures degrading to UNKNOWN. A public export upstream is
  the clean fix; that is scitex-dev's call, not this PR's.
- **The full local test suite did not finish in-session** (>11 min; another
  agent was running pytest on the same host). The full surface touched by this
  change is green: **108 passed** across `test__execstart_audit`,
  `test__jobs_plugin`, `test__jobs_audit`, `test__dev_jobs`. CI runs the rest.
- **`26_credentials-rotation-host.md:109`** describes the timer as firing
  `sac accounts refresh …`. Still true of the CLI verb, now imprecise about the
  declared form. Left alone for scope; flagging it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,47 @@ versioning follows [SemVer](https://semver.org/).
 
 ### Fixed
 
+- **A JobSpec declared a bare `sac`, and only an unmanaged drop-in made it
+  safe.** `sac.accounts-refresh` declared `command="sac accounts refresh …"`.
+  The generated unit read `ExecStart=/usr/bin/env sac accounts refresh …` —
+  `resolve_execstart`'s rule-3 LAST RESORT, reached only when the interpreter's
+  sibling bin and the ambient PATH both miss — which a systemd `--user` unit's
+  minimal PATH fails at `status=127`. It has worked since 2026-07-10 only
+  because a hand-added `override.conf` pinned the absolute path.
+
+  That is the defect: the safety lived in host state no PR can see, so deleting
+  the drop-in or regenerating the unit silently restored the hazard. The command
+  now declares `/home/ywatanabe/.env-3.11/bin/sac` outright, the same rule
+  `sac.heal-agent-auth` already followed — `resolve_execstart` passes an
+  absolute head through verbatim, so the declaration depends on neither the
+  ambient PATH nor on which interpreter ran `ecosystem up`. That second
+  dependency is not hypothetical: this host carries SEVEN `sac` installs at FIVE
+  versions, and rule 1 resolves against `sys.executable`'s sibling bin.
+
+### Added
+
+- **`sac dev audit-execstart` — the detector for deployed-vs-declared drift.**
+  Both instances of the `ExecStart` hazard were found by hand, comparing
+  `systemctl --user show -p ExecStart --value <unit>` against the JobSpec
+  source. Nothing ran that comparison, so both were found by luck; a hand repair
+  restores the state and leaves the system able to re-enter it.
+
+  The new `_execstart_audit` package asks systemd what each sac unit actually
+  runs and compares it to what the generator intends (obtained by calling
+  `resolve_execstart`, never by re-implementing it). A divergence means an
+  unmanaged local override or a generator bug — both worth knowing. It REPORTS
+  and never repairs: the `override.conf` above is a live example of a hand-patch
+  that was correct, and an auto-repair would have reverted it.
+
+  Five verdicts, because the failure modes are not binary: `MATCH`, `DIVERGED`,
+  `NOT_INSTALLED` (declared behind a deliberate deploy gate — not a divergence),
+  `UNVERIFIABLE` (the declared head is not absolute, so the intent is not
+  reproducible in the checking interpreter — the check refuses to compare rather
+  than emit a meaningless verdict), and `UNKNOWN` (systemd could not be asked at
+  all, e.g. inside a container). `UNKNOWN` is never folded into `MATCH`: a check
+  that cannot distinguish "matches" from "could not ask" is not a check. stderr
+  is captured and reported, never discarded.
+
 - **The version lie, caught by a test instead of by an incident.** `pyproject.toml`
   is bumped to `0.23.0`, and a guard now fails CI whenever the CHANGELOG lists
   pending work under `[Unreleased]` while the declared version is one the
--- a/src/scitex_agent_container/_execstart_audit/__init__.py
+++ b/src/scitex_agent_container/_execstart_audit/__init__.py
@@ -1,0 +1,195 @@
+"""Deployed-vs-declared detector: does the unit RUN what the source INTENDS?
+
+The axis :mod:`scitex_agent_container._jobs_audit` deliberately leaves
+out. That module answers "is this declaration reachable at all?" from the
+repo tree; this one answers the question underneath the two ``ExecStart``
+incidents we found BY HAND, both by running the same comparison in a
+terminal and eyeballing it::
+
+    systemctl --user show -p ExecStart --value <unit>   # what ACTUALLY runs
+    vs. what the JobSpec source intends
+
+Nothing ran that comparison on a schedule, so both instances were found
+by luck. A hand repair restores the STATE and leaves the SYSTEM able to
+re-enter it; for anything recurring, the deliverable is the DETECTOR.
+
+The two measured instances
+==========================
+1. ``sac.accounts-refresh`` — the generated unit read
+   ``ExecStart=/usr/bin/env sac accounts refresh ...``, which is
+   ``resolve_execstart``'s rule-3 LAST RESORT (interpreter sibling-bin
+   AND ambient PATH both missed). A systemd ``--user`` unit gets a
+   minimal PATH, so that form fails at ``status=127``. It kept working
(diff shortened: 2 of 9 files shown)


ywatanabe1989:

**What this PR attempted:** fix(jobs): a JobSpec must not declare a bare `sac` — and ship the detector

**Why it is being closed:** conflicts with its base; 21 days old, so CI never returned a verdict on it in its current form.

**Nothing is deleted.** Branch `fix/jobspec-must-not-declare-a-bare-sac` still holds every commit, and this PR can be reopened at any time. If the change is still wanted, a fresh PR against today's default branch is the faster path.

PR #806, opened 2026-07-21
feat(config)!: every spec field explicit — missing field is a load error (red-start)

ywatanabe1989:
## Summary

Implements the operator ruling of 2026-07-21, verbatim intent, not softened:

1. **EVERY field in an agent `spec.yaml` must be written explicitly.** An omitted field is an ERROR at config-load time, with a hint. ("スペックには全部明示的に書いてほしい / 書いて無ければエラーにしてヒントを出してください")
2. **NO migration machinery, NO warn phase, NO escape-hatch env flag.** ("一気にブート不能で大丈夫です。red スタートしましょう。ハードに。移行とかばかばかしい")
3. dataclasses + a validator. 4. The structure leaves no option but compliance ("『揃うしかなくなる』").

**THIS PR INTENTIONALLY BREAKS EVERY UNDER-SPECIFIED SPEC AT ITS NEXT BOOT.** That is the ruling, not a side effect. The error each red spec prints contains a paste-ready YAML block that makes it green again with byte-identical behaviour.

## What lands

- `config/_explicit_fields.py` — the required-key map: **86 required keys for `kind: Agent`, 78 for `kind: AgentProxy`**, across 14 sections. Derived from `dataclasses.fields()` of the section dataclasses wherever YAML key == field name; explicit alias tables where not (`watchdog.responses.{y_n,y_y_n,waiting}`, `restart.backoff.{initial,max,multiplier}`, `python-venv`).
- `config/_explicit_validation.py` — the walker. Collects ALL missing fields in ONE pass; raises a single `ExplicitSpecError(ValueError)` whose message lists, per missing field, its YAML path + expected type + current default, then a marker-delimited **paste-ready YAML block** (merge it in, become green — round-trip TESTED), ending with the file being loaded. The only signature is `validate(doc, path)` — no `strict=`, no env flag, no bypass.
- Wired **twice from one SSOT**: `load_v3` raises before any parsing (a red spec fails with the complete hint, never a parser TypeError), and `validate_raw` appends the same consolidated message so `sac agents check` reports identically to `load_config`.
- The 2026-06-23 nine-field "no hidden defaults" subset in `_validation.py` was superseded and removed.
- Presence semantics: a key present with `null` IS explicit (matches the prior checker). Loader derivations stay: `workdir: null` = per-agent runtime dir; `claude.session: null` = role-derived session. The paste block uses exactly those spellings so pasting cannot change behaviour (a pasted `session: fresh` would have silently flipped coordinators off `continue`; `null` cannot).

## Exclusion list (each documented in `_explicit_fields.py`)

| Excluded | Reason |
|---|---|
| `spec.host` / `spec.hosts` | mutually-exclusive pair; exactly-one presence already enforced loudly by `_placement_validation` — a require-all map cannot demand both |
| `spec.session` (top-level) | ergonomic alias of the canonical `spec.claude.session`; requiring both = forced double declaration |
| `spec.claude.*` on AgentProxy / `spec.proxy.*` on Agent | the kind-coupling validator FORBIDS the cross block |
| `spec.screen` | legacy inert metadata (no longer drives a multiplexer) |
| `spec.startup` | in `_KNOWN_SPEC_KEYS` but materialised by NO parser (dead key) |
| `multiplexer`, `env-file`, `exclude_hooks`, `exclude_skills` | **read by `load_v3` but ABSENT from `_KNOWN_SPEC_KEYS`** — `validate_raw` rejects them as unknown, so requiring them is red-both-ways. **Flagged for operator review** (see below) |
| banned/relocated keys (`access`, `scheduling`, `dockerfile`, top-level `image/env/model/mounts`, `skills`, `dot_claude`, `remote`, `metadata.name`, `apptainer.container_workdir`) | banned stays banned, never required |
| list-item internals (`listen[].port`, `startup_commands[].command`, bind entries, `mcp_servers.*`, `claude.provider` dict internals) | the list/mapping itself is the explicit declaration; item shapes are parser-validated |

## Deserves operator review

- **`multiplexer` / `env-file` / `exclude_hooks` / `exclude_skills`**: `load_v3` reads them but `validate_raw` rejects them as unknown spec keys — they are unreachable through any validated load today. Either add them to `_KNOWN_SPEC_KEYS` + the required map, or delete the dead reads.
- **`health.method` fossil default**: the dataclass default `multiplexer-alive` is REJECTED by the validator when written (`sdk-alive` is the only accepted value). The paste block emits `sdk-alive`; the dataclass default should probably be updated to match.
- **`claude.model` empty-string default**: pasting `model: ''` is valid and means "runtime default (sonnet)" — kept for behaviour preservation.

## Mutation-proof test evidence

`tests/scitex_agent_container/config/test__explicit_validation.py` (25 tests):

- A HAND-WRITTEN fully-explicit spec (all 86 fields typed out, not generated) loads green — the map and the fixture cannot agree by construction.
- Parametrized mutation-proof over **13 distinct fields spanning top-level, claude, apptainer, health, watchdog, a2a, restart, comms, context_management**: removing that one field turns the load RED and the error names that exact YAML path.
- Several-missing case asserts the single error counts and lists ALL of them at once.
- **Round-trip**: the paste-ready block extracted from the error, merged into the failing spec, loads green — and reproduces omission behaviour (`claude.session` heals to the role-derived value, not a hard-coded one).
- `inspect.signature(validate)` is asserted to be exactly `(doc, path)` — no bypass parameter can appear without failing a test.
- Direct `load_v3` call raises `ExplicitSpecError` (wiring is not only via `load_config`).

## Repo-owned specs made explicit (complete list)

- `examples/agents/{codex-agent,deepseek-agent,full-agent,hello-agent,minimal-agent,proxy-agent}/spec.yaml` — all 6, healed comment-preservingly; all load green.
- `sac agents create` inline templates (`minimal` / `full`) — render the complete explicit field set; extracted to `cli_pkg/_create_templates.py` for the 512-line cap (`_create.py` 609 → 333 lines).
- `sac start` cold-start template (`cli_pkg/lifecycle/_cold_start.py::_COLD_START_SPEC`) — sac's own rendered specs now carry the full set.
- Test fixtures across ~45 modules migrated onto `tests/scitex_agent_container/_helpers/explicit_spec.py` (`explicit_spec` / `explicitize_yaml` — deep-merge the validator's own paste defaults BENEATH each fixture, fixture wins), so fixtures track the map by construction while the hand-written green fixture keeps the map honest.

## Out-of-repo spec impact (the merge-safety question)

Every spec that reaches `load_config` / `load_v3` / `validate_raw` is subject to the new gate. My independent read of what boots from specs OUTSIDE this repo:

- **Live fleet specs** (`~/.scitex/agent-container/agents/<name>/spec.yaml`): every under-specified one goes RED at its next `start`/`restart`. That is the ruling's intent, not collateral — and the error each prints is a complete paste-ready fix. Running agents are untouched until their next lifecycle event.
- **Host dir-templates** (`_template_python_developer` / `_template_researcher` / `_template_generalist` + underscore-agent skeleton in the agents root): operator reports these were made explicit earlier today; `sac agents create` copies them verbatim, so an un-healed template would produce red children — worth a `sac agents check` sweep over the agents root after merge (the check command now reports the same consolidated hint).
- **Defensive surfaces stay standing**: the reconciler (`test_scaffold_dirs_are_not_agents` / malformed-spec tolerance), `refresh-acl` (`test_malformed_spec_does_not_abort_the_others`) and the listen daemon's per-agent status handle a failing spec per-agent and continue — a red spec degrades that one agent's row, it does not take the control plane down.
- **Inline specs over the listen `/agents` POST** (a2a spawn / twin): materialised inline specs must now be fully explicit too — an under-specified inline spec is rejected with the same hint (covered by the migrated `_listen` server tests).
- **Pre-existing dead surface, flagged for operator review (NOT newly broken here)**: the MCP `template_render_contributor_spec` template (`_mcp/_tools/_template.py`) still renders top-level `image`/`model`/`multiplexer`, `spec.skills`, `health.method: multiplexer-alive` and an unknown `a2a.handler` — ALL of which the validator already rejected before this PR. Its output could not load before and cannot load now; it needs a rewrite or retirement in a follow-up.

## Verbatim-move audit

- The only code MOVE in this PR is the two `sac agents create` inline template strings, `_create.py` → `_create_templates.py`. The surrounding machinery (`_TEMPLATES` dict, render call) moved verbatim; the template BODIES did not and were not meant to — they were deliberately rewritten to carry the full explicit field set (that rewrite is the point of the PR).
- Nothing touching the provider tool whitelist moved: `rg allowed_tools` over this branch's `src/` diff against develop returns zero hits — `ProviderSpec.allowed_tools` / `_parse_provider` are untouched.

## Known follow-up (from review, deliberately out of this PR)

The reviewer's finding stands: `_agent_list.py` folds a spec load failure into `labels == {}`, so a not-yet-explicit spec vanishes from every FILTERED `sac agents list` view while the red-start makes exactly that failure common. Follow-up: carry a `config_error` third state on the row + report an unreadable-spec count in filtered views.

## Version / changelog

- CHANGELOG `[Unreleased]` → BREAKING entry describing the ruling + hint contract (combined with develop's autobump + PS-221 entries at merge).
- `pyproject.toml` → **0.24.11** (develop moved to 0.24.10 while this PR was in flight; the branch's version stays one ahead of the latest shipped). `tests/integration/test_version_not_spent.py` passes (9/9).
- `origin/develop` is merged up through 0.24.10.

## Pre-existing audit findings (exist on base, not introduced here)

The blocking PS-221 audit error is fixed on develop (#805, merged here). A local `tests/develop/test_audit.py` run earlier also listed pre-existing, non-blocking findings untouched by this PR — §4b CliHelp free-form help, §6a `CODEX_HOME`/`LOG_PATH` env prefixes, §13 `skills` top-level command, §10w import-budget skip (environmental) — all verified present on `origin/develop` blobs.

https://claude.ai/code/session_01ASLR9KjueRGm4BnZnqUhRF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ versioning follows [SemVer](https://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- **BREAKING — every spec field explicit, red-start (operator ruling
+  2026-07-21).** `config.load_config` / `load_v3` / `validate_raw` now REQUIRE
+  every author-facing field of an agent `spec.yaml` to be WRITTEN — an omitted
+  field is a load-time ERROR. 86 required keys for `kind: Agent` (78 for
+  `kind: AgentProxy`), derived from the section dataclasses
+  (`config/_explicit_fields.py`) with alias tables where YAML keys differ
+  (`watchdog.responses.*`, `restart.backoff.*`, `python-venv`). There is NO
+  migration phase, NO warn mode, and NO bypass flag — the only entry point is
+  `_explicit_validation.validate(doc, path)`; every under-specified spec in
+  the fleet goes boot-red at its next start, as ruled. The hint contract:
+  ONE error listing ALL missing fields at once (YAML path + expected type +
+  current default), followed by a marker-delimited paste-ready YAML block
+  whose values reproduce exactly what omission used to mean (loader
+  derivations paste `null`: `workdir`, `claude.session`), ending with the
+  loaded file's path. The hint provably clears the condition (round-trip
+  tested). The 2026-06-23 nine-field "no hidden defaults" subset was
+  superseded and removed. Excluded from the required set (each documented in
+  the module): `host`/`hosts` (exactly-one enforced by placement validation),
+  top-level `session` (alias of `claude.session`), banned/relocated keys
+  (`access`, `scheduling`, `container_workdir`, ...), the dead `screen` /
+  `startup` keys, and the four keys `load_v3` reads but `validate_raw`
+  rejects (`multiplexer`, `env-file`, `exclude_hooks`, `exclude_skills` —
+  flagged for operator review). `sac agents create` templates and
+  `examples/agents/*` now render the full explicit field set.
+
 ## [0.24.2] - 2026-07-21
 
 ### Added
--- a/examples/agents/codex-agent/spec.yaml
+++ b/examples/agents/codex-agent/spec.yaml
@@ -21,6 +21,21 @@ spec:
   apptainer:
     image: ~/.scitex/agent-container/containers/sac-scitex.sif
     binds: []
+    env: {}
+    raw_args: []
+    post: ''
+    environment: {}
+    def_file: ''
+    nv: false
+    rocm: false
+    overlay: ''
+    overlay_size: ''
+    overlay_create_if_missing: true
+    tmpfs_size: 2G
+    relaxed: false
+    fakeroot: false
+    jail: false
+    nested_build: false
 
   # Do not set top-level `spec.provider: openai`; that selects a different
   # harness. The nested provider below keeps Claude Code and changes only its
@@ -29,15 +44,90 @@ spec:
     provider: codex
     model: gpt-5.6-sol
     flags:
-      - --dangerously-skip-permissions
-      - --effort=medium
+    - --dangerously-skip-permissions
+    - --effort=medium
 
+    channels: []
+    raw_options: {}
+    session:
+    continue_max_age_minutes:
+    resume_id: ''
+    auto_accept: true
+    account: ''
+    credentials_file: ''
+    credentials_files: []
   health:
     enabled: true
     interval: 60
 
+    timeout: 5
+    method: sdk-alive
   restart:
     policy: on-failure
     max_retries: 3
 
 # EOF
+    backoff:
+      initial: 30
+      max: 300
+      multiplier: 2
+    prune_on_stop: false
+
+  # Explicit-fields ruling (2026-07-21): every spec field is written; values below are the defaults.
+  provider: anthropic
+  python-venv: ''
+  startup_commands: []
+  startup_prompts: []
+  listen: []
+  extensions: {}
+  mcp_servers: {}
+  user: ''
+  to_home: ./to_home
+  container:
+    runtime: none
+    image: scitex-agent-container:latest
+    volumes: []
+    network: host
+    mount_host_claude: false
+  watchdog:
+    enabled: false
+    interval: 1.5
+    responses:
+      y_n: '1'
+      y_y_n: '2'
+      waiting: /speak-and-call
+  autonomous:
+    enabled: false
+    drive_until: DONE
+    max_turns: 50
+    idle_kick_after_s: 120
+    kick_text: Continue. Print DONE when finished.
+  hooks:
+    pre_start: []
+    post_start: []
+    pre_stop: []
+    post_stop: []
+    on_compact: []
+    on_restart: []
+    on_diff: []
+  context_management:
+    trigger_at_percent: 70.0
+    strategy: noop
+    warn_before_n_checks: 0
+    check_interval_seconds: 300
+    state_file: ~/.scitex/agent-container/state/<agent>.json
+  a2a:
+    host: 127.0.0.1
+    port: auto
+  comms:
+    outbound:
+      siblings: allow
+      parent: allow
+    inbound:
+      siblings: allow
+      parent: allow
+    a2a:
+      listen: true
+  lineage:
+    group: ''
+    may_spawn: true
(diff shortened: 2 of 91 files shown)


ywatanabe1989:
Review finding from the CI failure, recorded here because it outlives this PR.

**What failed:** three label-FILTER tests in `tests/scitex_agent_container/cli_pkg/_helpers/test__agent_list.py` returned `[]` where they expected a match. Root cause was a missed fixture: `_write_valid_spec` still wrote a minimal v3 spec, which the new explicit-field validator rejects. Fixed by routing it through `explicitize_yaml`, the same helper the rest of the migrated suite uses. 68 passed locally.

**The finding worth keeping**, which the fixture fix does NOT address:

At `cli_pkg/_helpers/_agent_list.py:265-281`, `load_config` is wrapped in a catch-all. When it raises, `labels` stays `{}` and the agent is then `continue`d out of every filtered view — `--capability`, `--machine`, `--group`. The agent still appears in an unfiltered listing with empty labels (the inline comment says so deliberately), but it silently disappears from every filtered query.

That catch-all was reasonable when a load failure meant malformed YAML, which is rare. This PR changes the population: **every not-yet-explicit spec now fails to load**, so the rare path becomes the common one. On the live fleet the operator manages agents through filtered views, which means an agent whose spec is not yet explicit would vanish from `sac agents list --group active` with no error shown.

That is a fail-silent, and it sits badly next to the intent of this change. The whole point of the red-start is that a missing field fails LOUDLY at boot. A listing that quietly drops the same agent contradicts it.

Note also why CI could not catch this on its own: the negative-case tests (`..._excludes_non_matching_agent`) assert `out == []`, and a spec that failed to load produces exactly the same `[]` as an agent correctly excluded by the filter. The two states are indistinguishable at the assertion, so the suite stayed green on the half that mattered.

**Suggested follow-up (deliberately not in this PR, to keep the diff bounded):** make the load failure a third state rather than folding it into "no labels" — carry a `config_error` on the row and have filtered views report a count of unreadable specs instead of silently omitting them. I am not blocking the merge on it, but it should not be lost.

PR #813, opened 2026-07-22
chore(runners): remove dead runtime dispatch advertising docker/podman/ssh-remote

ywatanabe1989:
## Why

A peer agent asked whether docker-on-macOS was the sanctioned way to run a sac
agent on a MacBook. Reading `spec.container.runtime: none | docker | apptainer`
in the type, plus a docker branch sitting in the runner, the honest conclusion
is "docker is supported". It is not. They would have written a spec, shipped a
proposal, and hit `ImportError` at execution.

Dead code that reads as a feature costs more than absent code, because it
survives review. This removes the advertisement and makes the enum match what
is actually implemented.

## What could not execute

`_runners/_tmux/claude_code.py` carried **four** dispatch blocks — `start`,
`stop`, `is_running`, `logs` — each with the same two dead halves:

```python
if _should_dispatch_remote(config):
    return SSHRemote.start(config, ...)
if config.container.runtime != "none":
    from .apptainer import ApptainerRuntime
    from .docker import DockerRuntime
    from .podman import PodmanRuntime
```

1. **`.apptainer` / `.docker` / `.podman` do not exist in that package.**
   Full listing of `_runners/_tmux/`: `auth_status.py`, `auto/`,
   `claude_code.py`, `_host_cwd.py`, `__init__.py`, `multiplexer.py`,
   `pane_capture.py`, `prompts.py`, `_tmux_probe.py`, `tmux.py`. Reaching the
   branch raises `ImportError`.
2. **`_should_dispatch_remote` read `config.remote.is_remote`**, but
   `RemoteSpec` was deleted in WI-6 (handoff §6, 2026-05-20) — `AgentConfig`
   has no `remote` field, only a comment recording the removal. Reaching it
   raises `AttributeError`.

## The live start path is untouched

Traced end to end before deleting anything:

`_lifecycle/_start.py:253` → `_runtime_select._get_runtime(config)`, which
branches **only** on the top-level `config.runtime` launch-mode selector
(`""`/`tui` → `TuiSessionRuntime`; `apptainer`/`claude-agent-sdk` →
`ClaudeSessionRuntime`), both landing in `runtimes/_apptainer_*`. It never
reads `config.container.runtime`, and never reaches `ClaudeCodeRuntime`.

Stronger: the whole module is already unimportable, verified empirically rather
than inferred —

```
IMPORT FAILS: ModuleNotFoundError No module named
'scitex_agent_container._runners._tmux.src_files'
```

`claude_code.py:42` imports a `.src_files` that 6fb9131 replaced with
`dot_claude/`. Nothing under `src/` imports `claude_code` at all. That import
is left in place here (its six functions have live call sites in the module
body, so removing it is a real refactor, not a deletion) and is called out
below as follow-up work.

## The enum

`spec.container.runtime` now accepts `none | apptainer`
(`VALID_CONTAINER_RUNTIMES`), and the rejection message is generated from that
set so it cannot drift from it.

docker is **not** implemented anywhere else — checked before narrowing.
`runtimes/claude_session.py:203-205` is explicit:

```python
# 2026-05-13 docker/podman ripout: apptainer is the only accepted runtime.
_CONTAINER_ENGINES: tuple[str, ...] = ("apptainer",)
```

Note the validator was **looser than the docstring**: it accepted
`none|docker|podman|apptainer`, so `podman` validated cleanly despite never
being advertised in the type comment.

## Other surviving RemoteSpec consumer

Swept `src/` and `tests/` for the retired type. One other live-looking
consumer, fixed here: `_skills/scitex-agent-container/03_python-api.md`
documented, as live Python API,

```
cfg.remote.host      # str ("" if local)
cfg.remote.is_remote # bool
```

Both raise `AttributeError` — the same trap one layer up, and worse, since a
skill doc is what an agent reads to decide what to write. Replaced with the
real accessor `cfg.hosts_spec.host` (verified against `_dispatch.py:366`,
`_attach.py:71`, `_host_routing.py:188`).

Every other `spec.remote` hit is benign: comments recording the deletion,
migration tables, or `config/_validation.py:130`, which correctly *rejects*
`spec.remote` and redirects to `spec.host`.

## Test

`tests/scitex_agent_container/config/test__validation_container_runtime.py`

The load-bearing test probes the **live** resolver (`_container_runtime_for`)
for every advertised value, rather than comparing two copies of a list — so it
fails on *implementability*, and adding an unbacked engine cannot be made to
pass by also editing an expected set.

```
/opt/venv-sac/bin/python -m pytest \
  tests/scitex_agent_container/config/test__validation_container_runtime.py -v
```

Mutation proof:

- baseline — `10 passed in 5.49s`
- inject `"docker"` into the enum — `4 failed, 8 passed`:
  ```
  FAILED test_every_advertised_engine_resolves_to_a_real_runtime[docker]
    AssertionError: container.runtime='docker' is accepted by the validator
    but _container_runtime_for resolves no engine for it
    assert None is not None
  ```
- inject a bogus `"bogusengine"` — `2 failed, 10 passed`, same probe fires
- revert — `10 passed`

## Pre-existing reds, untouched

The wider run (`config/`, `_lifecycle/`, `_runners/`, `test_config*.py`) is
`2 failed, 2511 passed, 38 errors`. Those reds are **not from this change** —
established by running the same three files on clean `origin/develop`
(`e9247565`, without this commit): `2 failed, 88 passed, 36 errors`.

They are scitex-cards store-bootstrap errors in `_lifecycle/_rename_*`
(`RuntimeError: canonical store .../board/cards.db does not exist`), plus a
flaky `test_claude_session.py` subprocess heartbeat timeout that passed on
rerun. Nothing was weakened, skipped, or xfailed to get green.

A second reviewer independently ran the same discriminator over
`tests/scitex_agent_container/_runners` and reached the same verdict:

```
on this branch      1 failed, 332 passed in 76.52s
on clean develop    1 failed in 9.36s
FAILED test__session_http.py::TestAgentCard::test_well_known_card_returns_status_200
E   TimeoutError: timed out
    test__session_http.py:275 -> asyncio.to_thread(
        _http_get, "http://127.0.0.1:{port}/.well-known/...")
```

Identical test, identical timeout, present without this change. It is an HTTP
agent-card test timing out binding/reaching a local port — unrelated to runtime
dispatch. The other 332 tests in that directory pass on this branch.

## Follow-up (not in scope here)

- `_runners/_tmux/claude_code.py` is orphaned and unimportable via its
  `.src_files` import; nothing in `src/` imports it. It is either revived or
  deleted wholesale — worth a decision rather than another salvage comment.
- The `cards.db` fixture reds above look like a test-isolation guard change
  landing without bootstrapping a per-test store.


ywatanabe1989:
**Reviewer correction to two claims in the PR body.** The PR itself is unaffected — the deletion is still correctly exonerated — but two statements would mislead a future reader.

### 1. These reds are environment artifacts, not a broken `develop`

The body reports pre-existing reds on clean `origin/develop`, which is accurate *as a comparison*. But it invites the reading that develop is broken. It is not.

CI on `develop` has been green five consecutive times today, in the authoritative environment:

```
07:26:12  workflow_dispatch  success
04:38:06  workflow_dispatch  success
03:22:59  pull_request       success
03:18:46  push               success
03:06:47  push               success
```

Full pytest matrix, py3.11/3.12/3.13, every run.

Both measurements here were taken **inside a container**, and every failure sits on an environment-dependent operation:

- `RuntimeError: canonical store .../board/cards.db does not exist` — the store is absent in this container
- `TimeoutError` binding/reaching `http://127.0.0.1:<port>`
- `test__broker_self.py` — collected 12, ran 2, then **hung**; a stall on port binding is an environment signature, not an assertion failure

**What still holds:** the branch-vs-develop comparison was run in the *same* environment both times, so it remains a valid discriminator and the conclusion — these reds are not caused by this change — stands.

### 2. Retracting the test-isolation-guard hypothesis

The follow-up section proposes the `cards.db` reds come from "a test-isolation guard change landing without bootstrapping a per-test store." Checked, and refuted:

```
git show origin/develop:tests/conftest.py | grep addaudithook   -> no match
branch fix/tests-cannot-write-real-runtime-dir                  -> merged nowhere
```

That guard is still unpushed work in a scratch worktree. It cannot be the cause. The hypothesis came from reading a commit message in a work-in-progress tree and treating it as a statement about `develop`.

### Method note

I reproduced the agent's reds in the *same kind of environment* and treated the agreement as corroboration. It was one instrument sampled twice — the corroborators could not have disagreed. Asking CI, the authoritative environment, was one `gh run list` away and settles it in a step.

The remaining follow-up in the body is still worth doing: `_runners/_tmux/claude_code.py` is orphaned and unimportable, and deserves a revive-or-delete decision rather than another salvage comment.

PR #814, opened 2026-07-22
feat(cli): honest output levels — short lines, no traceback on an expected refusal

ywatanabe1989:
Operator ruling 2026-07-22:

> フィードバックがクソ長い、ユーザ目線に立って短く、読みやすく。これはいつもそう。scitex-logging も使って。

Cards: `sac-cli-output-ssot-logging-levels-20260722` (umbrella),
`sac-error-messages-too-long-no-traceback-20260722`,
`sac-start-noop-reports-started-20260722`.

This is an **output-layer pass only** — no logic was refactored. Levels now
carry meaning, so the reader can sort the screen at a glance:
INFO (dim, skippable) · WARN (pulls the eye) · ERRO (unambiguous failure) ·
SUCC (success visible *as* success).

### 1. An expected refusal no longer prints a stack trace

`NoHealthyAccountError` is deliberately raised — a handled state. It was
escaping to the top as ~5 lines of prose **plus a full traceback repeating the
same paragraph verbatim**. A traceback tells the reader "sac crashed, file a
bug", which is false.

It now carries a one-line `brief` alongside its full diagnosis, and the CLI
boundary catches it explicitly:

```
ERRO: dotfiles: quota unknown for wyusuuke-gmail-com — run `sac accounts refresh-quota-cache`
```

Nothing is lost: the full per-account diagnosis (probe time, every candidate's
state, the evidence path) still prints, at DEBUG. `--json` gains a
`no-healthy-account` status. Genuinely *unexpected* exceptions keep their
traceback — that distinction is the point.

### 2. `sac accounts refresh-quota-cache` reports success as success

Was bare `click.echo` with no level, so success looked like neither success nor
noise. Now `logger.success`. The absolute cache path is gone (implementation
detail nobody asked for); **the per-account 5h/7d/ttl figures stay** — those
are the numbers the operator actually reads to decide which account has
headroom.

```
INFO:   wyusuuke-gmail-com    5h=0% 7d=100% ttl=2.36h
INFO:   ywata1989-gmail-com   5h=20% 7d=57% ttl=2.36h
SUCC: refreshed 3 account(s)
```

### 3. `sac start` on a live agent stops contradicting itself

It printed `No-op.` and then `SUCC: <name> started` — two adjacent lines
disagreeing, and "started" asserting an action that did not occur. The operator
read them and reasonably asked whether the command had failed.

```
INFO: scitex-dev is already running [ALIVE (delivery: 1 live inbox subscriber)] — nothing launched
INFO:   - restart it:          sac agents restart scitex-dev -y
INFO:   - force a fresh start: sac agents start scitex-dev --force
INFO:   - stop, then start:    sac agents stop scitex-dev && sac agents start scitex-dev
```

INFO rather than SUCC: the no-op is only a success if the user's intent was
"ensure it is running". **Exit code stays 0** — the requested end state holds,
matching `systemctl start` / `docker start` and PR #756's existing decision.
`--json` reports `already_running` so a supervisor can tell the two apart.

**Every hinted command was executed before shipping**, against the same
nonexistent target so the difference is attributable to the `-y` gate alone:

```
$ sac agents stop zzz-nonexistent-probe        # no -y
Error (zzz-nonexistent-probe): Agent 'zzz-nonexistent-probe' not found in registry

$ sac agents restart zzz-nonexistent-probe     # no -y
Refusing to restart agent 'zzz-nonexistent-probe' without --yes/-y.
```

`restart` requires `-y`; **`stop` does not** — its `-y` gate covers only the
fleet-wide selection flags (`--all-running` etc.), confirmed in
`cli_pkg/lifecycle/_stop.py`. The asymmetry is real, so the hints differ, and
two tests pin exactly that.

### 4. Credential-ranking detail leaves the default path

The ~8 lines of `policy=spread`, ranking inputs and per-account usage move to
DEBUG. A one-line headline naming the agent and the picked account stays at
INFO.

## Also swept

- `red` now maps to `scitex_logging.ERROR` (renders `ERRO`, the doctrine's
  name) instead of `FAIL`; `fail` remains for a caller reporting a failed
  *check*. 6 call sites; no test asserted the old prefix.
- `_announce_start_verdict`'s two `print(..., file=sys.stderr)` calls route
  through the logger; the UNKNOWN advisory becomes WARN and its prose is cut.
- `ResumePreflightError` renders through the logger instead of `console.print`.

## Out of scope by operator ruling

**The no-bare-print audit rule is not in this PR.** 「サイテクスデベロップの仕事な
ので連携して、デベロッパーの方に進めてもらってください、こちらからは現場の声を上げ
て、後からプリミティブ、抽象化されたルールに沿って書いて直す」 — the abstracted rule
belongs to scitex-dev; sac supplies field evidence and adopts their primitive.
A sac-local version would fork a shared standard. The inventory gathered for
that hand-off is in a PR comment.

## Contradictions found against the written design

1. **There is no honest source for `(since <time>)`.** The card's target reads
   `INFO: scitex-dev is already running (since 14:20:17)`. sac has no
   session-start timestamp reachable at that branch: no `session_created`
   anywhere in the tree, the registry record has no `started_at`, and the
   `instances` table column would need a new DB read on a hot path. Rather than
   fabricate a timestamp I kept the **liveness verdict evidence**, which is
   there by a deliberate prior decision — a no-op that says only "already
   running" is unfalsifiable from outside, so the operator could not tell an
   OBSERVED agent from a process-shaped shadow. If elapsed time is wanted it is
   a separate card (surface `instances.started_at`).

2. **The ranking notice could not simply "go through a logger".** A comment at
   `_start_preflight.py` warned that its `log_stream` seam is load-bearing:
   `preflight_from_config_path` passes a throwaway `StringIO` to *suppress* the
   dry-probe rotation notice. Routing unconditionally to a logger would have
   broken that suppression silently. A caller-supplied stream therefore still
   bypasses the logger; only the real CLI path (`log_stream is None`) goes to
   INFO/DEBUG.

## Test

```
/opt/venv-sac/bin/python3 -m pytest tests/scitex_agent_container/_lifecycle/ \
  tests/scitex_agent_container/cli_pkg/ tests/scitex_agent_container/_creds/ -q
```

New: `tests/scitex_agent_container/_lifecycle/test__start_noop_notice.py` (8),
`tests/scitex_agent_container/_creds/test__no_healthy_account_brief.py` (5).

Version **0.24.18**, re-checked at push time. develop is 0.24.16; **#813 took
0.24.17 after this branch first bumped**, so 0.24.17 was vacated rather than
shipped — a collision there is exactly the regression that bit #812 earlier
today. Remaining open PRs claim below both (#788 / #785 at 0.23.0, #729 at
0.21.22).


ywatanabe1989:
## Field evidence for scitex-dev: bare `print()` in sac's library/CLI paths

Per the operator's ruling, sac does **not** ship its own no-bare-print rule —
scitex-dev owns the abstracted primitive and sac conforms to it later. This is
the inventory they asked for *before* fixing the rule's shape.

**48 bare `print(` calls across 22 files** (after this PR removed 3 from
`_lifecycle/_start.py`). Not fixed here — inventory only.

The important finding is that **a naive "no print in src/" rule would be wrong
on roughly half of these.** They fall into four groups, and only the last is a
genuine violation.

### Group A — standalone scripts that CANNOT import the logger (14)

These are not imported as library code; they are executed as standalone
processes, some outside the venv entirely. `scitex_logging` is not guaranteed
importable, and for the Claude Code hooks their stdout/exit code IS the
protocol.

| File | n | What it prints |
|---|---|---|
| `_baseline_assets/claude_worktree_hooks/worktree_create.py` | 9 | hook protocol: the resolved worktree path on stdout; validation errors on stderr + `return 2` |
| `_baseline_assets/claude_worktree_hooks/worktree_remove.py` | 4 | hook validation errors (empty stdin, bad JSON, missing key, failed `git worktree remove`) |
| `_baseline_assets/claude_worktree_hooks/session_start_reap_stale.py` | 1 | `_summarize(reaped, skipped, roots)` to stderr |

`containers/sif_symbol_probe.py` (3) belongs here too — it runs *inside a SIF
during the build*, before sac is importable: `FATAL: scitex_todo shim is not
scitex_cards…` / `OK: artifact symbol probe passed`.

`_skills/scitex-agent-container/scitex_smoke_test.py` (1) is a shipped
standalone smoke script: `Summary: N pass, N fail…`.

**Rule implication:** the rule needs a path/entrypoint exemption, not just a
`--json` one. Detecting "has `if __name__ == '__main__'` and is not imported by
the package" would cover it.

### Group B — stdout IS the product (2)

| File | n | What it prints |
|---|---|---|
| `statusline.py` | 1 | `" | ".join(parts)` — the statusline string Claude Code consumes |
| `_hosted_runner_guard.py` | 1 | `format_report(repo, violations)` to an injected `stream`; exit code carries the verdict |

**Rule implication:** this is the `--json`-style exemption generalised —
"machine-readable stdout". Note neither of these has a `--json` flag, so keying
the exemption on that flag name alone would miss them.

### Group C — writes to an INJECTED stream, not stdout/stderr (4)

| File | n | Note |
|---|---|---|
| `_lifecycle/_start_preflight.py` | 2 | `file=log_stream` — a caller passes a throwaway `StringIO` specifically to **suppress** the notice. Routing these to a logger silently breaks that suppression (documented in this PR). |
| `_drift/_local.py` | 1 | `print(line, file=stream, flush=True)` |
| `_events/_verdicts.py` (partial) | — | same shape |

**Rule implication:** `print(..., file=<non-stdout/stderr expr>)` is a test/
capture seam and should not be flagged. This is the subtlest case and the one
most likely to produce a wrong autofix.

### Group D — genuine violations: diagnostics that should carry a level (~28)

These are the real targets. All are operator-facing diagnostics that today
arrive with no level, no colour, and an ad-hoc `[bracket-tag]` prefix instead
of the `XXXX:` column.

| File | n | Example |
|---|---|---|
| `_account/refresh_alarm.py` | 4 | `[refresh-alarm] {account}: recorded in sac's event log, but…` / `ALERT DELIVERY FAILED — {exc}` |
| `_events/_verdicts.py` | 3 | `[sac-events] could not read {target} — {exc}…` |
| `cli_pkg/never_stop_when_task_remains_cmds.py` | 3 | `never-stop-when-task-remains: gate crashed ({exc!r}); allowing the stop (fail-open)` |
| `_reconcile/_pass.py` | 2 | `[fleet-reconcile] REFUSING to restart anything: {detail}` |
| `_authheal/_pass.py` | 2 | `[login-expired-restart] REFUSING to restart {n} wedged…` |
| `_lifecycle/_session_reset.py` | 2 | `[force] cleared persisted session_id for {name!r}` |
| `cli_pkg/_create.py` | 2 | `Wrote {path} (template={kind}).` — a SUCC line with no level, same defect as `refresh-quota-cache` |
| `runtimes/claude_session.py` | 2 | `error: ClaudeSessionRuntime requires a container engine…` — literally the word "error" hand-written in place of a level |
| `_events/_log.py` | 1 | `[sac-events] FAILED to record {event} for…` |
| `_listen/_host_exec.py` | 1 | `host_exec: audit log append failed at {path}: {exc}` |
| `_listen/_standby.py` | 1 | `print(message, file=sys.stderr, flush=True)` |
| `config/_resolve.py` | 1 | `resolved '{name}' → '{matches[0]}' (prefix match)` — pure INFO noise on the default path |
| `_runners/_tmux/_host_cwd.py` | 1 | `WARN: tmux workdir {workdir!r} is not host-creatable…` — hand-written `WARN:` prefix, exactly what the level would supply |

### Two patterns worth encoding in the rule's *message*

1. **Hand-written level prefixes.** `WARN: …`, `error: …`, `FATAL: …` appear as
   string literals inside `print()`. These are the clearest true positives — the
   author wanted a level and had no rail. A rule that specifically calls these
   out would have a very low false-positive rate.
2. **Ad-hoc bracket tags.** `[sac-events]`, `[fleet-reconcile]`,
   `[login-expired-restart]`, `[sac:creds]`, `[force]` — each is a hand-rolled
   module attribution, i.e. a manual reimplementation of
   `getLogger(__name__)`. Worth naming in the fix hint.

### Suggested rule shape (sac's field opinion, not a design)

Flag `print(...)` when **all** hold:
- the module is imported by the package (not a standalone entrypoint), and
- the destination is `sys.stdout` / `sys.stderr` / default — not an injected stream, and
- the enclosing command is not in a machine-output mode.

Exempt-by-marker would also work and is cheaper to get right than inference.
sac will adopt whatever shape scitex-dev lands on.

ywatanabe1989:

**What this PR attempted:** feat(cli): honest output levels — short lines, no traceback on an expected refusal

**Why it is being closed:** conflicts with its base; 19 days old, so CI never returned a verdict on it in its current form.

**Nothing is deleted.** Branch `feat/cli-output-ssot` still holds every commit, and this PR can be reopened at any time. If the change is still wanted, a fresh PR against today's default branch is the faster path.

PR #815, opened 2026-07-22
fix(apptainer): argv guard names the culprit instead of always blaming raw_args

ywatanabe1989:
An orphan `- --env` in three sibling specs (`scitex-cards-{chat,gui,mobile}`) made `sac-start` fail, and the guard's message told the operator to fix `spec.apptainer.raw_args` without saying **which** spec — so the wrong file got audited.

## Mechanism
The orphan is the remnant of deleting an env var's **value** line while leaving its **flag**. An edit artefact, not a typo — it hit three specs at once, and the same variable was removed cleanly in a fourth.

Swept the real argv for **all 101 host specs** and partitioned them:
```
CLEAN 86 | BUILD FAILED 12 (unrelated, /dev/fuse absent) | SPEC-SIDE 3 | SAC-SIDE 0
```
The `SAC-SIDE 0` is also backed structurally, not just by the sweep: all 37 `--env` emission sites append flag+value in one statement, and the only argv-mutating step removes pairs whole (`i += 2`).

## Change
`validate_flag_argv(argv, *, raw_args=None, agent=None)` reports provenance as a **tri-state**:
- `raw_args` malformed → names the index and the agent
- `raw_args` clean → "NOT the spec … editing it cannot fix this … this is a sac bug"
- `raw_args` absent → `CAUSE UNKNOWN` (does not guess)

Those three branches are exactly the three inputs that make it say something different, so the attribution is testable rather than decorative. Detection logic untouched.

## Mutation proofs
| | |
|---|---|
| baseline | 17 passed |
| plant: blame a clean `raw_args` (old behaviour) | 1 failed, 16 passed |
| plant: guard made inert | 10 failed, 7 passed |
| revert | 17 passed |
| guard + build_argv | 68 passed, 2 skipped |
| standalone harness | **5/5 mutations killed, 0 survived** |

## Follow-up (not in this PR)
Wire `find_missing_value` into `sac agents check` so an orphan is caught pre-flight rather than at launch — carded as `sac-agents-check-should-catch-orphan-flags-preflight-20260723`.

### DIFF
--- a/src/scitex_agent_container/runtimes/_apptainer_argv_guard.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_argv_guard.py
@@ -37,6 +37,31 @@ Only the curated + raw flag region is checked. The inner command
 (``bash -c "<preflight>\nexec …"`` or the flat relaxed inner argv) sits
 AFTER the SIF and is the container's business, not apptainer's flag
 parser — so it is excluded.
+
+Naming the culprit
+------------------
+The first version of this message ended with a single unconditional
+sentence: "Fix the spec.apptainer.raw_args ordering." That is the right
+remedy for the common case and the WRONG one for two others — sac's own
+assembly can in principle emit the bad pair, and the message never said
+WHICH agent's spec to open. On 2026-07-23 an operator hit the ``--env
+--env`` form on ``scitex-cards-{chat,gui,mobile}`` (an orphan ``--env``
+left behind when a ``KEY=VALUE`` line was deleted), read the message,
+opened the similarly-named ``scitex-cards`` spec — which was clean — and
+concluded the guard was lying. It was not; it just never named the file.
+
+So the message now reports PROVENANCE as a tri-state, computed from the
+``raw_args`` the caller passes in:
+
+* ``raw_args`` malformed          → the SPEC is at fault; the offending
+  ``raw_args`` index is named.
+* ``raw_args`` present and clean  → the spec is EXONERATED in the text
+  and the reader is told sac's assembly introduced it.
+* ``raw_args`` not supplied       → provenance is stated as UNKNOWN
+  rather than guessed.
+
+The offending flag-region index and a window of neighbouring tokens are
+always printed, so the reader sees the actual argv instead of a rule.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
**CI is red for a reason unrelated to this PR — holding, not merging.**

`pytest-matrix` fails on py3.11/3.12/3.13 with:
```
1 failed, 12450 passed, 65 skipped
summary: 53 unmasked error(s), 0 masked by skip-rules (0 declared)
[E] [PS-220 §2 source-uses-print-not-scitex-logging]
```

Every violation is in a file **this PR does not touch** — `_events/_verdicts.py`, `_baseline_assets/claude_worktree_hooks/*`, `_runners/_tmux/_host_cwd.py`, `_skills/…/scitex_smoke_test.py`.

**Timing:**
| | |
|---|---|
| develop's last CI run | 2026-07-22 10:03Z — green |
| this PR's run | 2026-07-22 17:44Z — red |
| the `print()` calls | already on develop |
| `PS-220` in `.scitex/dev/config.yaml` | absent |

PS-220 was promoted to error severity inside that window. **develop is green by staleness, not by health** — it hasn't been asked the new question. Every open PR in this repo is blocked the same way, and develop goes red on the next merge.

**Not declaring exemptions to unblock this.** The rule requires a per-site path + line + mandatory reason; inventing 53 reasons to get a PR through is the grandfathering-by-blanket-flag the constitution names as worse than no gate. Reported to scitex-dev, who own the rollout. Happy to fix sac's 53 properly as its own scoped PR — the rule looks correct — but that's their sequencing call, not mine.

ywatanabe1989:
**Retracting the hold I posted above — it no longer applies.**

That comment said this PR was blocked by 53 PS-220 `source-uses-print-not-scitex-logging` errors and that I would not declare exemptions to clear them. Both the block and the reasoning are now obsolete, and the reason is not anything in this PR.

PS-220 was promoted to error ecosystem-wide on 2026-07-22 (scitex-dev PR #406). The operator restaged it back to **warning-by-default on 2026-07-23**, per the directive quoted verbatim in the rule's own source (`_cli/audit/_project/_check_no_print.py`), citing a measured blast radius of 44 repos newly failing on 1856 findings. A package now opts IN to error via a written `audit.enforce-logging: {level: error, reason: ...}` declaration once its migration lands.

CI installs `scitex-dev[cli-audit]` unpinned, so the re-run picked up 0.36.0 and those 53 became warnings. **This PR is now green with no code change** — the red was a transient of a promotion that has since been withdrawn, not a defect here.

Verified rather than assumed: calling the rule directly on this tree returns 53 findings at severity W; planting one more returns 54; removing it returns 53 again.

sac's print migration is still real work and is tracked separately — it is simply not this PR's gate.

PR #816, opened 2026-07-22
fix(accounts,tui): zero accounts is not success, and pane tails are not errors

ywatanabe1989:
Two operator-reported defects in the start path.

## 1. `refresh-quota-cache` treated "nothing to do" as success

```
$ sac accounts refresh-quota-cache     # container, empty account store
No accounts stored. … (wrote empty cache to …/quota-cache.json)
wrote 0 account(s) (0 failed)
exit=0
```

`if attempted > 0 and result["ok"] == 0` skipped the zero-account case entirely. The comment stated the intent correctly; the guard left the case uncovered.

**The harm is worse than the original report.** `merge=True` seeds from the existing cache before the account loop, so prior entries **did** survive — the cache was not destroyed. What actually happened, both measured:
1. `written_at` restamped to now — a stale cache that claims to be current.
2. With **no** prior cache, an empty `{"accounts": {}}` was **created** — and that file flips `quota_cache_present()` to `True`, which `_start_preflight.py:213,355` passes as `require_quota_evidence`. **So the suggested remedy can arm the boot gate that then refuses the start.**

Now: zero accounts returns `written=False`, `reason="no-accounts"`, writes nothing, and exits **3** — not 2, because click owns 2 for `UsageError` and a caller could not otherwise tell an empty store from a mistyped flag. "Every account failed" still writes, so the merge keeps meaning something.

Also corrected: "no accounts stored" never reaches the `_pick_healthy.py:396` refusal — it raises at :268, which already names `sac accounts save`. What :396 *can* distinguish is cache content, so the new `quota_cache_entry_count()` branches zero-entry vs populated-but-mismatched.

## 2. Pane transcriptions were logged at ERROR

Every line of another session's screen carried an `ERRO:` prefix, so a successful start read as a wall of failures before the final `SUCC:`. Pane content now emits as its own INFO record (`runtimes/_pane_context_log.py`); the real fault stays ERROR and is now single-line.

## Mutation proofs — baseline 47 passed
| mutation | result |
|---|---|
| restore the `attempted > 0` hole | 2 failed, 10 passed |
| write the empty cache again | 5 failed, 15 passed |
| one remedy for both causes | 3 failed, 4 passed |
| pane transcription back to ERROR | 3 failed, 5 passed |
| demote the real fault to info | 2 failed, 6 passed |
| **revert** | **47 passed** |

The last mutation exists so `test_the_fault_itself_stays_at_error_level` isn't vacuous. Every empty-store assertion has a populated-store counter-input that answers differently.

Regression: **1259 passed, 14 skipped**.

## Noted, not fixed
In a container the writer resolves to `$HOME/.scitex/agent-container/runtime/quota-cache.json` while the reader prefers the `/var/sac/quota-cache.json` bind when present. Without the bind they coincide — which is how the empty file armed the gate. That path split deserves its own look.


ywatanabe1989:
**CI is red for a reason unrelated to this PR — holding, not merging.**

`pytest-matrix` fails on py3.11/3.12/3.13 with:
```
1 failed, 12450 passed, 65 skipped
summary: 53 unmasked error(s), 0 masked by skip-rules (0 declared)
[E] [PS-220 §2 source-uses-print-not-scitex-logging]
```

Every violation is in a file **this PR does not touch** — `_events/_verdicts.py`, `_baseline_assets/claude_worktree_hooks/*`, `_runners/_tmux/_host_cwd.py`, `_skills/…/scitex_smoke_test.py`.

**Timing:**
| | |
|---|---|
| develop's last CI run | 2026-07-22 10:03Z — green |
| this PR's run | 2026-07-22 17:44Z — red |
| the `print()` calls | already on develop |
| `PS-220` in `.scitex/dev/config.yaml` | absent |

PS-220 was promoted to error severity inside that window. **develop is green by staleness, not by health** — it hasn't been asked the new question. Every open PR in this repo is blocked the same way, and develop goes red on the next merge.

**Not declaring exemptions to unblock this.** The rule requires a per-site path + line + mandatory reason; inventing 53 reasons to get a PR through is the grandfathering-by-blanket-flag the constitution names as worse than no gate. Reported to scitex-dev, who own the rollout. Happy to fix sac's 53 properly as its own scoped PR — the rule looks correct — but that's their sequencing call, not mine.

PR #877, opened 2026-08-05
fix(listen-probe): stop telling callers to retry an operation that probably succeeded

ywatanabe1989:
scitex-hub hit a 60s timeout on POST /agents/scitex-db/restart, reported it as
"dead agents cannot be revived", and — following this message's own advice —
retried. Both calls performed the restart. The second tore down the session the
first had just brought up. scitex-db confirmed afterwards that its inbox came
back at 07:28, the exact moment the caller recorded a failure.

The message was wrong in two ways, and the second is worse than the first.

1. It recommended RETRY for an operation that is not idempotent and that had
   just completed invisibly. "Retry" is the natural instinct on a timeout and
   here it is actively harmful: an agent mid-turn loses that turn.

2. It asserted a MECHANISM it had never established — "the route has been seen
   to answer and then degrade". Nothing degrades. Lifecycle routes await the
   whole operation before answering, so response time IS operation time: a
   SUCCESSFUL restart measured 77.4s on this fleet, a cold boot runs past the
   240s boot-drain window. A warm session-resume finishes inside 60s and a cold
   one does not — that is variable workload, not degradation. This function's
   own docstring says "do not speculate about *why*"; that clause did exactly
   that, and both scitex-hub and I built wrong conclusions on it.

The replacement says what was measured and what to do: a timeout here does NOT
mean the operation did not happen, verify whether the agent came up
(inbox_subscribers / session_id / sac agents list) BEFORE retrying, and report
only if it genuinely did not.

This is the advice half only, fixed first because it is one string and it is
causing restart loops today. The route itself still performs its work
synchronously and cannot answer inside any acceptable client timeout; per
constitution §2 the fix there is to make "still running" REPRESENTABLE — an
accepted-style code plus a handle to poll — not a longer timeout, which merely
moves the collapse to 303s. Tracked separately on
sac-agent-restart-route-wedged-cannot-revive-agents-20260806.

https://claude.ai/code/session_01JbkepAbn5pDEFCXXJx9KKd


ywatanabe1989:
## Live confirmation, 2026-08-06 — the old text would have caused a double restart

Hit this for real while restarting `scitex-cards`, and the sequence is exactly
the one this PR is about.

**What the route did**

```
POST /agents/scitex-cards/restart   -> NO RESPONSE within 60s (timed out)
GET  /v1/health                     -> HTTP 200 in 0.042s   (same seconds)
GET  /v1/host_exec/inflight         -> HTTP 200 in 0.020s   (same seconds, authed)
```

**What the old text told me to do**

> Retry, and if it recurs report POST /agents/scitex-cards/restart with these
> two timings; the route has been seen to answer and then degrade, so a single
> success afterwards does not mean it was fixed.

**What was actually true**

I verified instead of retrying:

```
tmux ls | grep cards
tui-scitex-cards: 1 windows (created Thu Aug  6 09:03:50 2026)
session count: 12   (back to full)
```

**The restart had succeeded.** The POST timed out at 60s because
`_agent_restart.py:307` awaits the subprocess to completion, so response time
IS operation time — and a restart measures ~77s. The 60s client timeout is
structurally shorter than the operation it waits on, so this route CANNOT
answer for a healthy restart. The timeout is the normal case, not the fault.

Following the old advice would have fired a second restart into an agent that
was already mid-restart. The advice inverts the correct action at precisely the
moment it is read.

This is the same class the health-gate work keeps landing on: a timeout is a
statement about the OBSERVATION CHANNEL, not about the world. It is a third
value — unknown — and collapsing it into "did not happen" is what produces the
duplicate destructive action.

Also worth noting for whoever picks up the route fix: the diagnostic is
otherwise excellent. It correctly rules out daemon-wide failure with two
contemporaneous positive controls on other prefixes. It earns trust with good
evidence and then spends that trust on the one instruction that is wrong.

ywatanabe1989:

ywatanabe1989:
## Conflict is semantic — stopping rather than resolving it

This PR was `DIRTY`. I set out to merge `origin/develop` in (base confirmed as `develop`,
not `main`) so CI could run again. The merge conflicts in exactly one file —
`src/scitex_agent_container/_lifecycle/_listen_probe.py` — and the conflict is **not
mechanical**, so I aborted it rather than pick a side. Nothing was pushed; the branch is
untouched.

### Why it needs the author, not a merge tool

**#956 landed on develop while this PR waited** — *"fix(listen-probe): stop asserting a
route-specific fault the probes cannot observe"*. It rewrote the **same advisory
paragraph** this PR rewrites, from a partly-overlapping premise.

The two sides of the conflict:

- **this PR** replaces the tail with `DO NOT RETRY BLIND — A TIMEOUT HERE DOES NOT MEAN
  THE {VERB} DID NOT HAPPEN` plus `VERIFY BEFORE RETRYING` and the 77s / 240s
  measurements.
- **develop (#956)** kept a retry paragraph but reframed it: *"The same route has been
  seen to answer and then not answer, which is equally consistent with a genuine fault
  AND with an operation whose duration varies either side of {timeout_s}s. One success
  discriminates neither."*

That is the crux. This PR's stated objection is that the old text *"asserted a MECHANISM
it had never established — 'the route has been seen to answer and then degrade'"*. #956
did not delete that observation; it **kept it and made it explicitly
non-discriminating**, which concedes much of this PR's point in different words.

So resolving requires a judgement only you can make:

1. Does #956's rewording already satisfy the objection, leaving only the
   **`VERIFY BEFORE RETRYING` guidance** as this PR's novel contribution? (That guidance
   is genuinely **absent** from develop — it is the part worth saving.)
2. Or does the surviving sentence still overclaim and need to go?

Taking *this* side verbatim deletes #956's landed refinement. Taking *develop's* side
drops the do-not-retry-blind guidance, which is this PR's whole reason to exist. A
verbatim union produces a message that says "verify before retrying" and then "if you
retry and it succeeds…" — coherent but redundant, and it would be me authoring your prose
in your voice.

### Suggested shape

Rebuild the tail on top of develop's current text: keep #956's "not daemon-wide"
sentence, insert this PR's `DO NOT RETRY BLIND` / `VERIFY BEFORE RETRYING` block with the
77s and 240s numbers, and drop whichever of the two closing retry sentences the other
makes redundant. One merge commit, your wording.

### Also worth knowing before you re-run

All three matrix legs are red on this PR's current head, but **that head predates the
gate change**: it emits a `py3.12` leg, and develop's PR gate is now min+max only
(`["3.11","3.13"]`, 3.12 moved to the hosted nightly). Only `py3.11` and `py3.13` are
required contexts now. Re-measure after merging develop before treating any of those
three failures as real — some of it is a leg that no longer exists, and develop carried
fleet-wide breakage during that window.

ywatanabe1989:
**This PR is `DIRTY`/`CONFLICTING` and has NO checks at all.** CI cannot build a merge commit, so its BLOCKED state carries no test verdict either way.

```
src/scitex_agent_container/_lifecycle/_listen_probe.py
```

The branch is **172 commits behind** `develop` (merge-base bd4b5fa, 2026-08-05) — the oldest in the open set. Before resolving the conflict, it is worth checking whether this is still needed: `develop` has since landed 53183ed, *"fix(listen-probe): stop asserting a route-specific fault the probes cannot observe"* (#956), and cd199ef, *"fix(listen-clients): a client timeout must OUTLIVE the server's deadline"* (#960) — both in `_listen_probe`'s territory and both about the probe's retry/verdict behaviour, which is what this PR set out to change.

(Posted by an automated PR-backlog sweep. Nothing was changed on this branch, and no PR was closed or retargeted — that call belongs to its author.)

ywatanabe1989:
**This PR is `DIRTY`/`CONFLICTING` and has NO checks at all.** CI cannot build a merge commit, so its BLOCKED state carries no test verdict either way.

```
src/scitex_agent_container/_lifecycle/_listen_probe.py
```

The branch is **172 commits behind** `develop` (merge-base bd4b5fa, 2026-08-05) — the oldest in the open set. Before resolving the conflict, it is worth checking whether this is still needed: `develop` has since landed 53183ed, *"fix(listen-probe): stop asserting a route-specific fault the probes cannot observe"* (#956), and cd199ef, *"fix(listen-clients): a client timeout must OUTLIVE the server's deadline"* (#960) — both in `_listen_probe`'s territory and both about the probe's retry/verdict behaviour, which is what this PR set out to change.

(Posted by an automated PR-backlog sweep. Nothing was changed on this branch, and no PR was closed or retargeted — that call belongs to its author.)

ywatanabe1989:

PR #881, opened 2026-08-06
fix(ci): make the xdist worker count an input, because its premise just died

ywatanabe1989:
## The premise in the code is no longer true

`run-in-sif.sh` sets `WORKERS=$NPROC` with no cap, and states why:

> Each matrix leg now runs on its own dedicated self-hosted node (one runner per
> node: scitex-agent-container-01/02/03), so there is no co-tenant to yield half
> the box to

That died on 2026-08-05, when Spartan CI was retired
(「スパルタンのCIは全面的にやめましょう」). The suite now runs on **scitex-04 —
one 32-core machine**. Put three runners on it so the matrix stops serialising,
and you get **3 × 32 = 96 xdist processes over 32 cores**.

Same shape as the `"runs here are serialised (one job at a time)"` claim that
once justified an unscoped `pkill` in `exec-in-sif.sh`: a default whose safety
rests on a fact about the deployment, left in place after the deployment
changed.

## The change

```sh
WORKERS="${CI_XDIST_WORKERS:-$NPROC}"
```

- **Unset keeps today's behaviour exactly** — a one-leg-per-node deployment is
  unchanged, so this cannot regress anyone who does not opt in.
- A non-numeric value warns and falls back to `nproc` rather than handing
  garbage to `-n`.
- Verified: unset → 32, `"10"` → 10, `"abc"` → 32 + warning.

Wired into **both** gates. The PR workflow's header exists to keep the PR gate
and the release gate the same bytes in the same environment — the drift that
left PyPI on 0.21.17 while nine tags shipped nothing. A worker knob on one and
not the other would re-open that gap.

## Measured

scitex-04, 32 cores, full suite, one leg at a time:

| workers | wall clock |
|---|---|
| 32 | 170s |
| 8 | 233s |

One leg alone is faster with every core. Three **concurrent** legs at ~10 each
finish the whole matrix in about one leg's time instead of three — which is the
point: with a single runner the matrix serialises today and takes **longer
end-to-end than Spartan did**, even though each leg is 3× faster.

## Scope

This does not set the variable. It makes setting it possible, which is the
prerequisite for adding runners to scitex-04 without oversubscribing it.

Related, and deliberately **not** claimed here: `develop` has a flaky
`test_hanging_remote_probe_respects_timeout_budget` that asserts a wall-clock
budget. It failed repeatedly earlier today and then passed a full clean run
(`12836 passed in 144.33s`, zero leaked processes). Every earlier measurement of
mine ran while the CI runner was also busy on the same box, so none of them
controlled for load. Oversubscription would make that flake constant, which is
another reason this knob should exist before more runners are added — but the
flake's cause is still open and is tracked separately.


ywatanabe1989:

**What it is waiting on:** its **own CI**. Failing or missing required checks: （required check missing or a required review is absent）.

ywatanabe1989:
## Unstalled: the missing matrix legs were a wedged run, not a protection mismatch

This PR sat `BLOCKED` with five green checks and **no `pytest-matrix` legs at all**. The
obvious hypothesis — branch protection requiring a context the PR never emits — is
**wrong here**, and the cheap detector says so:

| | |
|---|---|
| required on `develop` | `pytest-matrix-on-ubuntu-py3.11`, `pytest-matrix-on-ubuntu-py3.13` |
| resolves to | `"app_id": 15368` (GitHub Actions) — **not** `null` |

An unreported/never-emitted required context resolves to `app_id: null`. Both of these
resolve to a real app, and sibling PRs on the same base (#877, #884) emit exactly those
names. Protection is correctly configured. Nothing to reconcile.

### What was actually wrong

A `tests` run for head `d0cb7a7` **did exist** — and had been wedged since **2026-08-06**:

```
run 31122173390  name=tests  status=queued  conclusion=null
run_attempt=2    created=2026-08-06T17:07:32Z  updated=2026-08-06T21:54:58Z
jobs total_count = 0
```

`run_attempt: 2` with **zero jobs ever materialised**, parked in `queued`. It emits no
check runs, so the two required contexts never arrive and the PR can never leave
`BLOCKED`. And it cannot be nudged loose:

```
$ gh run rerun 31122173390
run 31122173390 cannot be rerun; This workflow is already running
```

GitHub believes the run is in flight, so it will neither finish nor restart. The only
escape is a **new head SHA**. Worth noting for whoever triages the next one: this was
**six days** stalled, not nine hours — the PR's `updatedAt` moved on 08-11 but the head
commit, and its wedged run, are from 08-06.

Ruled out along the way: the workflow YAML on this branch parses fine (`jobs: ['test']`),
so it was not a startup failure.

### What I changed

Merged `origin/develop` in (no rebase — force-push is barred). Clean merge, and it
**needed** checking, because develop had edited the same `WORKERS` lines. Both survive,
in the right order:

```sh
NPROC="$(nproc 2>/dev/null || echo 4)"
WORKERS="${CI_XDIST_WORKERS:-$NPROC}"   # this PR
case "$WORKERS" in
    ''|*[!0-9]*)
        echo "::warning::CI_XDIST_WORKERS='$WORKERS' is not a positive integer; using nproc=$NPROC"
        WORKERS=$NPROC
        ;;
esac
[ "$WORKERS" -lt 4 ] && WORKERS=4       # develop's floor, applied after the override
```

The branch also picks up develop's min+max PR gate (`["3.11","3.13"]`; 3.12 moved to the
hosted nightly), so the legs this PR now runs are the two that are actually required.

I did not touch this PR's code. Merge commit only.

### On #1011 — complements, not alternatives

**#1011** (`fix/xdist-workers-from-affinity`) changes the same line, deriving the worker
count from the kernel's affinity mask instead of `nproc`. These are not competing:

- an **override with a broken default** is still broken for everyone who does not set it;
- a **good default with no override** is inflexible.

**Whoever lands second should keep both** — the affinity-derived value becomes the
`:-` fallback, the `CI_XDIST_WORKERS` env override stays in front of it, and the
`-lt 4` floor stays after. All three compose in one expression. Posting the same note on
#1011.

### Correction to an earlier version of this comment

I first wrote that the idle `spartan-cpu` runners were stranded beside a saturated queue.
**That reading was out of date and I am retracting it.** The light-lane split has landed
(**#1006**, `03edfa1`): `lint`, `docs`, `quality`, `import-smoke` and
`no-hosted-runners` now pin `runs-on: ["self-hosted","Linux","X64","spartan-cpu"]`
directly, and come back in seconds. Those runners are working, not stranded — the five
green checks on this PR are exactly that lane.

What remains true: `CI_RUNS_ON` is `["self-hosted","Linux","X64","scitex-org-cpu"]` and
governs only the **heavy matrix**, which stays on the four org runners **deliberately**.
And the real argument for this PR's knob is sharper than "the queue is tight": on the
Spartan allocation `nproc` reported **1** inside a 48-CPU allocation, so the suite ran at
the `-lt 4` floor. That derivation is what **#1011** fixes — which is precisely why the
override and the corrected default belong together.

PR #882, opened 2026-08-06
fix(agents list): one shared deadline for the remote probe pass (NOT the CI-red fix its commit message claims)

ywatanabe1989:
## The commit message on this branch overclaims. Read this instead.

`95208a6` says this fixes the red that appeared when CI moved to scitex-04.
**It does not.** I ran the A/B afterwards — full suite, real entrypoint, fixed
arm vs unfixed arm — and **both arms failed identically (1 failure each)**. This
change is not the cause of that failure and does not repair it. I could not
force-push a correction onto a pushed commit, so the record is corrected here.

What actually explains the red is flakiness, proven by run
[31120506600](https://github.com/scitex-ai/scitex-agent-container/actions/runs/31120506600):
one commit, one machine, **py3.11 SUCCESS / py3.12 FAILURE / py3.13 SUCCESS**.
Three arms of the same matrix disagreeing on identical code is not a defect this
diff can reach. That failure is carded separately and needs a decision on the
assertion, not a fix here.

## What this change *is* worth on its own

An independently verifiable defect in `_probe_remote_statuses`, readable
straight from the source without any experiment:

```python
for future in list(future_to_name):
    statuses[name] = future.result(timeout=probe_timeout_s)   # deadline RESTARTS
```

`future.result(timeout=T)` called once per future in submission order gives each
future the **full** budget. `n` stalled probes therefore cost roughly
`ceil(n/workers) * T`, not `T` — the pool is parallel but the *waiting* is
serialized. That is exactly what the function's own docstring promises cannot
happen:

> so one wedged peer cannot serialize (or hang) the whole `sac agents list`

The fix is one shared wall-clock deadline, mirroring what
`_agent_list.get_agent_list_data` already does for the local pass. That local
fix even carries the comment *"Same defect, same fix, as
`_agent_list_discover._probe_remote_statuses`"* — while the fix had never landed
here. A comment asserting a remedy is not the remedy.

Abstention semantics are unchanged: a probe that runs out of budget is
`"unknown"` — hidden from the default view, counted in the footer, never a green
row.

## Line cap: this PR crosses it, but the cap is not a build gate

This section has been corrected twice. Here is the settled version.

| | lines | bytes |
|---|---|---|
| `origin/develop` | **512** | 23574 |
| `origin/fix/probe-budget` | **523** | 24323 |

My first draft said develop was already over at 513 — wrong. develop's copy of
this file sits *exactly at* 512, so this PR is what crosses the line, by 11.

But I then implied that made it a blocker, and the evidence says otherwise.
Counting every tracked `.py` on this branch:

- **`src/`: 9 files already exceed 512** on develop — up to **870**
  (`_workdir_audit.py`), including this file's own sibling `_agent_list.py` at
  532.
- **`tests/`: 103 of 762 files exceed it**, up to 3239.

develop is green with all of those in place, so `scitex-dev quality audit-lines`
does not hard-fail the build above 512. The cap is advisory here, or carries
exemptions. It is still worth respecting, and the split for this file is carded
separately — but it is not a reason to hold this PR.

I could not confirm that by running the auditor: the `scitex-dev` install in
this container is broken (two dist-info directories; `audit_line_limits.py`
missing from the installed tree; both the deprecated and current entrypoints
exit 2 without evaluating the rule). Carded. The conclusion above rests on the
file census plus develop being green, not on the tool.

## Verification — the gap is now closed

An earlier revision of this section said I had no test that fails before and
passes after, and offered to merge on a source-level argument alone. That gap is
filled; the second commit adds the test.

| source under test | result |
|---|---|
| develop's per-future deadline | **FAIL — 15.9x** (16 wedged peers cost 4.81s vs a 0.30s single-peer baseline) |
| this branch's shared deadline | **PASS** |

The measured **15.9x lands on n_peers = 16**. The test does not merely observe
"slower" — it observes slower by exactly the factor the defect predicts, which
is what makes it evidence for the mechanism rather than evidence of a busy
machine.

**It asserts a ratio, not a bound, on purpose.** The absolute-bound form already
exists — `test_cli.py` asserts `elapsed < 3.0` against a 1s budget — and that is
precisely the assertion that goes red under 32-way xdist load, because load
inflates the measurement while the bound stays put. Comparing against a
single-peer baseline taken on the same machine moments earlier moves both arms
together. At 15.9x vs ~1x, the 5.0x threshold has wide room on both sides.

**No mocks**, per the suite's own header: the probe is an ordinary function
waiting on a real `threading.Event` — what a peer whose ssh never answers looks
like from the pool's side — and `_probe_remote_statuses` runs its real
`ThreadPoolExecutor` over real futures. `max_parallel_probes=1` deliberately: a
pool wide enough to run the batch at once hides the defect, because what
serializes is the *waiting*, not the work.

Whole remote suite: **23/23 pass**. The before-arm swapped develop's source in
and back out; the tree is byte-identical to HEAD afterwards.

Still true, and unchanged by this: the CI red that the first commit message
blamed on this code is a **separate flake**, and this test deliberately does not
touch it.

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_helpers/_agent_list_discover.py
+++ b/src/scitex_agent_container/cli_pkg/_helpers/_agent_list_discover.py
@@ -323,6 +323,7 @@ def _probe_remote_statuses(
     not run is not evidence a remote agent died, but neither is it evidence it
     is running, so it must not render as a comforting green.
     """
+    import time as _time
     from concurrent.futures import ThreadPoolExecutor
     from concurrent.futures import TimeoutError as _FuturesTimeout
 
@@ -335,13 +336,23 @@ def _probe_remote_statuses(
         future_to_name = {
             pool.submit(probe, c["name"], c["host"]): c["name"] for c in candidates
         }
+        # ONE shared wall-clock budget for the batch. future.result(timeout=T)
+        # in submission order restarts the deadline PER FUTURE, so n stalled
+        # probes cost about ceil(n/workers)*T even though the pool is parallel
+        # -- precisely the serialization the docstring above promises cannot
+        # happen. The local pass in _agent_list.get_agent_list_data was fixed
+        # for todo#254 and its comment already claims "same defect, same fix, as
+        # _probe_remote_statuses"; the fix never landed here. A comment is not a
+        # fix, and only the test caught the difference.
+        _deadline = _time.monotonic() + probe_timeout_s
         for future in list(future_to_name):
             name = future_to_name[future]
+            _remaining = max(0.0, _deadline - _time.monotonic())
             # stx-allow: fallback (a per-probe timeout/exception is "unknown" —
             # hidden from the default view + counted in the footer; a probe that
             # could not run must not masquerade as a running remote row)
             try:
-                statuses[name] = future.result(timeout=probe_timeout_s)
+                statuses[name] = future.result(timeout=_remaining)
             except _FuturesTimeout:  # stx-allow: fallback (reason: see inline comment)
                 statuses[name] = "unknown"
                 future.cancel()
--- a/tests/scitex_agent_container/cli_pkg/_helpers/test__agent_list_remote.py
+++ b/tests/scitex_agent_container/cli_pkg/_helpers/test__agent_list_remote.py
@@ -26,6 +26,8 @@ from __future__ import annotations
 
 import importlib
 import os
+import threading
+import time
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Callable, Iterator
@@ -42,6 +44,7 @@ from scitex_agent_container.cli_pkg._helpers._agent_list import (
 )
 from scitex_agent_container.cli_pkg._helpers._agent_list_discover import (
     _default_remote_status_probe,
+    _probe_remote_statuses,
 )
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


ywatanabe1989:
## Local full-suite run — not green, and here is exactly what that does and does not mean

CI has still never run on this branch (GitHub Actions has been in `major_outage`
since ~17:07, jobs queued, all five of our runners online and idle). So I ran the
whole suite locally as the substitute:

```
27 failed, 12801 passed, 55 skipped, 55 errors — 22m31s, -n 8, nice 19
```

**Not green.** Rounding that off would be the easy thing and it would be wrong.

### The new regression test survives real parallel load

It does not appear anywhere in the failure output of the 8-worker run. That was
the specific risk of asserting on timing, and it is why the assertion is a
**ratio** against a single-peer baseline taken moments earlier rather than an
absolute bound — an absolute bound is exactly what reddens under load, which is
the flake this PR deliberately does not touch.

### The 27 failures are not attributable to this change

- All **13 files** containing failures/errors reference `_agent_list_discover` /
  `_probe_remote_statuses` **zero** times. Positive control:
  `test__agent_list_remote.py`, which does exercise it, matches **4** times — so
  the search works and the zeros are real, not an empty query.
- That static check is **not airtight** for the two failures inside
  `cli_pkg/_helpers/`, which could reach the function transitively via
  `get_agent_list_data`. For those, stronger evidence: the entire
  `cli_pkg/_helpers/` package passes **195/195 serially** with the fix in place,
  including both of those tests. They fail only under 8-way parallelism.
- The rest are this container's environment, not code: `develop/test_audit`
  (the `scitex-dev` install here is broken — see below), `_dev_jobs` (no
  systemd), `_broker_self` (listen subprocess), `_rename_board` /
  `_rename_cards` (card store), `runtimes/_sdk_common`.

### What this does NOT establish

**That the branch is green.** I have no develop-side arm of the same full run, so
I cannot claim these 27 are pre-existing rather than new-but-unrelated. The claim
is "not attributable to this change" — nothing stronger. Real CI on a clean
runner is still the verdict that matters, and it has not happened yet.

ywatanabe1989:
## The develop control arm ran — retracting my own caveat

When I posted the local full-suite result I attached a limit: *"I have no
develop-side arm of the same full run, so I cannot say these 27 are pre-existing
rather than new-but-unrelated."* That arm has now run.

Same container, same `-n 8`, same command:

| arm | result |
|---|---|
| `fix/probe-budget` (this PR) | **27 failed** / 12801 passed |
| `fix/flaky-probe-budget-assertion` | 34 failed / 12789 passed *(contained a real regression, since reverted)* |
| **unmodified `origin/develop` @1ebfb71f** | **28 failed** / 12799 passed |

**25 of this PR's 27 failures also fail on unmodified develop**, and develop has
*more* failures than this branch (28 vs 27).

Only two of the 27 are absent from develop, and neither is attributable here:

- `test__agent_list_perf.py::test_ports_come_from_a_single_list_claims_call` —
  passes **serially**; I ran that whole package on this branch and got 195/195,
  including this test. Parallel artefact.
- `test__sdk_heartbeat_loop.py::test_heartbeat_at_renders_as_iso_8601_for_the_read_side`
  — formats a timestamp for a read side. No path from `_probe_remote_statuses`.

**So: this PR does not introduce test failures.** The count does not rise with
the change; it varies around 27–28 as sampling noise in this container, and
develop sits at the high end of that range.

### Cross-check from real CI

A clean 32-worker runner, same day, same tree family: the `tests` leg failed on
**two** things only — `test_audit_all_clean` (a broken `scitex-dev` install in
CI, carded separately) and the four peer tests from the *other* branch's
regression. So roughly 27 of what my container produces does not reproduce on
real CI at all.

That is a fact about the container, not about the code — and it is exactly why
the local number should never have been read as a verdict on this branch. I
should have run the comparison before reporting the observation.

PR #883, opened 2026-08-06
fix(tests): the timeout-budget tests never ran their own fixture — repair the seam, then widen the margin

ywatanabe1989:
## The red I set out to fix was the smallest of three problems

develop's `tests` leg has been red on and off all day on
`test_hanging_remote_probe_respects_timeout_budget` — observed **3.78s against a
3.0s bound**. I opened this to widen the bound. Then I ran the control, twice,
and each time the real problem was one layer further down.

> Two commits. The first fixes the instance; the second fixes the cause and
> **removes the first commit's patch**, which it makes redundant. Read the diff
> as a whole — the net change to `test_cli.py` is the `_HANG_S` constant and the
> assertion, not the fixture bodies.

### Layer 1 — the fixture had stopped loading

```
probe_local_detail CALL COUNT = 0
load_config(test-remote.yaml) RAISED ValueError:
  - spec.host or spec.hosts is REQUIRED
```

`load_config` raises → `cfg` is `None` → `probe_targets` is empty → **no probe
is submitted at all**. `_HangingRuntime.is_running` — the entire point of the
fixture — was never called. Elapsed measured **0.366s** against a 10s hang.

All three tests in the class were wrong, in two different ways:

| test | what was actually happening |
|---|---|
| `respects_timeout_budget` | asserts `elapsed < 3.0` on a number that never approaches the hang. **Cannot fail for the reason it names**, but *can* fail from scheduling delay under 32-way xdist — a test that fails only when it shouldn't. |
| `marks_status_unknown` | passes **vacuously** — the row is `unknown` because the config didn't load, not because a probe timed out. |
| `marks_liveness_unknown_true` | same. Same word, different reason. |

### Layer 2 — one asymmetry, 58 files exposed

| helper | defaults `host` when the caller declares neither? | used by |
|---|---|---|
| `explicit_doc()` (dict path) | **yes** — "so single-purpose tests stay terse", per its own docstring | — |
| `explicitize_yaml()` (string path) | **no** | **58 test files** |

`spec.host` became REQUIRED on 2026-06-24. From that day, any fixture built
through the string path silently took the `cfg is None` branch. The second
commit closes the asymmetry, which is why the first commit's hand-patch to the
two fixture bodies is deleted rather than kept.

### Layer 3 — only then, the bound

With the fixture actually loading, a healthy run measures **1.26s**, so the
numbers mean something and can be chosen honestly:

- `_HANG_S = 30.0` (was an inline `10`) — the margin comes from the **gap**
  between healthy (~1.3s) and regressed (~30s), not from a tight constant.
- `assert elapsed < _HANG_S / 3` (10s) — **8x headroom** over healthy, and a
  regressed run sits 3x above it.

The old `< 3.0` was 3x a 1s budget. A loaded machine walks across that, and did.
Tightening or nudging that constant only moves the coin-flip.

## Controls, both directions

**The gate catches the regression it exists for.** Mutation: budget `1.0 → 60.0`,
so the 30s hang is not cut short — precisely the todo#254 regression.

| | mutation result |
|---|---|
| before this PR | **PASSES** — 3/3 green in 1.23s (dead gate) |
| after, fixture hand-patched | **FAILS** — 3/3 red in 92.87s |
| after, fixture fixed via the helper | **FAILS** — 3/3 red in 91.55s |

**The new barrier catches its own absence.** Revert *only* the `explicitize_yaml`
default: **2 failed, 3 passed** — and the two failures are exactly the two
guarding the string path; the dict-path and declared-placement tests correctly
still pass.

My first attempt at that revert was an uncounted `str.replace` and stripped the
default from *both* helpers, which would have made it a two-variable change. It
was repaired and re-run as a single variable before anything was read into it.

## Why a test and not another sweep

A fixture-migration pass ran on **2026-07-21**, edited **this very file**, and
missed these fixtures. It could not have known — a missed fixture and a correct
one are indistinguishable from outside, because the missed one keeps passing. A
sweep cannot establish its own completeness.

`tests/scitex_agent_container/test__explicit_spec_helper.py` asserts the helpers'
output **loads**, through the real `load_config` and the real `AgentConfig` — not
a re-parse of the text we just generated, which would only prove the helper
round-trips itself. When the next field becomes required, **one** test fails
loudly here instead of 58 fixtures silently changing what they exercise.

## Verification

- **47 passed** (`test_cli.py` 42 + the new file 5).
- **I am not claiming this turns develop green.** It fixes one test class and one
  helper. CI has not run on it — GitHub Actions is in `major_outage`, jobs are
  queued, and GitHub's own notice says self-hosted runners may see errors during
  recovery.


ywatanabe1989:

ywatanabe1989:
## Conflict is semantic, and the premise moved — stopping rather than resolving it

This PR was `DIRTY`. I went to merge `origin/develop` in (base confirmed `develop`) so CI
could run again. It conflicts in **three hunks**, all inside the test class this PR
rewrites, and the conflict is not mechanical. Aborted; nothing pushed, branch untouched.

### develop already landed this repair, by a different route

Two commits reached `develop` while this PR waited:

```
ce9ee9d  test(list): make the todo#254 timeout guard actually run its probe
7b57fef  test(list): stop the fake hung probe outliving its own test
```

`ce9ee9d` is **this PR's Layer 1 thesis** — the fixture that never loaded, the guard that
could not fail for the reason it names. That finding was correct and it is now in
`develop`.

### The three conflicts are two different designs for the same fix

| | this PR | develop |
|---|---|---|
| hung probe | `time.sleep(_HANG_S)` with `_HANG_S = 30.0` | `self._release.wait(10)` — an **event**, with a `time.sleep(10)` fallback |
| assertion | `elapsed < _HANG_S / 3` (10s), 8x headroom over a ~1.3s healthy run | — |
| spec fixture | inline v3 YAML per test | shared `TestListJsonTimeoutBudget._SPEC` |

develop's fixture is released by an **event** rather than a wall clock, which removes the
timing dependence this PR set out to widen a margin around. That is arguably the stronger
answer to the same flake — and it makes this PR's Layer 3 (choose a better constant)
largely moot, since there is no longer a constant doing the discriminating.

Deciding whether `_HANG_S`/8x-headroom supersedes the release-event design, or the other
way round, is an authoring call about your own test — not something a merge should settle
by picking a side.

### What is still novel here, and worth keeping

`tests/scitex_agent_container/test__explicit_spec_helper.py` (**+160 lines**) does **not
exist on develop**. Measured against develop, this branch's own delta is just that file
plus ~33 lines of `test_cli.py`.

That guard is the part that survives the collision intact, and it is the part with the
longest-lived value — your own argument for it stands: *"A sweep cannot establish its own
completeness."* A fixture-migration pass edited this very file on 2026-07-21 and missed
these fixtures precisely because a missed fixture keeps passing.

### Suggested shape

Merge `develop`, take **develop's** version of the three conflicted hunks (the
release-event fixture and the shared `_SPEC`), and keep
`test__explicit_spec_helper.py` unchanged. That preserves the durable contribution and
drops only the parts that landed ahead of you. If you still think the wall-clock bound
needs widening on top of the event-based fixture, that is worth its own small PR with the
new numbers measured against develop's design rather than the old one.

### One caveat on re-measuring

This branch predates develop's min+max PR gate — only `pytest-matrix-on-ubuntu-py3.11`
and `py3.13` are required contexts now (3.12 moved to the hosted nightly), so ignore any
`py3.12` leg on the old head. Your own note that CI had never run on this branch still
holds; it has still never run.

ywatanabe1989:
**This PR is `DIRTY`/`CONFLICTING` and has NO checks at all** — `gh pr checks` reports "no checks reported on the branch". CI cannot produce a merge commit to test, so its BLOCKED state is not a test failure; there is simply nothing to read.

```
tests/scitex_agent_container/test_cli.py
```

The branch is **155 commits behind** `develop` (merge-base 1ebfb71, 2026-08-06), the second-oldest in the open set. Given the age, it is worth checking whether the timeout-budget seam this fixes still looks the way it did — `develop` has since landed 3049676 (#1026, test-port holding) and 53183ed (#956, listen-probe assertions) in adjacent territory.

(Posted by an automated PR-backlog sweep. Nothing was changed on this branch — resolving the conflict needs its author.)

PR #884, opened 2026-08-06
docs(restart-login-expired): the help promised an exit 2 the code deliberately does not return

ywatanabe1989:
## The help promised an exit 2 the code deliberately does not return

`--help` listed **"a registered agent with no live session"** among the exit-2
conditions. `_authheal/_pass.py:indeterminate()` excludes exactly that, on
purpose:

```python
return tuple(r for r in self.of(Verdict.UNOBSERVED) if r.reason != "no-session")
```

The **code is right**. The sentence is stale — it describes behaviour that was
tried and reverted, for a reason the source records:

> the roster is spec FILES and this fleet has far more registered agents than
> running ones by design, so every pass carried sessionless reports and the
> supervisor could **NEVER return 0** — not "rarely", never, for any fleet state
> whatsoever. A gate that cannot go green is a gate nobody reads.

## This already cost something

On 2026-08-07 I ran the verb on the host, got `exit 0` with `UNOBSERVED=100`,
compared it against this sentence, and **filed a P1 against correct behaviour** —
proposing a "fix" that would have restored the never-green gate. Retracted the
same session, before anyone acted on it.

A doc that contradicts its own code does not merely fail to inform; it
manufactures false bug reports. That one is the existence proof, and it is why a
one-sentence docstring change gets its own commit rather than riding along
somewhere.

## So the replacement states the reason, not just the correction

Deleting the wrong clause would leave the next reader free to re-derive my
mistake. The new text says *why* sessionless is a determinate reading, notes
that counting it was tried and reverted, and points at
`_pass.py:indeterminate` for the full argument.

It also keeps the operationally important half: the unobserved count still
prints beside the exit code, so `unobserved: 92` next to `0` reads as *"we
looked at all we could, and it is clean"* — never as *"we failed to look"*.

## Verification

- Help renders with the paragraph break intact (`\b` honoured) — checked through
  the real Click runner, not by eyeballing the source.
- Checked `_jobs_plugin.py` for the same stale claim. It references this verb
  **5 times** (positive control — the search works) and does **not** repeat it,
  so the fix is contained to one file.
- Docstring only. No behaviour change, no test change.


ywatanabe1989:

**What it is waiting on:** its **own CI**. Failing or missing required checks: pytest-matrix-on-ubuntu-py3.13.

ywatanabe1989:
## Unstalled: merged `develop` in — and the blocker was not what the shape suggests

This PR read as `BLOCKED` + `MERGEABLE`, the same shape as #881, which invites the
"protection requires a context the PR never emits" diagnosis. **That is not what was
happening here.** This PR emitted all three matrix legs:

```
pytest-matrix-on-ubuntu-py3.11   SUCCESS
pytest-matrix-on-ubuntu-py3.12   SUCCESS
pytest-matrix-on-ubuntu-py3.13   FAILURE   <- required context, red
```

Nothing was missing. A **required context was failing** — `py3.13` is one of the two
contexts required on `develop` (with `py3.11`; both resolve to `app_id: 15368`, so
protection is correctly wired). `BLOCKED` was the gate doing its job.

Since this PR is docstring-only — no behaviour change, no test change, by your own
description — a red `py3.13` is almost certainly inherited breakage from the base rather
than anything you wrote.

### What I changed

Merged `origin/develop` in (base confirmed `develop`, not `main`; no rebase, force-push
is barred by hook). Clean merge, no conflicts. I did not touch this PR's code — merge
commit only.

That gives a fresh head, so CI re-runs against current `develop`, which should settle
whether the `py3.13` failure was inherited. Note the branch also picks up the min+max PR
gate: the PR matrix is now `["3.11","3.13"]` (3.12 moved to the hosted nightly), so the
next run emits exactly the two required legs and no more.

If `py3.13` is still red after this, it is worth reading as a real signal.

PR #922, opened 2026-08-09
feat(config): let every spec declare its own a2a bind address

ywatanabe1989:
## What

`sac agents declare-a2a-host` — a fleet sweep that makes every spec state its own a2a bind address instead of inheriting one from a code default. **Dry-run is the default.**

101 of the fleet's 102 specs already declare `spec.a2a.host`; exactly one (`scitex-todo`) omits it. Measured with a YAML parser recording full field paths, not grep:

```
total spec.yaml (excluding _* scaffolding): 102   parse errors: 0
  spec.a2a         102        spec.a2a.port    102     <- positive control
  spec.a2a.host    101        spec.host        102
  spec.comms.a2a   101
spec.a2a.host distinct values: {'127.0.0.1': 101}
specs with spec.a2a but NO spec.a2a.host: ['.../agents/scitex-todo/spec.yaml']
```

Live dry-run agrees:

```
102 spec(s) scanned; 1 to write; 101 refused
  101 already declares spec.a2a.host (canary-resume-test, claude-code-telegrammer, ...)
  + scitex-todo: spec.a2a.host: 127.0.0.1

Dry-run - nothing written. Re-run with --apply to write.
```

## Zero behaviour change - the proof

The value written is `127.0.0.1`, the value every reader of `spec.a2a.host` already falls back to. Run end-to-end against a **copy** of the real `scitex-todo` spec (the live registry was not mutated):

```
apply written:      ('scitex-todo',)     refused: None    rolled_back: None
verify summary:     1 spec(s) re-parsed; host declared as expected,
                    documents otherwise identical

EFFECTIVE BIND HOST
  raw-dict readers BEFORE (code default): '127.0.0.1'  hex=3132372e302e302e31
  raw-dict readers AFTER  (from spec):    '127.0.0.1'  hex=3132372e302e302e31
  byte-identical: True
  parse_a2a BEFORE: A2ASpec(host='127.0.0.1', port='auto')
  parse_a2a AFTER:  A2ASpec(host='127.0.0.1', port='auto')

lines added: ['    host: 127.0.0.1']    line count: 200 -> 201
parsed documents identical once the new key is removed: True
```

`raw-dict readers` is the expression `str(a2a.get("host", "127.0.0.1"))`, shared verbatim by `runtimes/a2a_sidecar.py:109` (**the bind** - it becomes `a2a serve --host`), `_lifecycle/health.py:110` and `cli_pkg/a2a_group.py:163` (probe URLs). The real spec has **no trailing newline**, so the terminator-repair path was exercised for real rather than only in a fixture.

There is deliberately **no `--host` option**. An operator-supplied value would make "zero behaviour change" a property of the invocation rather than of the command; rebinding an agent is a different operation and deserves its own PR.

## Anchoring by path, not by regex

Every real spec carries two decoys a bare anchor matches:

| key | present in | why it's a trap |
|---|---|---|
| `spec.host` | 102 / 102 | the **machine placement** (`ywata-note-win`), and it appears *earlier in the file* than the a2a block |
| `spec.comms.a2a` | 101 / 102 | a second `a2a` key at a deeper indent |

`spec.a2a` precedes `spec.comms.a2a` in all 102 files, so a "first match" anchor works *today* - by luck, invisibly, and only until someone reorders a spec. `find_block(bodies, ("spec", "a2a"))` matches each segment at exactly the enclosing block's child indent instead.

## Negative controls

A test that passes with its logic removed is not a test. Both branches were neutered in place, then restored.

**1. Anchor refusal** - replaced the refusal with a guess (insert after the `a2a:` key line):

```
FAILED test_an_a2a_block_without_a_port_line_is_refused
E  assert None == 'spec.a2a has no port: line to anchor to'
E   +  where None = LineEdit(text='spec:\n  a2a:\n    host: 127.0.0.1\n    handler: echo\n',
E                            changed=True, reason=None).reason
1 failed, 19 passed
```

Exactly one test failed, the right one, and the output shows the neutered code producing the guess.

**2. Block scoping** - replaced the scoped idempotence check with a naive file-wide `host:` scan:

```
7 failed, 13 passed
  ... reason='already declares spec.a2a.host'   <- but what matched was spec.host
```

This is the failure mode worth naming: the naive scan reports every spec as already-declared, so the sweep migrates **nothing across all 102 specs while printing a clean run**.

**3. Dry-run exit code** - replaced `SystemExit(0 if plan.safe_to_apply else 1)` with `SystemExit(0)`:

```
FAILED test_a_dry_run_that_found_a_problem_does_not_exit_zero
E  assert 0 != 0
1 failed, 11 passed
```

That contract was stated in a code comment and checked nowhere until this control was run, which is how a contract breaks silently.

## The half-built kit this completes

`_layers_migration_model.py` + `_layers_migration_apply.py` landed in #918/#919 with no middle: `plan_migration` is named in the model's own docstring but does not exist, and `git grep 'MigrationPlan('` finds hits only in tests - no production code could construct a plan. `_maintenance/_spec_sweep_plan.py` is that missing builder, written **once for all spec sweeps**; the per-migration variation is the edit function, so `to_home_layers` can adopt it unchanged rather than growing a second convention.

The shared model and apply are **reused unmodified** - `layers=()` is passed explicitly rather than given a default, so no file another agent may be mid-flight in gets touched.

## Files

| file | role |
|---|---|
| `config/_yaml_line_edit.py` | shared line primitives; `find_block` locates a key by full path |
| `config/_a2a_host_line.py` | the editor - refuses (never guesses) and names which refusal |
| `config/_a2a_defaults.py` | `DEFAULT_A2A_HOST`, and the record of why the five spellings aren't yet collapsed |
| `_maintenance/_spec_sweep_plan.py` | the missing plan builder, generic over the edit function |
| `_maintenance/_a2a_host_sweep.py` | post-write gate: three-valued (wrong / unparsable / drifted) |
| `cli_pkg/_declare_a2a_host.py` | the verb, beside `refresh-acl` |

No ruamel load+dump, for the reason `_to_home_layers_line` and `_group_sync` already record: it reformats quote styles, flow-vs-block, blank lines and comment placement across files nobody asked to touch.

## Verification

- **976 passed** across `tests/.../config/` + `tests/.../_maintenance/`, including every pre-existing `_to_home_layers_line` and `_layers_migration_*` test. Nothing regressed.
- **2440 passed / 2 failed** across the whole of `tests/.../cli_pkg/`. Both failures are `test__dev_jobs.py`, `ModuleNotFoundError: No module named 'scitex_agent_container._jobs_plugin'` — the module lives at `_jobs/_jobs_plugin.py`, so it is a pre-existing path mismatch in a package this branch never touches, and CI does not reproduce it.
- Run with `PYTHONPATH` pointed at this worktree's `src/`, so the tests exercised *these* changes and not the main checkout's code.
- The live fleet registry was **not** mutated. Applying is the operator's deliberate act - that is what dry-run-by-default is for.

One self-correction worth surfacing rather than burying: the first push shipped `test__a2a_host_equivalence.py`, an orphan test file with no `_a2a_host_equivalence.py` beside it in `src/` — a PS-204 §2 violation that turned all three pytest legs red. The tests now live in `test__a2a_host_line.py`, which is where they belonged, since they assert that module's claim. One of the eight was dropped as an exact duplicate of an existing idempotence test (see the comment thread).

## Known, not fixed here

- **The default is spelled five times** (`config/_types.py:212`, `config/_parsers/_a2a.py:24`, `a2a_sidecar.py:109`, `health.py:110`, `a2a_group.py:163`). They agree today, which is the only reason it has gone unnoticed. Collapsing them is the right fix and is blocked: `_types.py` (519 lines) and `a2a_group.py` (559) are both already over the repo's 512-line cap, so every edit to them is refused. A *partial* collapse would imply a canonicality that does not hold, so instead the agreement is **pinned by test** - if one drifts, the zero-behaviour-change claim fails loudly instead of quietly becoming false.
- **`spec.a2a.host` is still unread by two of the three bind paths.** The tui turn bridge (`_tui_turn_bridge_lifecycle.py:45`, hardcoded `DEFAULT_HOST`) and the SDK runner (`_apptainer_inner_argv.py` never emits `--a2a-host`) ignore it; only the sidecar honours it. So the fleet will now uniformly *declare* a bind address that two paths do not read. Threading it through is a **behaviour change** and belongs in its own PR - this one is deliberately inert.
- `_skills/.../17_inbound-turn-endpoint.md:86` claims the runner argv `--a2a-host` is set from `spec.a2a.host`. The argv builder never emits it. Stale doc, untouched here.
- `fleet_spec_paths` duplicates the four-line glob in `cli_pkg/refresh_acl._fleet_spec_paths`. Left alone to keep this PR's blast radius at one modified file; worth collapsing when either is next touched.


ywatanabe1989:
Correcting my own commit message on `d625310a`, which says "No test was dropped or weakened". That is imprecise, and the precise version matters:

**7 of the 8 equivalence tests moved; 1 was dropped as a genuine duplicate.**

The dropped one was `test_a_spec_that_already_declares_it_is_completely_unaffected`:

```python
declared = insert_a2a_host(_BEFORE).text
again = insert_a2a_host(declared)
assert again.text == declared
```

`test__a2a_host_line.py` already contained `test_an_already_declaring_spec_is_returned_byte_identical`, which asserts the identical property against a different fixture:

```python
once = insert_a2a_host(_SPEC).text
twice = insert_a2a_host(once)
assert twice.text == once
```

So the idempotence property is still pinned and no coverage was lost — but "no test was dropped" was the wrong way to say it. Counts: the two config test files went 20 + 8 + 12 = 40 before, 27 + 12 = 39 after.

ywatanabe1989:
CI is green, but the `pytest-matrix-on-ubuntu-py3.12` leg needed three attempts on the identical commit (`e46e3da3`). Recording why, since "re-ran until green" is exactly the move that hides a real defect:

| attempt | outcome | failing test |
|---|---|---|
| 1 | FAILURE | `_state/test_host_config.py::test_host_list_returns_empty_peers_with_no_config` — `json.decoder.JSONDecodeError` |
| 2 | FAILURE | `_listen/test__nodes.py::test_external_inbox_stream_event_preserves_from_agent` — `OSError: [Errno 98] address already in use` binding `127.0.0.1:37895` |
| 3 | SUCCESS | — |

A **different** test failing on each attempt is nondeterminism, not a deterministic defect. Supporting evidence that neither failure can reach this branch's code:

- py3.11 and py3.13 passed on all attempts of the same commit.
- Neither test imports anything this PR touches. `test__nodes.py` imports `_listen.server`, `_runners._session_state`, `_state.registry`, `_state.state_db`; `test_host_config.py` imports `_state.host_config`. This PR adds to `config/`, `_maintenance/`, and `cli_pkg/_declare_a2a_host.py`.
- Attempt 2 is unambiguously infrastructure: a port was already bound by another process on the shared self-hosted runner. Nothing in this PR binds a socket — `declare-a2a-host` only reads and writes YAML text.
- `test_host_config.py` already carries a `_parse_probe_json` helper whose docstring says it exists so "a flaky run produces an actionable diagnostic instead of an opaque `JSONDecodeError`" — i.e. this file's flakiness predates this branch and was known.

The honest residual: adding tests changes how xdist distributes work, which can shift the *timing* that makes a pre-existing port collision surface. That is a latent hazard in tests that bind real ports, not a defect introduced here, and it is worth its own card.

PR #924, opened 2026-08-09
feat(a2a): make spec.a2a.host reach all three bind paths, not one

ywatanabe1989:
## The gap, measured

PR #922 made all 102 fleet specs declare `spec.a2a.host` and disclosed that two
of sac's three a2a bind paths never read it. That is measured here, not
restated: the same driver was run against `origin/develop` and against this
branch, spawning the REAL launcher for each path and reading the bound address
back from the kernel via `lsof` on the resulting process.

**BEFORE (`origin/develop`, 0a255bd9)**

| bind path | spec says `127.0.0.1` | spec says `0.0.0.0` | spec says `192.168.11.121` |
|---|---|---|---|
| `runtimes/a2a_sidecar.py` -> `a2a serve --host` | `127.0.0.1` | `0.0.0.0` | `192.168.11.121` |
| `runtimes/_tui_turn_bridge.py` (TUI `/v1/turn`) | `127.0.0.1` | **`127.0.0.1`** | **`127.0.0.1`** |
| `runtimes/_apptainer_inner_argv.py` (SDK runner) | `127.0.0.1` | **`127.0.0.1`** | **`127.0.0.1`** |

Four cells where the spec and the socket disagree, and nothing anywhere reports
it — `sac agents health` is happy either way. On the SDK row the builder emitted
no `--a2a-host` at all (`<absent>`), so the runner fell back to its own separate
`127.0.0.1` literal.

**AFTER (this branch)**

| bind path | spec says `127.0.0.1` | spec says `0.0.0.0` | spec says `192.168.11.121` |
|---|---|---|---|
| `runtimes/a2a_sidecar.py` -> `a2a serve --host` | `127.0.0.1` | `0.0.0.0` | `192.168.11.121` |
| `runtimes/_tui_turn_bridge.py` (TUI `/v1/turn`) | `127.0.0.1` | `0.0.0.0` | `192.168.11.121` |
| `runtimes/_apptainer_inner_argv.py` (SDK runner) | `127.0.0.1` | `0.0.0.0` | `192.168.11.121` |

Zero mismatches. Column 1 is the **no-regression** case — the value all 102
fleet specs declare today still binds loopback on every path, so nothing moves
until an operator changes a spec.

## What changed

* **`_tui_turn_bridge_lifecycle.resolved_a2a_host(config)`** — reads
  `spec.a2a.host` exactly as `a2a_sidecar.py:109` does, falling back to
  `DEFAULT_HOST`. `start_turn_bridge` / `stop_turn_bridge` now take
  `host: str | None = None` and resolve from the spec when the caller does not
  name one; an explicit `host=` still wins (the test seam). The port-free /
  port-release probes get the same address, so they ask about the socket we are
  actually about to bind. The spawned bridge's own `--host` also falls back to
  the spec, so a hand-run bridge lands where its launcher-run twin would.

* **`_apptainer_inner_argv._a2a_argv`** — emits `--a2a-host` alongside
  `--a2a-port`, for `kind: Agent` and `kind: AgentProxy` alike. The two
  previously-duplicated port/card-yaml blocks are now one helper so they cannot
  drift. From there the value continues along code that already existed:
  `--a2a-host` -> `_session_cli.main` -> `claude_session.run(a2a_host=)` ->
  `_session_http.serve_inbound(host=)` -> `uvicorn.Config(host=)`.

A blank or missing host emits nothing / resolves to the default, so a spec that
declares nothing is byte-identical to before.

## The five-spellings wall: option (b), and why

The `127.0.0.1` default is spelled in five places. Two of them —
`config/_types.py` (519 lines) and `cli_pkg/a2a_group.py` (559) — are over this
repo's 512-line soft cap and refuse edits, so a collapse onto
`DEFAULT_A2A_HOST` would be *partial*, and a partial collapse advertises a
canonicality that does not hold. #922 declined for the same reason and pinned
the agreement by test instead. This PR keeps that choice and **extends the
pinning to the two paths it threads**:

* `_tui_turn_bridge.DEFAULT_HOST == DEFAULT_A2A_HOST`
* the runner's `--a2a-host` flag default `== DEFAULT_A2A_HOST`

Splitting those two oversized modules is a separate concern; doing it inside a
behaviour change would make both harder to review and harder to revert.

(Worth recording: the count is really seven, not five. `_session_cli.py:74`,
`_runners/claude_session.py:183` and `_runners/a2a_proxy.py:313`/`:465` carry
their own copies too. The two new pins cover the ones this PR's paths depend on.)

## Negative control

Each newly-threaded path was neutered in place and its protecting tests
re-run — a test that survives the removal of what it protects is not a test.
Results are in the PR thread.

## Test placement

Cross-module invariants live in the mirror of the module whose function they
CALL (PS-204): the argv assertions are in
`tests/.../runtimes/test__apptainer_inner_argv.py`, the bridge assertions in
`tests/.../runtimes/test__tui_turn_bridge.py`, and the flag-parse assertions in
`tests/.../\_runners/test__session_cli.py`. No orphan test file is added.
The bridge's bind assertions bind a REAL socket on port 0 and read
`server_address` back, rather than echoing the argument passed in.

## Also

`17_inbound-turn-endpoint.md` claimed `runtimes/claude_session.py::start` set
`--a2a-host` from `spec.a2a.host`. No module set that flag at all; the file now
names the real builder and records that the host reaches all three paths.

### DIFF
--- a/src/scitex_agent_container/_skills/scitex-agent-container/17_inbound-turn-endpoint.md
+++ b/src/scitex_agent_container/_skills/scitex-agent-container/17_inbound-turn-endpoint.md
@@ -16,7 +16,8 @@ The endpoint lives in `_runners/_session_http.py`, spawned by the SDK
 runner (`_runners/claude_session.py::run`) when its argv carries
 `--a2a-port N`. The runner — and therefore this HTTP server — runs
 **inside the apptainer container** (`apptainer exec`); the runtime adapter
-sets `--a2a-port` from `spec.a2a.port`. See
+sets `--a2a-port` from `spec.a2a.port` and `--a2a-host` from
+`spec.a2a.host`. See
 [15_claude-session.md](15_claude-session.md) for the container shape.
 
 ## YAML
@@ -83,7 +84,9 @@ non-SDK runtimes (see [07_a2a-protocol.md](07_a2a-protocol.md)).
 
 ## Wiring details
 
-The runner's argv `--a2a-port`/`--a2a-host` is set automatically by `runtimes/claude_session.py::start` from `spec.a2a.port` / `spec.a2a.host`. The handler enqueues a `TurnEnvelope` on the runner's `asyncio.Queue` and awaits `env.response`; the conversation task drains it, calls `client.query(text)`, drains `receive_response()`, and resolves the future.
+The runner's argv `--a2a-port`/`--a2a-host` is built by `runtimes/_apptainer_inner_argv.py::_a2a_argv` from `spec.a2a.port` / `spec.a2a.host` (this file previously named `runtimes/claude_session.py::start`, which never built either flag; `--a2a-host` was not emitted at all until the bind-path threading change). The host continues `--a2a-host` → `_session_cli.main` → `claude_session.run(a2a_host=)` → `_session_http.serve_inbound(host=)` → `uvicorn.Config(host=)`. The handler enqueues a `TurnEnvelope` on the runner's `asyncio.Queue` and awaits `env.response`; the conversation task drains it, calls `client.query(text)`, drains `receive_response()`, and resolves the future.
+
+`spec.a2a.host` now reaches all three of sac's a2a bind paths — the sidecar (`runtimes/a2a_sidecar.py` → `a2a serve --host`), the TUI turn bridge (`runtimes/_tui_turn_bridge.py`), and this SDK runner. Every fleet spec declares `127.0.0.1`, so all three bind loopback today; changing the declared value now moves all three together instead of one.
 
 ## Implementation files
 
--- a/src/scitex_agent_container/runtimes/_apptainer_inner_argv.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_inner_argv.py
@@ -200,6 +200,51 @@ def _resolve_restart_backoff_s(config: "AgentConfig") -> float:
     return 1.0
 
 
+def _a2a_argv(config: "AgentConfig") -> list[str]:
+    """``spec.a2a`` -> the session runner's ``--a2a-*`` flags.
+
+    Shared by both ``kind`` branches so the two can never drift apart.
+
+    ``--a2a-port`` is the port the runner's inbound-turn server listens on;
+    without it the sidecar never binds and POST /v1/turn is unreachable.
+    ``--a2a-host`` is that server's BIND ADDRESS, threaded from
+    ``spec.a2a.host``. Until this was added the builder emitted only the port,
+    so the runner fell back to its own ``--a2a-host`` default and the address
+    the spec declared reached exactly one of the three bind paths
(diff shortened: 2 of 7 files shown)


ywatanabe1989:
## Negative control — per path

Each newly-threaded path was reverted **in place** to the hardcoded default and
its protecting tests re-run. A test that stays green when you delete what it
protects is decoration, so this is the check that decides whether these tests
are real.

### Path 2 — `_tui_turn_bridge`

Neuter: `resolved_a2a_host()` body replaced with `return DEFAULT_HOST`.

```
FAILED test__tui_turn_bridge.py::test_resolved_a2a_host_returns_the_declared_spec_host
        AssertionError: assert '127.0.0.1' == '192.168.11.23'
FAILED test__tui_turn_bridge.py::test_bridge_binds_the_wildcard_address_the_spec_declares
        AssertionError: assert '127.0.0.1' == '0.0.0.0'
FAILED test__tui_turn_bridge.py::test_start_turn_bridge_spawns_with_the_declared_host
        AssertionError: assert '127.0.0.1' == '0.0.0.0'
3 failed, 8 passed, 29 deselected
```

The 8 that stayed green are the ones that *should*: the loopback/undeclared
cases, the blank-host fallback, the `DEFAULT_HOST == DEFAULT_A2A_HOST` pin and
the explicit-override seam all describe behaviour the neuter does not change.

### Path 3 — the SDK argv builder

Neuter: the `--a2a-host` emission removed from `_a2a_argv`.

```
FAILED test__apptainer_inner_argv.py::test_agent_runner_argv_emits_a2a_host_alongside_the_port
FAILED test__apptainer_inner_argv.py::test_agent_runner_argv_a2a_host_is_loopback_for_an_undeclared_host
FAILED test__apptainer_inner_argv.py::test_agent_runner_argv_a2a_host_follows_a_wildcard_spec_host
FAILED test__apptainer_inner_argv.py::test_agent_runner_argv_a2a_host_follows_a_lan_spec_host
FAILED test__proxy_runner_argv_a2a_host_is_loopback_for_an_undeclared_host
FAILED test__apptainer_inner_argv.py::test_proxy_runner_argv_a2a_host_follows_the_declared_spec_host
FAILED test__apptainer_inner_argv.py::test_build_inner_argv_carries_the_declared_a2a_host_into_the_exec_line
7 failed, 2 passed, 45 deselected
```

The 2 that stayed green are the two absence assertions (no resolved port, blank
declared host) — again, correctly indifferent to the neuter.

Both neuters were reverted; `git diff --stat` is byte-identical before and
after, and no `NEGATIVE CONTROL` marker survives in `src/`.

### Whole-system control

Independently of the unit tests, the six-cell bind table in the PR body was
produced by running the same driver against a `git archive origin/develop`
export of the pre-change tree. It reproduces the gap from the outside: the
sidecar followed the spec in all three cases, while the bridge and the SDK
runner bound `127.0.0.1` regardless — four mismatched cells. On this branch the
same driver reports zero.

PR #925, opened 2026-08-09
fix(a2a): key the sidecar pid/log per agent, so one workdir stops meaning one sidecar

ywatanabe1989:
## The defect

`runtimes/a2a_sidecar.py` derived its pid and log paths from `spec.workdir` alone — `{workdir}/a2a-sidecar.{pid,log}`, with no agent identity anywhere in the path. Two agents sharing a workdir therefore shared one PID file, and the second `start_sidecar` logged `already running; skipping` and returned the **first** agent's PID. The second agent then reported healthy while owning no sidecar at all and being undriveable over A2A.

Reported by the agent that landed #924, which hit it as a real false negative in its own test run. It is latent today only because each agent owns its workdir — cloning, twinning and relocating all break that assumption, which is exactly the work in flight.

## What changed

`src/scitex_agent_container/runtimes/a2a_sidecar.py`

| line | what |
|---|---|
| `70-96` | new `_state_dir(config)` — resolves the agent's own runtime state dir, **raises** on an empty `config.name` |
| `99-104` | `_pid_path` / `_log_path` now key off `_state_dir` instead of `expanded_workdir` |
| `107-109` | `_legacy_pid_path` — the pre-keying workdir path, read-only compatibility |
| `149-155` | `_proc_argv` — the kernel's record of a pid's argv |
| `158-183` | `_legacy_sidecar_pid` — legacy pid **iff** `/proc/<pid>/cmdline` says it is ours |
| `186-216` | `_adopt_legacy_sidecar` — take over a live pre-keying sidecar, move the record |
| `246-251` | `start_sidecar` consults the legacy path after the per-agent one |
| `327-353` | `_terminate_recorded` / `_stop_legacy_sidecar` |
| `356-380` | `stop_sidecar` drains the legacy path too |

Also: `_skills/scitex-agent-container/07_a2a-protocol.md` (documented paths + migration), and two stale `a2a_sidecar.py:109` line references in `tests/.../config/test__a2a_host_line.py`.

### Directory, not filename

`<runtime-root>/<agent>/a2a-sidecar.{pid,log}` — resolved through `tui_session.state_dir_for_config`, the SSOT both host runtimes already use, i.e. the same `runtime/<agent-id>/` that holds `pid`, `heartbeat.json`, `session.jsonl` and `runner.log`.

The identity goes in the **directory** because that is what every other per-agent artefact does, and the surrounding machinery is built around it:

* `_lifecycle/_rename.py` renames `runtime/<name>/` as a unit — a flat `a2a-sidecar-<agent>.pid` sitting in a workdir would be missed;
* `$SCITEX_AGENT_CONTAINER_RUNTIME_DIR` (the GPFS inode fix) relocates the tree as a unit, and now covers sidecar state too;
* `state_dir_for_config` keeps a **project-scoped** agent's state under its own repo, so two same-named agents in two different repos do not collide — a filename-keyed scheme reintroduces exactly that class of collision while fixing another;
* it stops writing runtime junk into what is usually a git checkout.

### Where the agent id comes from

`config.name`. It is a required (non-defaulted) `AgentConfig` field sourced from the spec's `metadata.name`, and it is the same key the registry, `state.db` and `screen_name` use, so it is always populated at sidecar-start time.

Rejected:

* **`SAC_NAME`** — injected *into the container* (`runtimes/_apptainer_listen_env.py:84`, `hooks.py:115`). `start_sidecar` runs on the **host**, inside `sac agents start`, where it is unset; worse, when one agent starts another it carries the **calling** agent's name, which is the same collision class in a new costume.
* **`SCITEX_AGENT_CONTAINER_AGENT`** — same story: an in-container identity fallback (`_never_stop_when_task_remains/_identity.py:43`), not a host-side fact.

An empty name **raises** rather than falling back to a shared path. A silent fallback there would reproduce the exact bug this fixes. Both call sites (`_runners/_tmux/claude_code.py:602-605, 626-629`) already log-and-continue around this module, so refusing costs at most the sidecar, never the agent.

## What happens to an agent running RIGHT NOW

Nothing, until its next stop or start — the sidecar is spawned at agent start, so landing this changes no running process.

On that next lifecycle event, a live pre-keying sidecar is still recorded only at `{workdir}/a2a-sidecar.pid` and is still holding the port. Ignoring it would make the fresh spawn fail to bind and the agent would come up sidecar-less. So both entry points consult the legacy path:

* **`start_sidecar` adopts it** — writes the pid to the per-agent path, unlinks the legacy file, returns that pid. No kill, no port churn, no downtime; the running sidecar simply becomes correctly bookkept.
* **`stop_sidecar` terminates it** — for the agent that gets stopped rather than restarted, so its sidecar does not outlive it holding the port.

Both are gated on **ownership decided by the kernel**: `/proc/<pid>/cmdline` must contain *this* agent's `config_path`. Two agents sharing a workdir necessarily have distinct spec files, so a co-tenant's pre-keying sidecar is never adopted and never killed — it keeps running and keeps serving, and the co-tenant adopts it on its own next start. A legacy file we cannot prove is ours is left untouched.

## Evidence — two agents, one workdir

Same script both sides; pids read back from the **kernel** (`lsof -t -i @127.0.0.1:<port> -sTCP:LISTEN`) and from `/proc/<pid>/cmdline`, never from `start_sidecar`'s return value.

**BEFORE** — `git archive origin/develop`, `PYTHONPATH` at the extracted tree:

```
shared workdir : /uvwork/tmp/twoagents-uxh5ipmy/SHARED-WORKDIR
alpha port     : 59224
bravo port     : 60350

start_sidecar(alpha) returned : 379599
start_sidecar(bravo) returned : 379599          <-- bravo handed alpha's pid

KERNEL pid LISTENing on alpha's port 59224 : 379599
   argv: ... a2a serve /uvwork/tmp/twoagents-uxh5ipmy/alpha.yaml --port 59224 ...
KERNEL pid LISTENing on bravo's port 60350 : None    <-- bravo never served

pid files on disk:
   /uvwork/tmp/twoagents-uxh5ipmy/SHARED-WORKDIR/a2a-sidecar.pid  ->  379599

   distinct pid FILES   : 1
   distinct KERNEL pids : 1
   both ports served    : False
   COLLISION            : True
```

**AFTER** — this branch:

```
shared workdir : /uvwork/tmp/twoagents-bll45uum/SHARED-WORKDIR
alpha port     : 58578
bravo port     : 58742

start_sidecar(alpha) returned : 391653
start_sidecar(bravo) returned : 391677

KERNEL pid LISTENing on alpha's port 58578 : 391653
   argv: ... a2a serve /uvwork/tmp/twoagents-bll45uum/alpha.yaml --port 58578 ...
KERNEL pid LISTENing on bravo's port 58742 : 391677
   argv: ... a2a serve /uvwork/tmp/twoagents-bll45uum/bravo.yaml --port 58742 ...

pid files on disk:
   /uvwork/tmp/twoagents-bll45uum/runtime-root/alpha/a2a-sidecar.pid  ->  391653
   /uvwork/tmp/twoagents-bll45uum/runtime-root/bravo/a2a-sidecar.pid  ->  391677

   distinct pid FILES   : 2
   distinct KERNEL pids : 2
   both ports served    : True
   COLLISION            : False
```

Two distinct sidecars, two distinct pids, two distinct pid files — and each kernel argv names its own spec and its own port.

## Negative control

Reverted the keying in place (`_pid_path` / `_log_path` back to `Path(config.expanded_workdir) / …`), same tests:

```
8 failed, 21 passed in 67.45s      <-- reverted
29 passed in ~66s                  <-- restored
```

The eight that fail are exactly the collision-catching ones:

```
test_pid_path_differs_per_agent_in_one_shared_workdir
test_log_path_differs_per_agent_in_one_shared_workdir
test_pid_path_is_keyed_by_the_agent_name_directory
test_pid_path_refuses_an_unnamed_agent
test_second_agent_in_a_shared_workdir_gets_its_own_pid
test_second_agent_in_a_shared_workdir_gets_its_own_pid_file
test_second_agents_recorded_pid_is_a_process_serving_its_own_port
test_start_sidecar_ignores_a_legacy_pid_file_owned_by_another_agent
```

with, e.g. — the reported bug reproduced by the suite:

```
test_second_agent_in_a_shared_workdir_gets_its_own_pid_file
    assert len(recorded) == 2
E   AssertionError: assert 1 == 2
E    +  where 1 = len({'489129'})
```

## Tests

29 in `tests/scitex_agent_container/runtimes/test_a2a_sidecar.py` (was 19), all real: real pid files, real `Popen`, real socket binds, real `os.kill`, `/proc` read for argv. No mocks, no monkeypatch, one assert each.

Wider run — `runtimes/` + `config/` + `_lifecycle/test_health.py`: **932 passed, 1 failed**. The one failure is `test__tui_turn_bridge.py::test_stop_turn_bridge_force_frees_survivor_on_own_port`, a load flake in a file this PR does not touch (`python -c pass` exceeding a 5 s `wait`); it passes on its own in 4.42 s.

https://claude.ai/code/session_01BWwZWke4rh2Civ11ybUw4k

### DIFF
--- a/src/scitex_agent_container/_skills/scitex-agent-container/07_a2a-protocol.md
+++ b/src/scitex_agent_container/_skills/scitex-agent-container/07_a2a-protocol.md
@@ -32,7 +32,9 @@ sac a2a doctor <agent.yaml>    [--host H] [--port N] [--timeout 5.0] [--json]
 
 ## Auto-launch via `spec.a2a`
 
-When a v3 YAML declares `spec.a2a.port`, `sac agents start` spawns the A2A server as a sidecar subprocess after the multiplexer is up. PID lives at `{workdir}/a2a-sidecar.pid`, output at `{workdir}/a2a-sidecar.log`; `sac agents stop` SIGTERMs it via the PID file.
+When a v3 YAML declares `spec.a2a.port`, `sac agents start` spawns the A2A server as a sidecar subprocess after the multiplexer is up. PID lives at `<runtime-root>/<agent>/a2a-sidecar.pid`, output at `<runtime-root>/<agent>/a2a-sidecar.log` — the agent's OWN runtime state dir, next to its `pid` / `heartbeat.json` / `session.jsonl`. `sac agents stop` SIGTERMs it via the PID file.
+
+Both files used to sit at `{workdir}/a2a-sidecar.*`, with no agent identity in the path: two agents sharing one workdir shared one PID file, and the second `start` logged "already running; skipping" and returned the FIRST agent's PID — looking healthy while having no sidecar of its own. Keying on the agent directory removes that collision by construction. A sidecar started before this change is still recorded at the legacy workdir path; `start` adopts it (and moves the record) and `stop` terminates it, but only when `/proc/<pid>/cmdline` confirms the process is serving THIS agent's spec — a co-tenant's sidecar is never touched.
 
 ```yaml
 apiVersion: scitex-agent-container/v3
--- a/src/scitex_agent_container/runtimes/a2a_sidecar.py
+++ b/src/scitex_agent_container/runtimes/a2a_sidecar.py
@@ -11,8 +11,9 @@ Lifecycle:
 
 * :func:`start_sidecar` — called from ``ClaudeCodeRuntime.start`` after
   the multiplexer is up. ``Popen``-spawns the server, captures stdout/
-  stderr to ``{workdir}/a2a-sidecar.log``, writes the PID to
-  ``{workdir}/a2a-sidecar.pid``. No-op if ``spec.a2a.port`` is unset.
+  stderr to ``<runtime>/<agent>/a2a-sidecar.log``, writes the PID to
+  ``<runtime>/<agent>/a2a-sidecar.pid``. No-op if ``spec.a2a.port`` is
+  unset.
 * :func:`stop_sidecar` — called from ``ClaudeCodeRuntime.stop`` (and
   ``cleanup``). Reads the PID file, sends SIGTERM, removes the file.
   No-op if the file is absent.
@@ -20,6 +21,30 @@ Lifecycle:
 Errors are logged and swallowed — a failed sidecar must NOT block
 agent start/stop. The PID file is the source of truth; if the file
 is gone the sidecar is considered stopped.
+
+**Per-agent keying.** The pid/log pair used to live at
+``{workdir}/a2a-sidecar.{pid,log}``, with NO agent identity in the
+path. Two agents sharing one workdir therefore shared one pid file,
+and the second :func:`start_sidecar` logged "already running;
+skipping" and handed back the *other* agent's PID — a silent
+false-positive: the second agent looked healthy while having no
+sidecar of its own and being undriveable over A2A. That is latent only
+while every agent owns its workdir; cloning / twinning / relocating
+breaks the assumption. The files now live in the agent's own runtime
+state dir (``state_dir_for_config`` — the same
+``<runtime-root>/<agent>/`` that already holds ``pid``,
+``heartbeat.json``, ``session.jsonl`` and ``runner.log``), so the
+identity is in the DIRECTORY, exactly like every other per-agent
+artefact. A config with no ``name`` REFUSES to resolve a path rather
+than falling back to a shared one.
+
+**Migrating a sidecar started before that change.** A live pre-keying
+sidecar is still recorded at ``{workdir}/a2a-sidecar.pid`` and still
+holds the port; ignoring it would make the next spawn fail to bind.
+Both entry points therefore consult the legacy path as well — but only
+adopt/kill a process the kernel confirms is *ours*
+(``/proc/<pid>/cmdline`` names THIS agent's spec file). A stranger's
+pre-keying sidecar is left strictly alone.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


ywatanabe1989:
CI: 9/9 SUCCESS — same 9-name leg set as the merged control #924, zero pending, zero skipped.

First attempt had `pytest-matrix-on-ubuntu-py3.11` red on `tests/scitex_agent_container/_listen/test__node_channel_keepalive.py::test_idle_inbox_stream_emits_keepalive_frames`:

```
OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 42153): address already in use
RuntimeError: uvicorn loopback on port 42153 CRASHED during startup after 0.88s
```

Not this change. That file picks its port with the claim-and-release pattern (`test__node_channel_keepalive.py:74-83` — bind port 0, read `getsockname()`, close, rebind later), which is a TOCTOU any concurrent binder can win, and three python legs run concurrently on the shared self-hosted runner. py3.12 and py3.13 passed the same commit on the first attempt; a `--failed` rerun of py3.11 went green with no code change. Nothing in this PR touches `_listen/` or allocates a port.

Worth naming honestly: the new tests do spawn six additional short-lived real `a2a serve` processes per run (three `shared_workdir_pair` instances), so they add some ephemeral-port pressure to a suite that already had this race. The race lives in the other test's port handling rather than in the number of neighbours — but if that keepalive test starts flaking more often, this is a plausible contributor, and binding-then-handing-off the socket (or `SO_REUSEADDR`) is the fix on its side.

PR #928, opened 2026-08-09
fix(runtimes): forward GITHUB_TOKEN into agent containers so gh can authenticate

ywatanabe1989:
## The defect

The GitHub token lives in the operator's dotfiles secrets, which **only a login shell sources**. sac starts agent containers without a login shell, and nothing forwarded the token — so `gh` inside had neither an env token nor a `~/.config/gh`, and reported *"not logged into any GitHub hosts."*

Agents could commit and push but **could not open a pull request**. That is half a delivery loop, and it needs a human every time.

Measured on scitex-compute-04, 2026-08-09 (lengths only — values never read):

| probe | result |
|---|---|
| login shell | `GITHUB_TOKEN` **len=40** — the secret is present |
| non-login shell | `GITHUB_TOKEN` **len=0** — it disappears here |
| spec grep for `GITHUB_TOKEN`/`GH_TOKEN` | **0 matches** |
| `~/.config/gh/` | empty, no `hosts.yml` |
| `gh` on the host | `command not found` — it exists only inside the SIF |

Three agents each lost work time to this, and two finished, tested fixes (#926, #927) had to be opened by a different session on their authors' behalf.

## The change

`_github_token` resolves from the **same pool as `CCT_BOT_TOKEN`** (`SAC_SECRETS_ENVRC`, with the canonical `$HOME` fallback) rather than adding a second secrets path — one place to rotate, one place to reason about. It mirrors that module's loud-warning shape:

- An absent token warns **at start**, naming what will fail (`gh pr create`) and how to fix it — instead of letting an agent discover it only after finishing work it cannot deliver.
- A **defined-but-empty** token is treated as absent. That is exactly the state the containers were in; forwarding it would reproduce the bug while looking like a fix.
- Both `GITHUB_TOKEN` and `GH_TOKEN` are set. `gh` prefers the latter, git credential helpers and most CI read the former; setting one is a half-fix that works for one tool and mystifies the next.

## Security

The value is emitted as `--env`, which is safe **only** because `redact_secret_env_to_file` lifts secret-shaped pairs out of the world-readable argv (a tmux `bash -c` pane cmd; `/proc/<pid>/cmdline` leaks it to any local process) into a per-agent 0600 env-file before exec — and its predicate matches `*_TOKEN`.

`TestRedactionCoupling` pins that relationship. If the predicate ever narrows so these names stop matching, the token would silently start leaking with no other symptom, and this test is what catches it.

The value is never logged — only its length, which keeps a *wrong* token diagnosable without exposing it.

## The extraction

`_apptainer_build_argv` was at **510 of its 512-line budget**, so the feature did not fit without genuine extraction. The quota-cache block moved verbatim to `_apptainer_quota_env`, joining the existing `_apptainer_*` env-wiring family. Behaviour unchanged; net **509**.

This is called out because "drive-by refactor in a feature PR" is usually a smell — here the line budget made it a precondition, not a preference.

## Tests

- 15 new in `test__github_token.py`.
- 109 passed / 2 skipped across `build_argv` + `secret_env` + this module.
- **Negative control:** reverting the blank-token guard fails 5 tests; restoring passes 15. The tests bite rather than passing vacuously.
- Run with `PYTHONPATH` pointed at the worktree src, so they exercise *this* branch's code and not the main checkout's.

## Not in scope

The git **push** identity is still scitex-hub's read-only deploy key, so pushes need an explicit `git@gh-ywatanabe:…` URL. That is a privilege decision for the operator (per-repo write deploy keys / a fleet machine-user key / the personal key as default) and is tracked separately in `sac-fleet-git-push-blocked-wrong-repo-readonly-deploy-key-20260809`.

Card: `sac-github-token-never-reaches-agent-containers-20260809`

### DIFF
--- a/src/scitex_agent_container/runtimes/_apptainer_build_argv.py
+++ b/src/scitex_agent_container/runtimes/_apptainer_build_argv.py
@@ -149,18 +149,10 @@ def build_run_argv(
     if env_file.is_file():
         argv += ["--env-file", str(env_file)]
 
-    # Quota-cache bind (#16) — see module-level docstring in
-    # _apptainer_runtime. Bind read-only + advertise the in-container
-    # path to the telegrammer bridge so its default-path lookup hits
-    # the bind without any per-agent spec change.
-    quota_src = _resolve_quota_cache_host_path()
-    if quota_src.is_file():
-        argv += [
-            "--bind",
-            f"{quota_src}:{QUOTA_CACHE_CONTAINER_PATH}:ro",
-            "--env",
-            f"CCT_QUOTA_CACHE_PATH={QUOTA_CACHE_CONTAINER_PATH}",
-        ]
+    # Quota-cache bind (#16) — extracted to _apptainer_quota_env.
+    from ._apptainer_quota_env import quota_cache_flags
+
+    argv += quota_cache_flags()
 
     # Host access + working directory are the SOLE responsibility of the
     # explicit ``apptainer.binds`` + ``spec.workdir`` — there is NO
@@ -310,6 +302,13 @@ def build_run_argv(
     from ._apptainer_listen_env import listen_env_flags
 
     argv += listen_env_flags(config)
+
+    # GitHub token for the in-container ``gh``. Safe as ``--env`` ONLY because
+    # redact_secret_env_to_file below lifts ``*_TOKEN`` out of the argv. Why
+    # this is needed at all: see _github_token.
+    from ._github_token import github_token_env_flags
+
+    argv += github_token_env_flags(agent_name=config.name)
 
     # TUI parity with the SDK's telegrammer wake (apply_channels →
     # _wire_telegrammer_wake): inject CLAUDE_CODE_TELEGRAMMER_TURN_URL so an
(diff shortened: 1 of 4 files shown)


ywatanabe1989:
Closing in favour of #929, which fixes the same defect better.

We built this twice — I on the laptop, the compute-04 agent on its own host, from the same card, without either of us seeing the other's branch until both were pushed. That duplication is a real coordination failure and is worth naming rather than quietly deleting one branch.

**Why #929 wins, and it is not close:**

This PR forwards the token as `--env` and relies on `redact_secret_env_to_file` lifting it out of the world-readable argv into a 0600 file. That works, and I added `TestRedactionCoupling` to pin the dependency — but it is still a dependency. If that predicate ever narrows so `*_TOKEN` stops matching, the token silently starts leaking into `/proc/<pid>/cmdline` with no other symptom. A test that guards a coupling is strictly worse than not having the coupling.

#929 writes into `dest/.env` via `deploy_to_home`, so the value **never enters argv at all**. Nothing to redact, no window, no invariant to maintain. It is also the path `CCT_BOT_TOKEN` already takes, so the codebase ends up with one secrets mechanism instead of two — which is what the card actually asked for and what I wrote in the card myself before implementing the other thing.

**What is lost by closing this, and what is not:**

- The `_apptainer_quota_env` extraction here was a *precondition of this approach only*: `_apptainer_build_argv` sat at 510 of its 512-line budget, so my change did not fit without it. #929 does not touch that file, so the extraction is unnecessary rather than merely unmerged. Not carrying it over.
- The blank-token-is-absent rule and the both-spellings rule exist in #929 too, independently derived, with its own tests. Nothing behavioural is lost.

Reviewers: read #929, not this.

PR #942, opened 2026-08-09
fix(acl): say WHICH kind of 'no group' a denial found (ungrouped vs no policy row vs no caller)

ywatanabe1989:
## The bug

`resolve_group_name` returned `""` for **three** different situations, and the `host_exec` ACL denial built on it said only:

```
caller 'alice' resolves to group ''
```

| state | what it means | what the operator must do |
|---|---|---|
| `no_caller` | nothing was looked up | — |
| `ungrouped` | a policy row **exists**, `group_name` empty | label the agent |
| `no_policy_row` | **no row at all** | check the DB path |

`no_policy_row` is the dangerous one. A wrong `db_path` presents **every** agent as ungrouped, so an operator reading `group ''` can spend an entire investigation labelling agents that were already labelled — in the database they meant to be reading.

## Where the collapse comes from

Not a bug in the resolver. `read_comms_policy` **documents** it:

> *"A missing row yields `DEFAULT_COMMS_POLICY` so the 'no-row' vs 'row-with-default-values' distinction is invisible to callers."*

That is correct for every caller that wants defaults, and fatal for the one caller that reports to a human. So this **adds a second reader** rather than widening that contract:

```
resolve_group_name -> str              membership tests, UNCHANGED
resolve_group      -> GroupResolution  group + provenance, for humans
```

`GroupResolution` is frozen and **validated**: a `named` source with an empty `group_name`, or a `group_name` on any other source, raises at construction rather than three layers downstream. `describe()` names the state **and** the remedy.

## Blast radius kept small on purpose

`host_exec` keeps its **decision** on `group_resolver` (a plain string compare that the seam tests inject through). Only the **explanation** is enriched, via a separate `group_detail` parameter defaulting to `resolve_group`. A test injecting `group_resolver` alone is unaffected, and the detail lookup is wrapped so diagnostics can never break a deny.

## A forced refactor came with it

`state_db_nodes.py` was **already 587 lines against a 512 budget before this change** — the line-limit hook blocked every edit to it, including a one-line import. Extracted the named-group axis (`resolve_group_name`, `same_named_group`, `is_developer`, `is_researcher`) into `_state/state_db_groups.py`, re-exported from the original so **no call site changes**.

```
state_db_nodes.py   587 -> 412
state_db_groups.py  new
```

## Tests

- **15 new**, covering all four states, both `describe()` remedies, and the validator rejecting malformed resolutions at construction.
- **2128 passed / 3 skipped** across `_state` + `_listen` on py3.11, confirming the extraction is behaviour-neutral.

One test asserts the **old** resolver *still* collapses `ungrouped` and absent. If that ever changes, the reason this module exists gets re-examined rather than silently invalidated.

### DIFF
--- a/src/scitex_agent_container/_listen/_host_exec.py
+++ b/src/scitex_agent_container/_listen/_host_exec.py
@@ -108,6 +108,7 @@ from starlette.requests import Request
 from starlette.responses import JSONResponse
 
 from .._lifecycle._off_loop import run_blocking
+from .._state.state_db_groups import resolve_group
 from ._acl import deny_response, resolve_group_name
 from ._host_exec_child import (
     _POST_KILL_DRAIN_S,
@@ -188,6 +189,9 @@ async def host_exec(
     request: Request,
     *,
     group_resolver=resolve_group_name,
+    # Diagnostics only — never consulted for the allow/deny decision, so a
+    # test that injects `group_resolver` alone keeps its existing behaviour.
+    group_detail=resolve_group,
     audit_writer=_append_audit,
 ) -> JSONResponse:
     """``POST /v1/host_exec`` — see module docstring for the full contract.
@@ -277,10 +281,20 @@ async def host_exec(
 
     group = group_resolver(name=caller)
     if group not in ELIGIBLE_GROUPS:
+        # The DECISION stays with group_resolver (a plain string compare, and
+        # the seam tests inject through). Only the EXPLANATION is enriched:
+        # `group ''` is three different situations wearing one face, and the
+        # operator reading this denial has to pick between "label the agent"
+        # and "you are looking at the wrong database". Saying which costs one
+        # query on a path that is already failing.
+        try:
+            detail = group_detail(name=caller).describe()
+        except Exception:  # noqa: BLE001 - never let diagnostics break a deny
+            detail = f"group {group!r}"
         return deny_response(
             reason=(
                 f"host_exec is restricted to groups {sorted(ELIGIBLE_GROUPS)}; "
-                f"caller {caller!r} resolves to group {group!r}"
+                f"caller {caller!r} has {detail}"
             )
         )
 
--- a/src/scitex_agent_container/_state/state_db_groups.py
+++ b/src/scitex_agent_container/_state/state_db_groups.py
@@ -1,0 +1,222 @@
+"""NAMED-GROUP membership — the second grouping axis, and its provenance.
+
+Extracted from :mod:`.state_db_nodes` (587 lines, over the 512 budget) as one
+cohesive responsibility. Named groups are explicitly "a SECOND grouping axis
+layered on top of the lineage-derived group mesh", which makes them separable
+from lineage recording, spawn authorisation and host resolution.
+
+Two resolvers, and the difference between them is the point of this module:
+
+* :func:`resolve_group_name` returns a bare string. Correct for MEMBERSHIP
+  TESTS, which genuinely do not care why a group is absent.
+* :func:`resolve_group` returns :class:`GroupResolution` — the group AND the
+  provenance of that answer. Correct for anything a HUMAN will read.
+
+WHY THE SECOND ONE EXISTS. ``resolve_group_name`` returns ``""`` for three
+different situations, and an ACL denial built on it said only::
+
+    caller 'alice' resolves to group ''
+
+which does not tell an operator whether to label the agent or to go looking
+for the right database. The states need different actions:
+
+    ungrouped       a policy row exists, group_name empty  -> label the agent
+    no_policy_row   no row at all                          -> check the DB path
+    no_caller       nothing was looked up
+
+The collapse originates one layer down and is deliberate there:
+``read_comms_policy`` documents that "a missing row yields
+DEFAULT_COMMS_POLICY so the 'no-row' vs 'row-with-default-values'
+distinction is invisible to callers". That is right for every caller that
+wants defaults — and is exactly what has to be undone when the answer is
+going to be shown to a person. So :func:`resolve_group` queries the table
+directly rather than reaching through that helper.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from .state_db_acl_policy import read_comms_policy
+
+#: How a group resolution came about. FOUR states, deliberately, because they
+#: call for different operator actions and a bare ``""`` made them one.
+GROUP_SOURCES = ("named", "ungrouped", "no_policy_row", "no_caller")
+
+
+@dataclass(frozen=True)
+class GroupResolution:
+    """A named group PLUS why it came out that way.
+
+    ``group_name`` is non-empty only when ``source == "named"``; the
+    validator enforces that so a malformed answer fails where it is built
+    rather than three layers downstream.
+    """
+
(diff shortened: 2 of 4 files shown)


ywatanabe1989:

ywatanabe1989:
**This PR is `DIRTY`, so it cannot run CI at all** — the one green `CLAssistant` and zero test jobs read like "queued behind a busy pool" and are not. I was sweeping the four conflicted PRs and unstuck the other three, but **I am deliberately leaving this one alone**: resolving it requires an ACL semantics decision that belongs to you, not to a conflict sweep. The branch is untouched (still `4361a3a`); I merged, read the conflict, and aborted.

Here is the conflict, so you can decide quickly.

**`src/scitex_agent_container/_state/state_db_groups.py` is an add/add of two different modules that happen to share a filename.** Both were extracted from `state_db_nodes.py` to get under the 512-line cap, independently, for different reasons:

* **This branch** (222 lines): *"NAMED-GROUP membership — the second grouping axis, and its provenance"* — adds `GroupResolution` + `resolve_group`, the four-state provenance that is the point of this PR.
* **`develop`** (272 lines): *"MULTI-value named-group readers — the AUTHORITY half of the group ACL"* — adds `resolve_group_names`, `in_named_group`, `is_privileged`.

They collide on four function names — `resolve_group_name`, `same_named_group`, `is_developer`, `is_researcher` — and the two `is_developer`s are **not** the same function:

```python
# this branch — single primary group
return is_developer_group(resolve_group_name(name=name, db_path=db_path))

# develop — membership over the FULL set
return in_named_group(name=name, group=DEVELOPER_GROUP, db_path=db_path)
```

`develop`'s docstring says why: *"MEMBERSHIP, not primary-group equality (incident 2026-08-10): an agent whose spec says `groups: [generalist, developer]` is a developer. It previously was not, because only the first element reached the DB."*

So the resolution a conflict sweep would naturally reach for — keep the PR's own code — **would re-introduce the 2026-08-10 `grant` authority incident**, where an agent's spawn / manage / host-exec authority depended on the order of a YAML list. That is a live ACL gate, and picking a winner there is not mine to do.

There are two further conflicts in exactly the call sites of those predicates — `_state/state_db_nodes.py` (1 region) and `_listen/_host_exec.py` (3 regions) — which will follow whatever you decide above.

My read of the shape of the fix, offered only as a starting point: `develop`'s multi-value authority predicates should survive as-is, and this PR's `GroupResolution` / `resolve_group` provenance work should be layered on top of them rather than beside them. Note that the union is ~490 lines before any glue, against the 512 cap both modules cite as their reason to exist, so a further split is likely part of the answer — which is more evidence this wants its author rather than a merge sweep.

Once you have merged `origin/develop` in (a merge, not a rebase — force-push is barred by hook), `mergeStateStatus` should go `DIRTY → BLOCKED` and the check count 1 → 8. That transition is the proof, not the merge succeeding.

ywatanabe1989:
Closing: `develop` already makes this exact distinction, and merging this now would **undo** two ACL fixes that landed after it.

This PR has been `CONFLICTING` since it was opened, so GitHub could never build a merge ref and **neither required context ever ran** — not `pytest-matrix-on-ubuntu-py3.11`, not `pytest-matrix-on-ubuntu-py3.13`. The single `cla` check made it read as "1 passed / 0 failed". Absent is not pending; zero failures is not green.

## The fix landed by another route, one day later

`db70eda0` — *"fix(acl): a host_exec denial names WHICH empty-group case, and where it looked (#932)"*, merged 2026-08-10 — solves the same incident (2026-08-09, three agents refused in fifteen minutes) at the same call site. `develop`'s `_host_exec.py` now branches on a `registration_probe` (`comms_policy_row_exists`):

```python
registered = registration_probe(name=caller)
if registered:
    cause = (f"caller {caller!r} IS registered here but resolves to groups "
             f"{sorted(groups)}, none of which is eligible")
else:
    cause = (f"caller {caller!r} has NO policy row in this store, so its "
             "group could not be determined at all — this is NOT the same "
             "as being denied a group you hold. ...")
return deny_response(reason=(
    f"host_exec is restricted to groups {sorted(ELIGIBLE_GROUPS)}; "
    f"{cause}. Store consulted: {_state_db.DEFAULT_DB_PATH}"))
```

All four states this PR names are covered, and the fourth is covered *earlier*:

| this PR's state | on `develop` today |
|---|---|
| `no_caller` | separate early return — `"host_exec requires a resolvable caller (per-node bearer or 'caller' body claim)"` |
| `ungrouped` | `"IS registered here but resolves to groups [...], none of which is eligible"` |
| `no_policy_row` | `"has NO policy row in this store ... NOT the same as being denied a group you hold"` + likely causes |
| `named` | the resolved set is printed verbatim |

`develop` is also **strictly more informative** on the point this PR cares most about. `GroupResolution.describe()` says *"verify the db path"*; `develop` prints the path — `Store consulted: {DEFAULT_DB_PATH}`. The same distinction was additionally applied to the **spawn** denial (`state_db_nodes.py`), which this PR does not touch, and that one names the remedy command (`sac agents refresh-acl`).

## Merging it now would regress two things

**1. It would reintroduce the misdiagnosis it set out to fix.** `#970` (`577c10b2`, *"group authority is configuration — read the spec, not a per-host cache"*) made group authority resolve from `spec.yaml` first, with the DB row as fallback for nodes with no visible spec. The reason is measured: inside a SIF, `$SCITEX_AGENT_CONTAINER_STATE_DB` is a private per-agent shard holding no policy row for anyone — observed on scitex-compute-04, 2026-08-11 — while the spec on the same filesystem said `['active', 'developer', 'infra']`.

This PR's `resolve_group()` goes **straight to raw SQL**, bypassing the spec:

```python
row = conn.execute(
    "SELECT group_name FROM node_comms_policy WHERE name = ?", (name,)
).fetchone()
if row is None:
    return GroupResolution(group_name="", source="no_policy_row")
```

Post-`#970` that returns `no_policy_row` for a containerised agent whose groups are perfectly readable, and `describe()` would then send the operator to *"verify the db path before treating this as a grouping problem"* — when the db path is fine and the spec already had the answer. That is the same class of misleading diagnostic this PR was written to eliminate.

**2. The add/add conflict is a permissions trap.** `develop` independently created a file at the same path, `_state/state_db_groups.py`, for the same stated reason (`state_db_nodes.py` over the 512-line budget). But `develop`'s version is the post-`#950` **membership** model — `resolve_group_names()`, `in_named_group()`, `is_privileged()`, and `is_developer()` answering *any-of* over the full `labels.groups` set. This PR's version is the older *first-of* model: `is_developer()` compares only the primary group, and there is no `is_privileged` at all.

Resolving that add/add in this PR's favour anywhere would silently revert `ef044b88` (*"authority is group MEMBERSHIP, not the first element of labels.groups"*) — the fix for the 2026-08-10 `grant` incident, where an agent's authority depended on the order of a YAML list. That is a security-relevant regression that no conflict-resolution diff would advertise.

## What survives

Nothing here needs re-doing: the behaviour shipped. The one idea not carried over is the *shape* — `develop` expresses the distinction as inline prose at two call sites rather than as one validated `GroupResolution` carrier. If that duplication is worth removing, it should be a fresh change written against the current spec-first, multi-group model, not this branch. The forced `state_db_nodes.py` extraction also already happened on `develop`.

Closing as superseded by #932, #950 and #970.

PR #943, opened 2026-08-09
fix(state): tell ABSENT / EMPTY / SCHEMA-LESS / POPULATED apart — a 0-byte state.db must not read as an empty fleet

ywatanabe1989:
First of the three the operator ordered — **make the reads honest before migrating anything**, because a migration on top of dishonest reads just hides them.

## The bug

`open_db` calls `init_schema` **unconditionally**. Point it at a wrong path or a zero-byte file and it **creates the schema there**, after which every query truthfully reports zero rows. The store then answers *"no agents registered"* with total confidence — about a database it just invented.

## Measured, 2026-08-09

A 0-byte `state.db` at the pre-cutover path was accepted by `sqlite3.connect`, the tables were built on it, and the fleet read as empty. The same day `sac agents list` reported **6 of 21** agents and `a2a_peers` reported **zero** — while twelve agents were running and answering on that very rail.

An empty read was taken as a fact about the world when it was a fact about the reader.

## Four states, four responses

| state | meaning | what to do |
|---|---|---|
| `absent` | no file | wrong path, or nothing ever wrote here |
| `empty` | 0 bytes / no SQLite magic | a placeholder or truncated create — **not** a database, though `sqlite3` and `init_schema` both accept it |
| `schemaless` | real SQLite, our tables missing | a **different** database, or one never initialised — verify the path |
| `populated` | our tables present | the **only** state where "zero agents" means zero agents |

## `inspect_store` never creates anything

Read-only connect, no `init_schema`. A diagnostic carrying the side effect that *causes* the bug would be useless on the one path you most need to point it at. Two tests pin exactly that:

- inspecting an absent path must not create it
- inspecting a 0-byte file must leave it 0 bytes

## Wired into `sac db show`

Where this bit hardest — it reported all-zero counts from inside a container and was read as an empty fleet. It now classifies the store **before** counting, prints the state and its remedy when not populated, and adds to the JSON:

```json
"store_state": "empty",
"store_path": "/…/state.db",
"counts_are_authoritative": false
```

A zero in that output now means "zero rows" **only** when `counts_are_authoritative` is true.

## Design note

Carried from `GroupResolution` (#942) and deliberately **not** generalised: the multi-state discipline belongs at the **reporting boundary**, not everywhere. Writers keep calling `open_db` and keep getting auto-init, which is correct for them. Only readers that report to a human need to distinguish "empty" from "I looked in the wrong place".

## Tests

**14 new**; **3693 passed / 13 skipped** across `_state` + `cli_pkg` on py3.11.

## Not in this PR

Moving `state.db` into Postgres, and the `config.yaml` peer-topology question. Those are the next two steps in the ordered plan; this one only makes the current reads truthful.


ywatanabe1989:

ywatanabe1989:
**Unstuck: this PR was `DIRTY`, so it could not run CI at all** — one green `CLAssistant` and zero test jobs, which is easy to misread as "queued behind a busy pool". Merged `origin/develop` in (merge, not rebase — force-push is barred by hook). Nothing of the PR's own behaviour was changed.

Single conflict, `cli_pkg/db_group.py`, where both sides independently answered the *same* question — "which store did these counts come from":

* `develop` added a `store` key to the JSON payload plus a dim `store: ...` console line.
* This branch added `store_state` / `store_path` / `counts_are_authoritative` plus the not-populated warning.

**Kept both.** Dropping either would break a consumer that already reads that key, and the two are complementary rather than competing — `store` names the path, `store_state` says whether the counts from it mean anything.

One thing worth flagging rather than silently fixing: `develop`'s added comment warns to read `DEFAULT_DB_PATH` **through the module** (`_state_db.DEFAULT_DB_PATH`) rather than via a from-import, because a captured copy goes stale. That warning does not bite this branch's `inspect_store(DEFAULT_DB_PATH)` call — the `from .._state.state_db import DEFAULT_DB_PATH` is inside the function body, so it re-reads the module attribute at call time. I left it exactly as you wrote it. If you would rather have one spelling for both, that is your call, not a merge decision.

Verified after pushing: `mergeStateStatus` `DIRTY → BLOCKED`, `mergeable` `CONFLICTING → MERGEABLE`, check count 1 → 8. The resolved file passes `py_compile`.

ywatanabe1989:
**Branch updated to `develop`. Your `test_audit_all_clean` failure is this branch's own — it is not the one the other blocked PRs had.**

Twelve PRs were failing `test_audit_all_clean` with the identical assertion text, but the underlying rule differed per PR. Yours was never the `PS-226` one that develop has since deferred:

```
[PA-307 §3 test-quality] tests/scitex_agent_container/_state/test_state_db_health.py:40: STX-TQ005
[PA-307 §3 test-quality] tests/scitex_agent_container/_state/test_state_db_health.py:51: STX-TQ005
  Fixture acquires an external resource (open(...) / connect(...))
```

`test_state_db_health.py` is added by this branch's own commit 8f72357, so updating the branch will not clear these — the two fixtures need to stop acquiring the resource directly.

Everything else on this PR was inherited and should now be gone.

PR #946, opened 2026-08-10
fix(cct): tell "no bot by design" apart from "asked for one and got none"

ywatanabe1989:
`prune_tokenless_telegrammer_mcp` removed the `claude-code-telegrammer` MCP server whenever no token resolved, and logged the same INFO either way:

> this is the intentional no-bot path, not an error

For an agent whose spec **requests** `server:claude-code-telegrammer`, that sentence is false — and it is the only trace the removal leaves.

## Measured

2026-08-10: `scitex-agent-container-04`, `scitex-app`, `scitex-db` and `scitex-priv-setup` each declared the channel, each had the server stripped, each logged only that INFO — while the operator concluded his agents were ignoring him. **91 of 102** fleet specs declare the channel.

Removed is strictly worse than silent: the agent loses **inbound** too, and loses every in-agent means of noticing, because the only self-check is itself an MCP tool.

## Why it was possible

The function's docstring asserted the two cases are indistinguishable. True of the `.mcp.json` entry in isolation, false here — `config.claude.channels` separates them exactly, and `ensure_cct_bot_token` (which runs immediately before) already reads it. The function simply was not passed `config`.

## Changes

- `prune_tokenless_telegrammer_mcp(dest, config=None)`. Case (B) — channel requested, no token — logs **ERROR** naming the agent, the slots tried, the pool source, and that the agent is now mute *and* deaf. Case (A) keeps its INFO verbatim.
- `config` is **optional**: existing one-arg callers keep the old behaviour, because without a spec the function genuinely cannot tell (A) from (B), and silence beats guessing in either direction.
- `ensure_cct_bot_token`'s WARNING said the MCP "comes up without a token, so the bot stays silent". It does not come up at all. A warning whose stated consequence is milder than reality invites the "probably fine" reading.
- One `_requested_telegrammer()` helper, since two functions now need the same fact.

## Verification

13 new tests across all four quadrants (requested+unresolved / bot-less by design / no config / has a token).

No regression, established against a control rather than assumed:

| | passed | failed | errors |
|---|---|---|---|
| clean `develop` | 1712 | 3 | 19 |
| this branch | 1725 (+13 new) | 3 (same) | 19 (same) |

Those 3 pre-existing failures are **carded separately** (`sac-develop-runtimes-suite-is-red-...`), not absorbed here.

### DIFF
--- a/src/scitex_agent_container/runtimes/_cct_token_pool.py
+++ b/src/scitex_agent_container/runtimes/_cct_token_pool.py
@@ -95,6 +95,22 @@ def _upper_snake(text: str) -> str:
     return re.sub(r"[^A-Za-z0-9]+", "_", text.strip()).strip("_").upper()
 
 
+def _requested_telegrammer(config) -> bool:
+    """Did this spec ASK for the Telegram rail?
+
+    The one fact that separates "bot-less by design" from "misconfigured", and
+    the reason both :func:`ensure_cct_bot_token` and
+    :func:`prune_tokenless_telegrammer_mcp` need it rather than one of them.
+    A ``config`` of None means the caller did not say, which is not the same as
+    "no" — callers without a config get the conservative old behaviour.
+    """
+    if config is None:
+        return False
+    claude_spec = getattr(config, "claude", None)
+    channels = list(getattr(claude_spec, "channels", None) or [])
+    return any(str(c).strip() == _TELEGRAMMER_CHANNEL for c in channels)
+
+
 def _slot_candidates(name: str, workdir: str) -> list[str]:
     """Ordered, deduped mechanical slot candidates for an agent.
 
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
Closing in favour of #947, which is merged, covers the same files, and gets the trigger right where this PR gets it wrong.

**This PR keys the ERROR on `config.claude.channels` containing `server:claude-code-telegrammer`.** #947 measured why that is the wrong signal:

| specs requesting the channel | resolve a token | resolve nothing |
|---|---|---|
| 80 | 14 | **66** |

and those 66 include `_template_generalist`, `_template_python_developer` and `_template_researcher`. The templates request the channel, so a channel request measures *scaffolding*, not intent. This PR would have emitted 66 ERROR lines on every deploy into the one panel the operator actually checks — reintroducing, as a fix, precisely the noise the prune was written to remove.

#947 splits on what was **declared** instead, which is the distinction that actually separates "bot-less by design" from "misconfigured".

The diagnosis this PR carried was right and is preserved in #947: the prune's docstring claimed the two cases are indistinguishable, and at that point in the code they are not. Deafness being the surprising half — the absent entry kills inbound too, and `health` cannot self-diagnose because it is itself a tool on the server that went away — is also in #947.

No part of this branch needs salvaging. The 13 tests here assert the wrong contract (they require an ERROR whenever the channel is requested), so porting them would re-import the defect.

PR #949, opened 2026-08-10
feat(start): a bad spec refuses to start — one named override per condition, and the duplicate warning is gone

ywatanabe1989:
## What the operator asked for

He pasted the output of `sac-restart grant`:

```
sac-drift WARNING for agent 'grant':
  the spec source is 5 commit(s) BEHIND origin/develop — you may be launching a STALE spec.
WARN: to_home: agent 'grant' declares no 'to_home_layers'; inheriting the implicit cascade ...
WARN: to_home: agent 'grant' declares no 'to_home_layers'; inheriting the implicit cascade ...
```

and said 「ワーニングだと誰も直さない」— nobody fixes warnings. Then, when asked
how far to take it: 「スペックがおかしかったら起動不可っていうのをデフォルトにして、
ただフォースオプションをつけて起動するみたいな」— refuse by default, with a flag to
start anyway.

Both conditions in his paste now do exactly that.

## 1. The duplicate line was structural, not cosmetic

`to_home: agent 'grant' declares no 'to_home_layers'` printed **twice for one
agent**. It was a `logger.warning` inside `settings_layer_dirs`, which is a
**pure resolver** — and a single start calls it **twice**:

* `claude_session.py` → `deploy_to_home(config, home_dir)`
* `claude_session.py` → `deploy_to_home_overlay(config)` → `deploy_to_home(...)`

(`_tui_workspace.py` does the same pair.)

Suppressing the second copy would have been a patch. The real problem is that a
missing declaration is a property of the **spec**, decided **once per launch**,
not of a path lookup that legitimately runs per deploy. So the check moved to
`_lifecycle/_layers_preflight.check_to_home_layers_at_launch`, which
`agent_start` calls exactly once — and, being a launch preflight, it now refuses
**before** the runtime is built rather than after half a workspace is deployed.

`settings_layer_dirs` logs nothing at all now. The `quiet_undeclared_warning()`
context manager that existed solely to mute it inside `migrate-layers` is
**deleted** rather than left in place muting a logger nothing writes to.

## 2. `to_home_layers` undeclared → refusal (gated on the spec migration)

`UndeclaredToHomeLayers` is raised before launch. Override:
`--allow-undeclared-layers` / `SAC_ALLOW_UNDECLARED_LAYERS=1`.

**It is not on by default yet, and that is the sequencing you asked for.**
Measured on the live registry today:

```
root      : /home/agent/.scitex/agent-container/agents
             -> /home/ywatanabe/.dotfiles/src/.scitex/agent-container/agents
specs     : 102
undeclared: 101
```

Flipping it before those specs are migrated stops the entire fleet booting. So
`_layers_preflight.ENFORCE_BY_DEFAULT = False`, `SAC_ENFORCE_TO_HOME_LAYERS=1`
turns it on per-invocation today, and a test pins the constant so the flip is a
deliberate one-line change with a failing test to update.

**What remains, and who owns it.** The 102 specs are TRACKED files in
`~/.dotfiles`, so the sweep is a dotfiles PR, not a sac sweep — and the
`enforce_worktree_path` hook confines this agent's worktrees to this project, so
I could not create that branch. Handed to `dotfiles` as card
`dotfiles-declare-to-home-layers-in-101-specs-20260810`, with the plan verified
read-only against the live registry:

```
with the scitex-todo spec removed (dotfiles PR #201, already open):
specs     : 101      writable  : 101     already: 0
refused   : 0        unreadable: 0       malformed: 0
layer sets: 65 x [user-shared]
            36 x [user-shared, per-agent]
safe_to_apply: True
```

Today the sweep **refuses** (`safe_to_apply: False`) because the `scitex-todo`
spec is unreadable — 68 missing required fields. dotfiles PR #201 deletes that
spec, which is why it is the unblocker.

Order of operations: **dotfiles spec PR merges → flip `ENFORCE_BY_DEFAULT` →
delete `SAC_ENFORCE_TO_HOME_LAYERS`.** Merging this PR first is safe; it changes
nothing for an undeclared spec beyond printing one line instead of two.

## 3. Spec drift → refuse, and the judgement call I want overruled if you disagree

`--strict-drift` was opt-in. Strict is now the **default**: a spec source that is
BEHIND or DIVERGED refuses to start. Override: `--allow-stale-spec` /
`SAC_ALLOW_STALE_SPEC=1`.

**AHEAD does not refuse.** This is the one place I read the ruling rather than
applying it literally, so it is stated plainly for you to overrule in one line:

* BEHIND / DIVERGED = the spec on disk **may be out of date**. That is
  「スペックがおかしい」and it refuses.
* AHEAD = unpushed local commits. The spec will not **propagate** to other
  hosts, but the spec this host is about to launch is the **newest one that
  exists**. Nothing is wrong with it. Refusing here would ground every host that
  legitimately carries local commits — spartan does, routinely — for a condition
  that is not a bad spec. AHEAD stays a loud warning.
* NOT_A_REPO / UNREACHABLE also keep starting: drift is *unknown* there, not
  present, and refusing on "I could not check" grounds the fleet on a network
  hiccup.

**The consequence you are accepting, stated so it is not a surprise:** a host
whose `~/.dotfiles` checkout falls behind `origin` **stops booting agents** until
someone pulls or passes `--allow-stale-spec`. That is live today — this host's
dotfiles is 5 commits behind, which is exactly the paste that started this. If
that trade is wrong, the fix is one word: make BEHIND a warning and keep only
DIVERGED as the refusal.

## 4. Why named flags and not `--force`

One flag per condition. A blanket `--force` / `--ignore-warnings` skips every
check at once, leaves no record of **which** check was bypassed, and gets reused
by the next person for an unrelated failure — `git --no-verify` is the
cautionary example, and this fleet already bans that idiom by hook.

Using an override logs at **ERROR** naming the condition *and* the agent. A
silent override is just a slower version of the warning nobody read.

Both flags reach their gate through the gate's **env var**, set from a click
callback (`expose_value=False`). That is not a shortcut around plumbing — it is
the only transport that survives the parallel path, which re-execs `sac agents
start` as one **subprocess per agent**. Legacy `SAC_STRICT_DRIFT` is still
honoured in **both** directions, so an existing `SAC_STRICT_DRIFT=0` export keeps
meaning "do not block me" instead of silently becoming a no-op.

## Tests

```
tests/scitex_agent_container/_lifecycle/test__layers_preflight.py ......  [ 11%]
................                                                         [ 42%]
tests/scitex_agent_container/_lifecycle/test__start_drift.py ...........  [ 63%]
...                                                                      [ 69%]
tests/scitex_agent_container/runtimes/test__to_home_resolve.py .........  [ 86%]
.......                                                                  [100%]

============================= 52 passed in 27.27s ==============================
```

The duplication is guarded by two facts, not one: the resolver stays silent
across **two** consecutive calls (the exact shape of the old bug), and a real
`agent_start` produces the finding **exactly once**.

Wider regression run over every touched area:

```
================ 2566 passed, 10 warnings in 447.21s (0:07:27) =================
```
(`_lifecycle/`, `_maintenance/`, `config/`, `runtimes/test__to_home_resolve.py`,
`cli_pkg/test__agents_migrate_layers.py`)

```
============================= 66 passed in 54.67s ==============================
```
(`_drift/`)

https://claude.ai/code/session_01X9aNXY7Tx3jP7vRBSpRJYw

### DIFF
--- a/src/scitex_agent_container/_drift/_local.py
+++ b/src/scitex_agent_container/_drift/_local.py
@@ -19,9 +19,20 @@ Design constraints (per the work item):
 * **RESILIENT** — never raises. A missing git binary, a
   non-repo source dir, an unreachable remote, or any subprocess error
   degrades to ``NOT_A_REPO`` / ``UNREACHABLE`` and the launch proceeds.
-* **DEFAULT = warn loud, NOT block** — hosts like spartan legitimately
-  carry local commits; a hard block would break every start there.
-  ``--strict-drift`` / ``SAC_STRICT_DRIFT=1`` escalate to a hard block.
+* **DEFAULT = REFUSE on a STALE spec** (operator ruling 2026-08-10:
+  「スペックがおかしかったら起動不可っていうのをデフォルトに」). A spec source
+  that is BEHIND / DIVERGED may launch an old spec, so the start is
+  refused and the named override ``--allow-stale-spec`` /
+  ``SAC_ALLOW_STALE_SPEC=1`` is the way past it.
+
+  AHEAD is deliberately NOT a refusal — see :attr:`DriftStatus.is_stale`.
+  Unpushed local commits mean the spec will not PROPAGATE; the spec about
+  to launch is still the newest one that exists, and hosts like spartan
+  legitimately carry local commits. AHEAD stays a loud warning.
+
+  NOT_A_REPO / UNREACHABLE never refuse: drift is *unknown* there, not
+  present, and refusing on "I could not check" would ground every agent
+  the moment the network hiccups.
 """
 
 from __future__ import annotations
@@ -43,6 +54,12 @@ _FETCH_TTL_S = 60
 # unreachable remote must not hang a launch; we bound it and treat a
 # timeout as UNREACHABLE.
 _GIT_TIMEOUT_S = 15
+
+# The NAMED override for the stale-spec refusal. One flag per condition —
+# never a blanket ``--force`` / ``--ignore-warnings``, which skips every check
+# at once and leaves no record of WHICH one was bypassed.
+ALLOW_STALE_FLAG = "--allow-stale-spec"
+ALLOW_STALE_ENV = "SAC_ALLOW_STALE_SPEC"
 
 
 def _run_git(repo: Path, *args: str, timeout: int = _GIT_TIMEOUT_S):
@@ -278,13 +295,19 @@ def check_spec_source_drift(
     )
 
 
-def drift_warning_lines(status: DriftStatus, *, agent: str | None = None) -> list[str]:
-    """Build the loud stderr warning lines for a drifted spec source.
+def drift_warning_lines(
+    status: DriftStatus, *, agent: str | None = None, refusing: bool = False
+) -> list[str]:
+    """Build the loud stderr lines for a drifted spec source.
 
     Returns an empty list when ``status`` is CURRENT (nothing to warn).
     NOT_A_REPO / UNREACHABLE produce a single soft note (drift unknown).
     BEHIND / AHEAD / DIVERGED produce a prominent banner that names the
     drift and the exact fix command.
+
+    ``refusing`` switches the banner from WARNING to ERROR and adds the named
+    override, so the message the operator reads matches what actually happened.
+    A refusal that still says "WARNING" trains the reader to expect a launch.
     """
     if status.state is DriftState.CURRENT:
         return []
@@ -318,14 +341,18 @@ def drift_warning_lines(status: DriftStatus, *, agent: str | None = None) -> lis
             "stale AND unpushed."
         )
     bar = "!" * 72
-    return [
+    lines = [
         bar,
-        f"sac-drift WARNING{who}:",
+        f"sac-drift {'ERROR' if refusing else 'WARNING'}{who}:",
         f"  {why}",
         f"  repo:     {repo}",
         f"  fix:      {fix}",
-        bar,
     ]
+    if refusing:
+        lines.append("  refusing to start — resolve the drift, or start anyway with:")
+        lines.append(f"  override: {ALLOW_STALE_FLAG}   /   {ALLOW_STALE_ENV}=1")
+    lines.append(bar)
+    return lines
 
 
 def warn_if_spec_source_drifted(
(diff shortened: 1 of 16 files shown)


ywatanabe1989:
### Measured on `grant` — the agent from the paste, real spec, nothing started

Loading grant's actual `spec.yaml` and replaying the exact double-resolution a
start performs (`deploy_to_home(home)` then `deploy_to_home_overlay →
deploy_to_home(overlay upper)`):

```
agent          : grant
to_home_layers : None
resolver records: 0 (was 2 before this change)
gate records    : 1
    to_home: agent 'grant' declares no 'to_home_layers' and is inheriting the implicit cascade (user-shared, per-agent). Fix: sac agents migrate-layers --apply. This becomes a start REFUSAL once the fleet's specs are migrated.

--- what a start prints once ENFORCE_BY_DEFAULT flips ---
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
sac-layers ERROR for agent 'grant': refusing to start.
  the spec declares no 'to_home_layers', so what gets merged into
  this agent is invisible from the spec alone.
  spec:      /home/ywatanabe/.dotfiles/src/.scitex/agent-container/agents/grant/spec.yaml
  inherits:  user-shared, per-agent
  fix:       sac agents migrate-layers --apply   (writes the one line it already resolves)
  override:  --allow-undeclared-layers   /   SAC_ALLOW_UNDECLARED_LAYERS=1
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

ywatanabe1989:
### A defect I introduced and then caught — and the test that would have caught it

While re-checking the CLI chain after CI went green, I found that this PR as
first pushed **disabled the drift gate on every CLI start**.

`--strict-drift` is a click flag, so click hands the command `False` when it is
not passed. But since the refusal became the *default*, `_resolve_strict_drift`
reads an explicit `False` as **"the caller demanded leniency"** and returns it
unchanged. So the flag nobody typed was the one turning the refusal off:

```
sac agents start x   ->  strict_drift=False  ->  _resolve_strict_drift(False)  ->  False (lenient)
```

Every test passed, because the unit tests call `agent_start` **without** the
keyword — i.e. `None`, the value the CLI was supposed to send and did not.
Green tests, dead gate.

Fixed: `--strict-drift` now resolves to `True` when passed and **`None`** when
absent (`_true_or_unset`). New file
`tests/.../cli_pkg/lifecycle/test__start_gate_options.py` pins exactly this —
`test_the_cli_default_still_refuses_a_stale_spec` composes the two halves
(what the CLI sends × what the resolver does) rather than testing either alone,
which is the seam the bug lived in.

### Blast radius of the drift refusal, measured rather than assumed

Worth knowing before you judge the trade: the health monitor's **automatic**
restart of a crashed agent does **not** go through this gate.
`_instances.restart_and_record` calls `runtime.start(config)` directly, not
`agent_start` — so a stale dotfiles checkout does **not** stop the supervisor
from recovering an agent that died on its own.

The refusal applies to deliberate starts: `sac agents start`, `sac-restart`, the
MCP `agent_start` tool, and `sac listen`-brokered spawns. That is a much smaller
blast radius than "the fleet stops healing", and it is the part I would have
guessed wrong without reading it.

PR #952, opened 2026-08-10
feat(hooks): a spec can declare the hooks it may not run without, measured IN the container

ywatanabe1989:
## The measurement this is built on

An agent's effective hook set is the union of TWO stacked mounts over `/home/agent`. Measured today for `scitex-agent-container` on this laptop (root reached through a symlink into `~/.dotfiles/src/.scitex/agent-container/`, not the literal `~/.scitex` path):

```
runtime/<agent>/home/.claude/hooks/pre-tool-use                 67   <- layer 1 alone
containers/overlays/<agent>/upper/home/agent/.../pre-tool-use    71
$HOME/.claude/hooks/pre-tool-use    (INSIDE that container)      71   <- effective
```

Same shape for post-tool-use: 8 / 13 / 13. Diffing the layer-1 read against the container's own listing names exactly the hooks it missed — including `log_post_tool_use.sh`, which the host-side read called absent while it was armed:

```
> check_ci_status.sh      > check_develop_branch.sh      > log_post_tool_use.sh
> log_post_tool_use_v01.sh                              > orochi_activity_post.sh
```

Fleet-wide the union spans **15 to 71** pre-tool-use hooks, with nothing declaring which count is correct and nothing detecting the difference.

**The governing conclusion: a guarantee must be measured where it is consumed.**

## What landed

**`spec.required_claude_hooks`** — a floor of NAMES per hook event directory, never a count (71 of the *wrong* hooks would satisfy a count):

```yaml
spec:
  required_claude_hooks:
    pre-tool-use: [enforce_git_dash_C.sh, force_background_bash.sh]
    post-tool-use: [log_post_tool_use.sh]
```

Optional exactly the way `to_home_layers` is: allowlisted in `_KNOWN_SPEC_KEYS`, absent from `_explicit_fields`, `None` when the key is absent. Unlike `to_home_layers` there is **no fleet-wide flip pending** — the declaration IS the opt-in, so not one existing spec changes behaviour. An explicit `{}` stays distinguishable from absence (a spec deliberately requiring no hooks is a statement).

**The inventory is a plain listing** of `$HOME/.claude/hooks/<event>/`. It does not re-implement the union, because inside the container the kernel has already resolved the mount stack — the listing *is* the effective set. Event dirs come from `runtimes._host_merge._HOOK_EVENT_SUBDIRS`, the repo's one enumeration; entry eligibility mirrors that walk's rules so the two cannot disagree about what a hook is. (That turns 71 raw `ls` entries into 69 actual hook scripts — `__pycache__/` and a README are not hooks. Both numbers are "correct"; only one is a count of hooks.)

**The gate runs on the container's own startup path — deliberately NOT in `agent_start`.** `agent_start` runs on the bare host, where the only `$HOME` it can read is the operator's; a gate there would reproduce the exact undercount this PR exists to end, and answer confidently about the wrong machine. So `build_inner_argv` emits one in-container step before `exec`, for declared specs only:

```
<sac> agents hooks <name> || exit 1
```

A non-zero exit aborts the bash chain before `exec`, the container dies, and the existing `_start_failure_diag` path carries the refusal text back. `|| exit 1` is explicit rather than inherited from `set -e`, because the git-alias steps run before any `set -e` and a spec with no `startup_commands` never emits one.

**Refusal semantics mirror PR #949** — same shape, naming, and error style:

| state | behaviour |
|---|---|
| floor undeclared | start; **no refusal, no log line at all** |
| floor satisfied | start, silent |
| floor UNKNOWN | start, logged at ERROR naming what could not be measured |
| floor unmet | **REFUSE**, naming each missing hook + where it should have come from |
| `--allow-missing-hooks` / `SAC_ALLOW_MISSING_HOOKS=1` | start; bypass logged at ERROR naming every missing hook |

One named override per condition, never a blanket `--force` — same reasoning #949 gave.

**UNKNOWN is a third outcome, not a rounding error.** Two sources:

* the hooks root or an event dir could not be read — an agent whose home is not mounted yet must not be reported as one whose guards were removed;
* **the measurement site is wrong.** If `$SAC_NAME` is unset or names a different agent, `measurement_site` reports UNKNOWN and forces the floor verdict to `null`. This is the load-bearing one: on the host the operator's `~/.claude/hooks` is a real, fully readable directory and the *wrong* one. There is deliberately no `False` here — reading the wrong home is never a finding *about* the agent.

UNKNOWN never refuses (refusing on "I could not check" grounds the fleet, exactly as `_drift._local` declines to refuse on NOT_A_REPO / UNREACHABLE) and is never a pass — `ok` stays `null` and the summary names it.

**Read-only reporting: `sac agents hooks [NAME] [--json]`**, plus the MCP tool `agent_hooks`. Output is the cross-package standard shape `{package, ok, checks:[{name,ok,detail,hint}], summary}` that `scitex_cards._health.health` emits — one shape for the fleet, not a second differently-shaped health surface — extended with the raw `inventory` and `floor` so "what does this container enforce" returns the LIST, not only a verdict about it. Every failing **and** every unknown check carries an actionable `hint`; a missing hook's hint names the cascade layer path it should have come from, not just the hook. It is the same command the boot gate runs, so the gate's verdict is reproducible by hand.

Real output from this container:

```
CHECKS
  [ OK ] measurement_site: measuring this agent's own home ($HOME=/home/agent SAC_NAME=scitex-agent-container, container=True)
  [ OK ] hooks_root_readable: /home/agent/.claude/hooks: total=101 (post-tool-use=13, pre-tool-use=69, ...)
  [????] required_hooks_declared: spec declares no 'required_claude_hooks' — nothing is enforced
  [????] required_hooks_present: no floor declared — nothing to satisfy
SUMMARY  2/4 checks passed; unknown: required_hooks_declared, required_hooks_present
```

## Tests — 82 new, and the load-bearing ones would have failed before this change

* **The 67-vs-71 discrepancy is encoded as a fixture** (`_trees.py` — real directories, the measured counts, the actually-missed hook names). Asserted in both directions and at two levels: the layer-only view calls `log_post_tool_use.sh` missing, the effective view finds it, a **control test asserts the two genuinely disagree** (so the others cannot pass for a reason that has quietly evaporated), and the same declaration fails against the layer view while passing against the effective one.
* Floor declared + hook absent -> refuses, and the message names the hook **and** where it should have come from.
* Floor declared + satisfied -> starts. Floor undeclared -> starts and emits **zero** ERROR records.
* Override -> starts, and `BYPASSED` + the missing hook name are recorded.
* **The CLI seam, driven through the real `main` group.** PR #949's defect was that click passed an explicit `False` for an absent flag and the resolver read it as "be lenient", with every unit test green because none crossed that seam. The load-bearing test here invokes the real CLI **with no flags at all** against an unmet floor and asserts exit 1.
* Off-agent reads: a foreign `$HOME` reports `satisfied: null` and does not refuse.
* `--json` stays parseable on stdout **while the gate is refusing on stderr** — a diagnostic that corrupts the machine-readable channel it shares is how a caller learns to stop parsing it.
* A misspelt event dir (`pre_tool_use`) is called a typo rather than a missing hook, so the refusal does not send the reader off to create the bogus directory.

## Two pre-existing over-cap files split to make room

Both were **already over** the 512-line cap before this change, so the field could not be added without a genuine extraction. Import surfaces do not move (everything is re-exported):

* `config/_types.py` 523 -> 472 — supervision-policy dataclasses to `_supervision_types`. `_explicit_fields` derives required spec keys from these via `dataclasses.fields()`, so the import path is load-bearing and is called out in the new module's docstring.
* `config/_loaders.py` 515 -> 497 — declaration parsers to `_parsers/_declarations`, joining the existing one-parser-per-file package.

## Verification

* new suites: **82 passed**
* `config/` + `_mcp/`: **1315 passed**
* `runtimes/`: **1775 passed**, 20 skipped, plus 3 failed / 19 errors in `test__mcp_spec_env.py` + `test__sdk_common.py` that are **pre-existing and unrelated — proven by control, not asserted**: `env -u SAC_SPEC_ENV_KEYS pytest <those two files>` -> **63 passed**. The container exports `SAC_SPEC_ENV_KEYS` naming `SCITEX_AGENT_CONTAINER_AGENT`; the tests unset that key without neutralising the list that names it. Neither file imports `_apptainer_inner_argv`. CI never sees it (a clean SIF has no such var), which is why it survives — filed as `sac-tests-leak-ambient-sac-spec-env-keys-20260810` rather than folded in here.
* `cli_pkg/` (2533 items) exceeds the local time ceiling available to this agent even under `-n 4`; CI runs it in full.

https://claude.ai/code/session_01X9aNXY7Tx3jP7vRBSpRJYw


ywatanabe1989:

**What it is waiting on:** its **own CI**. Failing or missing required checks: pytest-matrix-on-ubuntu-py3.11, pytest-matrix-on-ubuntu-py3.12, pytest-matrix-on-ubuntu-py3.13.

ywatanabe1989:
**Branch updated to `develop`. Your `test_audit_all_clean` failure is this branch's own.**

Twelve PRs were failing `test_audit_all_clean` with the identical assertion text; the underlying rule differed per PR. Yours is:

```
tests/scitex_agent_container/config/test__declarations.py: src likely moved
```

`tests/scitex_agent_container/config/test__declarations.py` exists only on this branch (it is absent from `develop`), so the audit sees a test file with no surviving src counterpart. Updating the branch cannot clear it — the test needs to be pointed at the module's new home, or removed if the declarations it covered are gone.

Note this branch was 68 commits behind before the update, the oldest of the batch.

ywatanabe1989:

PR #959, opened 2026-08-11
feat(dev): the sac dev job group IS the JobSpec kind — service/timer/cron

ywatanabe1989:
## What

`sac dev` job groups are now named after the **`JobSpec` kind**, not the delivery mechanism:

```
sac dev service {list,status,start,stop,restart,enable,disable,install,uninstall}
sac dev timer   {list,status,enable,disable,install,uninstall}
sac dev cron    {list,enable,disable,install,uninstall}
sac dev systemd {list,install,uninstall}          # DEPRECATED, dated, stderr notice
```

This is the ecosystem grammar `scitex-<pkg> dev {service,timer,cron} <verb>` (operator decision, 2026-08-11). `scitex_dev.jobs.ALLOWED_KINDS` has been `{service,timer,cron}` since scitex-dev #153 — this aligns sac's CLI to a taxonomy that already exists.

## Why it is not a cosmetic rename

The old groups were named on the MECHANISM axis while the filter used the KIND axis — two axes, one name. `_load_sac_jobs` was then called with the GROUP NAME, so `sac dev systemd list` asked for `kind="systemd"`, which `JobSpec.validate()` rejects at construction. Nothing could ever match, so every timer sac owns was invisible to its own CLI for weeks behind `No sac systemd-kind jobs.` and exit 0.

Collapsing the group name onto the kind leaves **no name that can be passed where a kind is expected and mean something else**. `_jobs_audit` now machine-checks that:

- `Form.GROUP_IS_NOT_ITS_KIND` — a non-deprecated group whose name is not a legal kind, or that filters on anything other than exactly itself.
- `Form.ALIAS_ONLY_KIND` — a kind reachable ONLY through a deprecated alias, i.e. one with a scheduled loss of its CLI.

Both read the real production `GROUP_KINDS` / `DEPRECATED_GROUPS` from the consumer, not a re-declared copy — same doctrine the module already used.

## Verb sets differ per kind, on purpose

A verb that makes no sense for a kind does not exist for that kind, rather than existing and erroring.

- `service` — a long-running unit, so the full lifecycle applies.
- `timer` — `enable --now` is the systemd idiom; `start`/`stop`/`restart` would be a second spelling with different edge cases, so they are omitted rather than aliased.
- `cron` — a crontab line is present or commented out. There is no runtime object to start or query, so `status`/`start`/`stop` do not exist here.

`install`/`uninstall` take an optional job NAME; every named verb accepts the short local name (`accounts-refresh`) as well as the canonical id (`sac.accounts-refresh`). An unknown name exits **5** and lists the real ones.

## Deprecation with an expiry that is enforced

`sac dev systemd` keeps working with **exactly** its historical three verbs, so nothing new gets built on something with a removal date. It carries real metadata in code:

```
since = 2026-08     remove_after = 2026-10     replacement = sac dev service / sac dev timer
```

`test_the_deprecation_deadline_has_not_passed` turns the build red once that window closes — "当面" becomes permanent unless a date is written down, and a date nobody enforces is the same thing.

The notice goes to **stderr**, never stdout. A courtesy message on stdout is indistinguishable from data; that is exactly how a stale-registry `WARN:` on stdout corrupted `sac host list --json` and turned 7 tests red across three unrelated PRs.

## The delegation seam (`_dev_jobs_backend.py`)

sac resolves a target and shells `scitex-dev ecosystem <group> <verb>`; it does **not** call `systemctl` itself. The argument is in the module docstring: one owner or two disagreeing owners; N packages would otherwise ship N copies of the same `--user`/unit-name/escaping bugs; and the verb set is still moving.

The seam **probes the installed scitex-dev's real Click tree** rather than consulting a hard-coded table, so it picks up `ecosystem service` / `ecosystem timer` the moment those ship — no sac release, no flag. Until then it falls back to today's `ecosystem {cron,systemd}`. A verb neither surface serves exits **4**, naming what was probed and printing the exact manual command:

```
$ sac dev timer status accounts-refresh
`sac dev timer status` is part of the SciTeX job grammar, but the installed scitex-dev
cannot serve it yet: the installed scitex-dev serves neither `ecosystem timer status`
(has: no such group) nor `ecosystem systemd status` (has: ['install', 'list', 'uninstall']).
Run it by hand meanwhile: systemctl --user status sac.accounts-refresh.timer
# exit 4
```

Capability answers are three-state (supported / unsupported / **cannot tell**). An unreadable Click tree degrades to *attempting* the verbs that shipped in 0.16.0, never to a refusal — a false negative would block a command that would have worked.

## Two `--json` tests were asserting on the wrong stream

`tests/…/test__dev_jobs.py::test_systemd_list_filters_out_foreign_jobs` and `::test_systemd_list_json_reports_the_real_kind` failed on plain `develop` with `ModuleNotFoundError: No module named 'scitex_agent_container._jobs_plugin'`. Measured cause, not the one it looks like:

1. `/opt/venv-sac/lib/python3.12/site-packages/~citex_agent_container-0.21.13.dist-info` — a pip-interrupted-install leftover — registers the OLD flat module path under `scitex_dev.jobs`. The repo's `pyproject.toml` and the current dist-info are both correct. **Environmental, not a repo bug.**
2. `discover_jobs` correctly logs that failed provider to **stderr**.
3. The two tests parsed click's `Result.output`, which merges stdout AND stderr. Real `sac dev … --json` stdout was clean the whole time.

So the in-repo defect is the assertion, and it is the one that matters: a stdout-purity test that reads a merged stream can pass while stdout is filthy or fail while it is clean. Every JSON assertion now reads `Result.stdout`, and `test_json_stdout_parses_in_a_real_subprocess` proves the contract end-to-end where the two streams are genuinely separate files (with `PYTHONPATH` pinned to the same source tree the test imported, so a linked worktree is never silently tested against the main checkout).

The stale dist-info itself is container-image state, not repo state — see "Left out" below.

## Proof

```
$ sac dev timer list          # WORKTREE source, real process
  accounts-keepalive       every 15min
  accounts-refresh         every 2h
  fleet-reconcile          every 5min
  freshness-refresh        every 1h
  heal-agent-auth          every 10min
  host-sync-check          every 1h
  restart-login-expired-agents every 5min
  spartan-sif-bake         every 10min
  worktree-gc              every 1d
REAL_RC_TIMER_LIST=0          # nine timers, non-empty — the whole point

$ sac dev service list
No sac service jobs.
REAL_RC_SERVICE_LIST=0        # a TRUE empty: sac declares no kind="service" job
                              # (`sac listen` is deliberately NOT federated)
```

Tests (worktree source via `PYTHONPATH`):

```
tests/…/cli_pkg/test__dev_jobs.py            41 tests
tests/…/cli_pkg/test__dev_jobs_backend.py    13 tests
tests/…/_jobs/test__jobs_audit.py            21 tests
tests/…/_jobs/test__names.py                 10 tests
tests/…/_jobs/test__jobs_plugin.py           55 tests
140 passed in 73.85s          REAL_RC=0
```

## Deliberately left out — and why

- **The canonical rename to `scitex-agent-container-<name>`.** It changes the DERIVED UNIT FILENAME of every sac job, so installing the renamed units without first stopping and removing the old ones leaves two units running the same command. For `sac.accounts-refresh` — the fleet's SOLE OAuth refresher against a SINGLE-USE refresh token — two refreshers revoke each other's access token fleet-wide. That rename must ship WITH the migration verb that enforces stop → remove → install and makes install-before-uninstall impossible, so it is the next PR in this stack, not a rider on a CLI-surface change. `_jobs/_names.JOB_PREFIX` is the seam it flips. **This PR changes no job name, so merging it changes nothing on any host.**
- **A `service` JobSpec for `sac listen`.** Same hazard, documented at length in `_jobs_plugin.py`: the hand-written `sac-listen.service` (hyphen) is supervising the host right now, and a `sac.listen` JobSpec derives `sac.listen.service` (dot) — a different unit. It belongs with the migration.
- **Cleaning the stale `~citex_agent_container-0.21.13.dist-info`.** It is container/venv state baked into the image, shared with other agents; deleting it from inside a code card is an unrequested side effect on someone else's environment. Remediation is an image/host action: remove that `~`-prefixed leftover (pip already treats it as garbage) and `pip install -U --force-reinstall scitex-dev` for the duplicate `scitex_dev-0.38.0` / `0.43.1` dist-infos, which sac's own linter warns about on every run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,13 +48,24 @@ versioning follows [SemVer](https://semver.org/).
 
 - `cli_pkg/_dev_jobs_backend.py` — the explicit, testable delegation seam
   between `sac dev <kind> <verb>` and scitex-dev. It probes the INSTALLED
-  scitex-dev's real Click tree rather than consulting a hard-coded table, so
-  it picks up `ecosystem service` / `ecosystem timer` the moment those ship,
-  with no sac release; until then it falls back to today's `ecosystem
-  {cron,systemd}`. A verb neither surface serves exits 4 naming what was
-  probed and printing the exact `systemctl --user …` command to run by hand.
-  sac deliberately does NOT call `systemctl` itself — the argument is in the
-  module docstring.
+  scitex-dev's real Click tree rather than consulting a hard-coded table, and
+  resolves to a PATH rather than a name, so it survives the ecosystem moving
+  its job groups. Measured on scitex-dev 0.43.1: the groups already live at
+  `ecosystem dev {cron,systemd}`, while `ecosystem cron` / `ecosystem systemd`
+  are deprecated forwarding `Command` shims whose own help says "Removed in
+  v0.50" — they still run, but targeting them breaks on the next upgrade, and
+  because a shim is not a `Group` it enumerates as zero verbs while working
+  perfectly. Candidate paths are tried `dev <kind>` → `<kind>` → `dev <legacy>`
+  → `<legacy>`, so `ecosystem dev service` / `dev timer` (scitex-dev #566) are
+  picked up the moment they ship, with no sac release. An all-empty read is
+  the third state ("cannot tell"), never "unsupported" — including the live
+  case where `ecosystem dev` exists but its per-kind children do not. A verb
+  no surface serves exits 4 naming every path probed and printing the exact
+  `systemctl --user …` command to run by hand. `--dry-run` / `--yes` are
+  forwarded verbatim, because scitex-dev's gate on mutating verbs is what
+  stops `timer disable sac.accounts-refresh` from stopping the fleet's sole
+  OAuth refresher. sac deliberately does NOT call `systemctl` itself — the
+  argument is in the module docstring.
 
 - `_jobs/_names.py` — the local-vs-canonical job-name grammar, with the
   canonical prefix as a single named constant. That constant is the seam for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,73 @@ versioning follows [SemVer](https://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- **`sac dev` job groups are now named after the `JobSpec` KIND, not the
+  delivery mechanism** — `sac dev {service,timer,cron} <verb>`, the
+  ecosystem-wide grammar every SciTeX package adopts (operator decision,
+  2026-08-11). This is not a rename for tidiness. The old groups (`cron`,
+  `systemd`) were named on one axis while the filter used another, and
+  `_load_sac_jobs` was called with the GROUP NAME — so `sac dev systemd list`
+  asked for `kind="systemd"`, which `JobSpec.validate()` rejects at
+  construction, and every timer sac owns was invisible to its own CLI for
+  weeks behind "No sac systemd-kind jobs." and exit 0. With the group name and
+  the kind collapsed into one axis, that bug has no way to be expressed, and
+  `_jobs_audit` machine-checks the identity (`Form.GROUP_IS_NOT_ITS_KIND`)
+  plus the case where a kind is reachable only through a deprecated alias
+  (`Form.ALIAS_ONLY_KIND`).
+
+  The verb set differs per kind on purpose — a verb that makes no sense for a
+  kind does not exist for it rather than existing and erroring. `service` gets
+  the full lifecycle (`status`/`start`/`stop`/`restart`/`enable`/`disable`);
+  `timer` gets `status`/`enable`/`disable` (`enable --now` is the timer idiom,
+  so `start` would be a second spelling of it); `cron` gets `enable`/`disable`
+  (a crontab line has no runtime object to query).
+
+  `install` / `uninstall` now take an optional job NAME, and every named verb
+  accepts the SHORT local name the operator types (`accounts-refresh`) as well
+  as the canonical id (`sac.accounts-refresh`). An unknown name exits 5 and
+  lists the real ones instead of silently doing nothing.
+
+- **`sac dev systemd` is deprecated with a DATE, not "for the time being".**
+  It keeps working and keeps exactly its historical three verbs, so nothing
+  new gets built on it, and it carries machine-readable `since=2026-08` /
+  `remove_after=2026-10` / replacement metadata that a test enforces: the
+  build goes red once the window closes, which is what stops a temporary alias
+  from becoming permanent API. The notice is printed to **stderr** — a
+  courtesy message on stdout is indistinguishable from data, and that is
+  exactly how a stale-registry `WARN:` on stdout corrupted `sac host list
+  --json` and turned 7 tests red across three unrelated PRs.
+
+### Added
+
+- `cli_pkg/_dev_jobs_backend.py` — the explicit, testable delegation seam
+  between `sac dev <kind> <verb>` and scitex-dev. It probes the INSTALLED
+  scitex-dev's real Click tree rather than consulting a hard-coded table, so
+  it picks up `ecosystem service` / `ecosystem timer` the moment those ship,
+  with no sac release; until then it falls back to today's `ecosystem
+  {cron,systemd}`. A verb neither surface serves exits 4 naming what was
+  probed and printing the exact `systemctl --user …` command to run by hand.
+  sac deliberately does NOT call `systemctl` itself — the argument is in the
+  module docstring.
+
+- `_jobs/_names.py` — the local-vs-canonical job-name grammar, with the
+  canonical prefix as a single named constant. That constant is the seam for
+  the ecosystem-wide rename to `scitex-<pkg>-<name>`, which is deliberately
+  NOT in this change: renaming derives different unit filenames, so it must
+  ship with the migration that enforces stop → remove → install.
+
+### Fixed
+
+- **Two `--json` tests asserted on the wrong stream.** They parsed click's
+  `Result.output`, which merges stdout AND stderr, so they passed only while
+  nothing else wrote to stderr and went red the moment an unrelated
+  third-party jobs provider failed to load and `scitex_dev.jobs` warned about
+  it — correctly, on stderr. Real `sac dev … --json` stdout was clean
+  throughout. Every JSON assertion now reads `Result.stdout`, and one test
+  proves the contract end-to-end in a real subprocess where the two streams
+  are genuinely separate files.
+
 ## [0.24.25] - 2026-08-05
 
 ### Fixed
--- a/docs/worktree-gc.md
+++ b/docs/worktree-gc.md
@@ -154,10 +154,14 @@ could not remove anything a daily pass would miss.
 Install (operator-side). sac's own wrapper works again — it used to query
 a dead kind (`jobs_of_kind("systemd")`) and report "No sac systemd-kind
 jobs to install." forever, so this page told you to route around it. Both
-forms below are equivalent; the wrapper just filters `sac.*` for you:
+forms below are equivalent; the wrapper just filters `sac.*` for you.
+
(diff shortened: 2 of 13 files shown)


ywatanabe1989:
### Update after the first CI run — what it caught, and what it did not

**CI found a real bug the local run could not.** The capability probe read `Group.commands`. A **lazy** click group leaves that empty and populates `list_commands(ctx)` instead — so the same code read four cron verbs in my container and **zero** in CI, purely because the scitex-dev installed there builds its tree lazily. Every delegation then resolved to "unsupported", and `sac dev timer list` would have refused a verb that works.

Fixed in two places, because reading the right attribute is not enough on its own:

- `_leaf_verbs` asks `list_commands(ctx)` first, falling back to the eager dict.
- `resolve` now treats *"neither candidate group reported a single verb"* as the **third state — cannot tell** — not as absence. The shipped surface has had `list`/`install`/`uninstall` since 0.16.0, so zero-everywhere is not a credible reading of the world; it is a probe that failed. It degrades to the same "attempt what has shipped" path as a tree that would not build at all. That makes the fix robust to whatever mechanism the *next* scitex-dev uses, rather than to this one.

The probe's positive control now asserts group **presence** rather than verb sets, since static readability of the leaves is exactly the thing that varies; the verb-level contract is asserted through `resolve`, which holds in both worlds.

**One of my tests was asserting two contracts at once.** `test_json_stdout_parses_in_a_real_subprocess` checked stream purity *and* the job count. Those have different failure modes: `list` reads the `scitex_dev.jobs` **entry point** while the expectation imports the provider directly, and the two disagree where the package is on `sys.path` but its metadata is not resolvable — measured, the subprocess returned `[]` in CI while the in-process command returned all nine. Split: this test owns stdout purity, a sibling owns the exit code, and non-emptiness stays pinned in-process where it is deterministic.

**Two gates I had to satisfy:**
- PS-140 §2 — `scitex_dev._cli.ecosystem` is a new cross-package import and must be declared in the runtime import gate.
- The release gate — adding an `[Unreleased]` entry against a pyproject version the CHANGELOG already records as released lets a wheel cache serve the published 0.24.25 and ship none of this. Bumped to **0.25.0** (new CLI surface = MINOR under SemVer).

Re-run of every test file touching a changed source file, against this worktree's source: **158 passed, 1 skipped, real rc=0** — and against the *newly upgraded* scitex-dev (the shared venv moved 0.38 → 0.47 mid-session), which is a useful robustness check for the three-state fix.

### Pre-existing failures in the same CI run — NOT from this PR

Six tests fail on `host probe --json` / `host list --json` with `JSONDecodeError`, because `scitex_dev.hosts._retired` prints its stale-registry `WARN:` to **stdout**:

```
FAILED tests/…/_state/test_host_config.py::test_host_probe_reports_reachable_with_remote_canonical
FAILED tests/…/_state/test_host_config.py::test_host_probe_surfaces_parsed_remote_canonical
FAILED tests/…/_state/test_host_config.py::test_host_probe_reports_unreachable_on_nonzero_exit
FAILED tests/…/_state/test_host_config.py::test_host_probe_surfaces_ssh_exit_code
FAILED tests/…/_state/test_host_config.py::test_host_list_returns_empty_peers_with_no_config
FAILED tests/…/cli_pkg/test_host_group.py::test_host_probe_malformed_remote_stdout_yields_null_canonical
```

This is the exact defect that motivated putting *this* PR's deprecation notice on stderr, and a sibling agent is already on it. Untouched here deliberately — it is a different module with a live owner.

ywatanabe1989:
### Verified against scitex-dev #566's mount point — the probe needed a fix, and it found more

**Q1: did the probe descend into `dev`? NO.** It enumerated `ecosystem`'s direct children only. Fixed, and proved with fixtures shaped like each real ecosystem version (`_tree_566`, `_tree_moved_no_kinds`, `_tree_pre_move` — real `click.Group`s, so the walk has to find its way down):

```
test_the_walk_descends_into_the_dev_subgroup            walked[("dev","timer")] ⊇ {list,status,enable,disable}
test_timer_status_resolves_against_the_566_tree         -> (("dev","timer"), supported=True)
test_service_restart_resolves_against_the_566_tree      -> (("dev","service"), supported=True)
test_the_pre_move_tree_still_resolves_at_the_top_level  -> (("systemd",), supported=True)
```

`Delegation` now carries a **path tuple**, not a name, and `resolve` tries `dev <kind>` → `<kind>` → `dev <legacy>` → `<legacy>`. #566 is picked up the moment it installs, with no sac release.

**Q2: does "`dev` present, kind-children absent" land in cannot-tell? YES —** and it is the LIVE state today, not a hypothetical:

```
test_dev_present_with_kind_children_absent_is_cannot_tell   resolve("service","install").supported is True
test_a_totally_empty_tree_is_cannot_tell_not_unsupported    resolve("timer","install").supported is True
test_cannot_tell_prefers_dev_when_a_dev_subgroup_exists     -> ("dev","systemd")
```

**Q3: are `--dry-run` / `--yes` preserved? `--yes` was; `--dry-run` did not exist at all.** Both now travel verbatim, `--dry-run` is offered on every mutating verb, and `build_argv` is split out so the exact argv is asserted:

```
test_the_pass_through_forwards_dry_run   ["sd","ecosystem","dev","timer","disable","--name","sac.accounts-refresh","--dry-run"]
test_every_mutating_verb_offers_the_gate_flags   {} missing
```

End-to-end, real process, worktree source:

```
$ sac dev timer install worktree-gc --dry-run
# sac.worktree-gc.service … # sac.worktree-gc.timer …    REAL_RC_DRYRUN=0
```

(`--dry-run` demonstrably reached scitex-dev — without it, install refuses without `--yes` and exits 2.)

### What the measurement turned up that nobody asked about

The move is **already live in the installed scitex-dev 0.43.1**, and the old names are worse than merely relocated:

```
ecosystem cron        -> click.Command  "(deprecated) Forwards to 'ecosystem dev cron'. Removed in v0.50."
ecosystem systemd     -> click.Command  same deprecation + removal
ecosystem dev         -> SpecGroup  ['cron','systemd']
ecosystem dev cron    -> SpecGroup  ['exec','install','list','uninstall']
ecosystem dev systemd -> SpecGroup  ['install','list','uninstall']
ecosystem dev service -> ABSENT      (arrives with #566)
ecosystem dev timer   -> ABSENT      (arrives with #566)
```

Three consequences:

1. The shims **still run**, so `sac dev timer install` was not broken today — it was aimed at something scitex-dev **deletes in v0.50**. A test now asserts the live resolution is not `("systemd",)`.
2. A shim is a `Command`, not a `Group` — no `list_commands` — so it enumerates as **zero verbs while working perfectly**. Zero verbs is evidence about the probe, never about the surface.
3. **Correction to my previous commit:** I attributed the "four cron verbs locally, zero in CI" split to lazy Click groups. That was wrong — I asserted a mechanism I had not measured; the real cause is the shim. Recorded in the code and the commit rather than quietly edited out, because the next person to see a zero will reach for the same wrong explanation. Reading `list_commands` is still correct, it just was not the fix for that symptom.

### Verb sets — matched, with two deliberate divergences

`service` **loses** `enable`/`disable` (I had them; #566 does not, so they would be a permanent exit-4 — a declaration with no live counterpart). `cron exec` is **not** mirrored: sac declares no cron job, and `exec` takes a positional name plus `--apply` rather than the uniform `--name`, so wiring an untestable special case for an empty group is the same disease facing the other way. It belongs with the first cron job sac owns. Both choices are pinned by `test_no_verb_is_exposed_that_the_counterpart_will_not_serve` against the transcribed #566 sets.

### Correction carried

The 7 host-registry failures are **not** a stdout leak in scitex-dev — the WARN goes to stderr; the breakage is click's `result.output` interleaving, fixed in PR #961 (and cherry-picked onto #958). Same root cause as the two `--json` tests fixed here, which is a nice corroboration. #959 will be rebased once either lands; no scitex-dev change is being waited on.

Targeted re-run, worktree source: **178 passed, 1 skipped, real rc 0.**

PR #964, opened 2026-08-11
feat(relocate): the last three phases — a canary that ran end to end to DONE

ywatanabe1989:
Completes `sac agents relocate`. `target_standby`, `handshake` and `handover` are built, so **no phase returns "no adapter" any more**, and the canary ran **end to end to DONE**. Based on #963's head (still open at the time of writing).

Each refusal's own sentence was the specification. Every finding below was **measured on two real hosts**, not reasoned about.

---

## The canary, run to DONE

`sac agents relocate canary-resume-test --to scitex-compute-04 --no-dry-run`, run host-side, exit code **0**.

| phase | result | measured |
|---|---|---|
| preflight | **PASS** | 11/11 GO |
| source_drain | **PASS** | vacuous — tmux on ywata-note-win answered and has NO session `tui-canary-resume-test` (positive evidence of absence) |
| source_stop | **PASS** | already stopped, verified by a second independent observation |
| transport | **PASS** | `afb6da24-…jsonl` **3,707,306 bytes / 4,165 lines**, counted independently on EACH host; `memory/` refused by name |
| target_standby | **PASS** | spec dir **6 files / 30,692 bytes**, tree identical on both hosts; marker seeded + read back; started as standby (no lease) |
| handshake | **PASS** | nonce `a91407d4f351ff54`, answer `scitex-compute-04` observed after 3 looks; 3 occurrences of our own placeholder correctly ignored |
| handover | **PASS** | lease `ywata-note-win → scitex-compute-04` at **fence 1** (bootstrapped at fence 0), re-read to confirm |
| done | **PASS** | both hosts observed; residency written; source transcript moved aside to `.old/20260811T143929Z/` |

### Exactly ONE owner, fleet-wide

```
ywata-note-win:      0 live instances
scitex-compute-04:   1 live instance
residency (open):    canary-resume-test → scitex-compute-04
lease:               holder=scitex-compute-04 fence=1
```

### Attempt history accumulates (the requested fix, demonstrated live)

```
journal attempts: [(1,'aborted'), (2,'aborted'), (3,'aborted'), (4,'done')]
```
The three failed attempts are all still there. Under the old one-row-per-agent key, attempt 4 would have erased them.

### Resume continuity — proven, not assumed

The agent was asked, **from memory only**, for a cron id it had reported nine times in the carried conversation. That id appears **18×** in the carried transcript and **0×** in the fresh session file:

```
memcheck-c93e77 cron=6edceddc   ✅
```
Before the fixes below, the same question returned a cron id it had created that minute.

---

## Four bugs found by running it

**1. The unit is the spec DIRECTORY, not `spec.yaml`.** Carrying the single file left `to_home/` behind and sac's own spec-drift guard refused the boot — correctly. The whole directory now travels, with a full-tree (path + size) comparison on both hosts and the credential-basename refusal repeated at directory level so a dir copy cannot smuggle one past the per-file allowlist.

**2. The standby start needs `--no-redispatch`.** `sac agents start` routes by the spec's `host:` pin, and a carried spec still pins the host being LEFT — so the start issued **on the target ssh'd back and booted the agent on the SOURCE**. Measured: `'canary-resume-test' started on 'ywata-note-win'` from a command running on scitex-compute-04. The gate held (nothing was handed over), but both hosts are now observed *before* either is judged, because judging the target first hid the live instance on the source.

**3. `load_lease` could not build a `Lease` at all.** The table had no `token` column and `Lease` refuses an empty token, so the persistence layer for a carefully pure decision module was effectively **write-only**. Column added; a pre-token row reads as `None`, because a holder that cannot present a token cannot prove it holds anything.

**4. The TUI continue-gate looked in the overlay while the spec bound the transcript.** This is the one that would have shipped as silence. The relocation completed *perfectly* — bytes verified, marker seeded, agent started, handshake answered — **and the agent started fresh**, with no memory of 3.7 MB sitting one directory away. `_home_has_resumable_conversation` checked the overlay upper-home (real, and empty) and never consulted the spec's binds, so `-c` was silently omitted. It now walks candidates most-specific-first through the *same* `transcript_home_from_spec` derivation the transport uses, so where a relocation WRITES and where the gate LOOKS cannot drift apart. It can only ever *add* `-c`, and only when a transcript was actually found.

**And a fifth, on top of #4:** `--session continue` means *"the latest conversation for this home"*, which is only the right session while the carried one is the only one there. A stray newer session on the target made a continue-mode boot resume **that** — the agent came up *with* a memory, just not the carried one, which no byte count catches. The standby now resumes **by id**: `sac agents start --resume <carried uuid>`. That also makes the boot agree with the marker the standby wrote (`--resume` implies `--session resume`). **This needs no deployment anywhere** — `--resume` is an existing flag, and it was proven against 04's unmodified sac install.

---

## Design notes worth reading

**The handshake's honesty.** The challenge is POSTed to the agent's sidecar on its own host; the answer is read out of the agent's **transcript**, whose bytes then travel to the source where the coordinator parses them. So *the coordinator on the source* observed the reply — a network path from target back to source is **not** what was measured, and `observed_by` says so in those words rather than implying the stronger thing. It has to be sought there: a TUI agent's `/v1/turn` returns `{"text": ""}` by design, so a synchronous send proves DELIVERY only — the exact accepted-but-silent shape.

The brief we inject *contains* `nonce=…`, so a naive search finds our own message first. `_relocate_reply` tells them apart by the angle-bracket placeholder the brief writes and a hostname never has. UNKNOWN stays UNKNOWN: only a delivery that was **accepted** and a search that **ran to the end of its patience** may be called "no reply".

**The handover's bootstrap.** sac still does not claim a lease when an agent starts, so for an agent that has never relocated the store is empty. `check_write` is explicit that an empty store must be resolved *deliberately* rather than inherited, so this bootstraps, records the bootstrap as such in the journal, and starts the fence at 0 so the handoff lands at 1. A **live** lease held by a third host is a refusal, not a bootstrap. Nothing renews the lease yet — stated plainly in the phase's own output, because it bounds what the lease buys: after expiry a fresh claim is permitted and the **fence** is what still excludes a stale holder.

**`done` now measures before it writes.** Both hosts are observed for exactly one live instance before residency is written. A residency row that disagrees with the machines is worse than none — it is a confident wrong answer everything downstream believes.

---

## Left out, deliberately

- **`source_drain` still has no adapter for a RUNNING source.** Nothing here can tell an agent to finish its in-flight work, take no new work, and confirm it did. A stopped source drains vacuously (measured, not assumed), which covers 23 of the 24 agents. The readiness notice still says so rather than claiming otherwise.
- **Nothing claims a lease at agent start-up.** Wiring that into `agent_start` fleet-wide is a separate, riskier change; the handover bootstraps instead and says it did.
- **The handshake does not prove a target→source network path.** See above — it proves the agent's loop and tools work, which is what the gate exists for.

## Test status

**4768 passed, 14 skipped, 0 failed.** 36 errors, all in `test__rename_cards.py` / `test__rename_board.py` — the pre-existing scitex-cards outage (`notifications row has no record_json payload`), unrelated to this branch.


ywatanabe1989:
### Measured on the live canary: the relocate CLI reads a different residency source than the relocation writes, and then reports the disagreement as agreement

Verifying this PR against `canary-resume-test` (currently resident on `scitex-compute-04` after attempt 4 reached `done`). The relocation itself is correct — the journal, the lease and the residency table all agree the agent is on compute-04, and the agent really is running there on the seeded session. The **dry run's host notice is wrong**, and wrong in the specific way this PR's own docstrings say must be avoided.

**What the dry run prints today**

```
note: the spec still carries host: ywata-note-win; it is ignored (the db agrees, at ywata-note-win).
      Delete the field — it will not be consulted again
```

**What the store actually holds**

```
current_residency('canary-resume-test')  -> scitex-compute-04
read_residency_history(...)              -> Residency(host='scitex-compute-04', from_ts=1786459210.41, to_ts=None)
load_lease(...)                          -> Lease(holder='scitex-compute-04', fence=1,
                                                  token='f78c9889...', expires_at=1786545610.33)
```

**The mechanism**, measured in one process, same env, same db (`/home/ywatanabe/.scitex/agent-container/runtime/state.db`):

```
list_active_instances() for canary          -> []
_relocate_cmd._residency_history(...)       -> ()                      # CLI's reader knows nothing
resolve_host((), legacy_spec_host='ywata-note-win', now=...)
                                            -> host='ywata-note-win'
                                               code=201 (CODE_SEEDED_FROM_SPEC)
                                               seeded_from_spec=True
current_residency(...)                      -> scitex-compute-04       # the truth
```

Two separate defects, both in `cli_pkg/_relocate_cmd.py` / `_lifecycle/_relocate_host_record.py`:

1. **The CLI reads the superseded source.** `_residency_history()` reads `instances` and its docstring still asserts *"THERE IS NO RESIDENCY TABLE YET"*. This PR adds exactly that table — `agent_residency`, with `record_residency` / `read_residency_history` / `current_residency` — and the `done` phase writes it (`residency reads scitex-compute-04`). So after a successful relocation the CLI reports the agent on the host it *left*. It should call `current_residency()`.

2. **The notice is self-confirming.** When `resolve_host` falls through to `CODE_SEEDED_FROM_SPEC`, `db_host` is *copied from* `spec_host`, so `stale == db_host` holds by construction and `legacy_spec_host_notice` takes the "the db agrees" branch. It tells the operator the db confirms the spec when the db said nothing at all. That is precisely the failure the function's own docstring names: *"the failure mode being avoided is an operator reading `host: nas-03` in a file, believing it, and being wrong"*. The notice needs to branch on the code and say "seeded from the spec — the db held nothing", never "the db agrees".

Defect 1 is what makes defect 2 reachable here: because `_residency_history()` returns `()` for a stopped-on-both-hosts agent, the seed path runs on an agent whose residency *is* recorded, just in the table the CLI doesn't read.

**Not affected:** the executing path. Attempt 4's journal is complete and correct — `preflight → source_drain → source_stop → transport → target_standby → handshake → handover → done` in 40.6 s, transport byte+line verified, standby seeded to resume `afb6da24-6a6b-457a-a154-5aa048d065ed`, handshake nonce `a91407d4f351ff54`, lease `ywata-note-win → scitex-compute-04` at fence 1, source transcript moved aside to `.old/20260811T143929Z/`. The `(agent, attempt)` journal key also verified working: all 4 attempts are retained rather than overwriting each other, so the three earlier failures are still readable.

PR #968, opened 2026-08-11
fix(relocate): measure the source at the instant you read it, and name the session

ywatanabe1989:
Five fixes, all from the relocation that failed on the night of 2026-08-12. Each was approved by the operator; none is a redesign.

## The abort that produced the first three

```
MISMATCH b68520e1-….jsonl:
  sent   108,412,278 bytes / 58,325 lines
  target 108,423,995 bytes / 58,325 lines
```

Identical **line** counts, 11,717 bytes apart, and the **target holds more**. No record was appended — the final line got longer. The agent had already been stopped and `source_stop` had confirmed it; what grew the file was the dying process's shutdown flush finishing its last line after tmux had let go of the session. The source was *measured* at one instant and *read* at another, and nothing guaranteed it was unchanged in between.

### 1. `source_stop` waits for the file, not only for tmux

A tmux session disappearing is not the process exiting and releasing its file descriptor. The stop phase now goes through `_relocate_quiescence`, which samples every `*.jsonl` in the source's project directory and requires **two consecutive identical `(size, mtime)` readings, 2 s apart, bounded at 30 s**.

- 2 s is long enough that a shutdown flush cannot span both readings and still look stable, and costs a healthy relocation one interval.
- 30 s is ~15 readings — long enough for a process slow to let go, short enough that a wedged host fails the phase promptly rather than sitting in front of the copy's 900 s budget.
- A wait that runs out returns **UNKNOWN naming what moved and by how much**, never success, and never `False` (the agent *is* stopped; that would send someone to stop a process already down).
- **Both** exits from the liveness check go through it, including `already stopped` — which is the exit the failing run actually took.

An open-descriptor check would be stronger and is deliberately not used: `lsof`/`fuser` are absent on the fleet's busybox targets and need privileges elsewhere, so its unavailable case is UNKNOWN on hosts where the cheap check answers plainly.

### 2. Transport snapshots a byte offset first

Before anything moves, `snapshot_transcripts` records each file's offset of its **last newline**; the copy carries exactly that many bytes; and arrival is verified against **that recorded number**, not a fresh reading of a source that may have moved. The source is now free to grow underneath the transfer without changing the answer.

The cut is at a newline specifically: an arbitrary offset lands mid-record and hands the target malformed JSON inside an otherwise valid JSONL file. Losing at most one partially-written record is the accepted trade — the module's own abort text already predicts it.

**The copy changed from tar to `head -c <offset>` per file.** tar cannot carry part of a file, and a whole file is by definition not a snapshot. The exact-names property tar was chosen for is kept (paths are built in Python; nothing globs on either side). The cost is one ssh per transcript, chained under `&&` inside a single host-side `bash -c`, so it is still one `host_exec` and one exit code. Measured before choosing it: the busiest project directory on this fleet holds six transcripts. Truncating on the *target* instead would keep one connection but ship the extra bytes and then discard them, and would need `truncate` or a temp-file rewrite on hosts that may have neither.

### 3. The mismatch message distinguishes growth from loss

`CODE_TRUNCATED` was documented as *"fewer bytes or lines"* and fired on **more**, reporting it identically.

| shape | code | what it means |
|---|---|---|
| `target < sent` | 422 | bytes went missing; the copy is suspect |
| `target > sent` | **409 (new)** | nothing went missing; the source settled after being measured |
| absent | 404 | the copy did not happen |

Both still refuse. The mismatch line now names the direction and the signed delta, and when the line count is unchanged it says out loud that no record was appended — a line that was already there simply got longer. Worst-first: one short file in a mixed set still reports 422.

### 4. The session to resume is chosen, not assumed

The transport named it with `if len(plan.files) == 1`. Measured on ywata-note-win: **not one of the ten agents left to move has exactly one transcript** (3, 4, 4, 5, 3, 2, 3, 2, 4, 4). Every one returned `GO — every check passed (11 checks)` and then aborted at `TARGET_STANDBY` — *after* `source_stop` had taken it down. Confirmed live on scitex-clew: three transcripts, all byte-and-line verified, then `phase='aborted'`, no marker seeded, target never started, source left stopped.

`_relocate_session_choice` prefers, in order:

1. the source's own `runtime/<agent>/session_id` marker, when it names one of the carried files — the answer, written by the thing that knows;
2. otherwise the most recently modified carried `*.jsonl`.

Neither resolving is a refusal that **names the candidates**: a marker naming an uncarried file (409), a tie on the newest mtime (300), an unread marker with several candidates (503), no transcripts at all (404). An unread marker is deliberately not the same as an absent one — it may name a file other than the newest, which is exactly where guessing is wrong.

The same pure function answers at **preflight**, as the new `session_resolvable` check, so the refusal happens while the agent is still up rather than three phases later. The gap was as much the bug as the guard was: a check that passes on ten agents that cannot proceed is not a check.

## Module splits

Two files were split to stay under the 512-line cap, each along its own seam:

- `_relocate_transport_verify` (did it arrive) out of `_relocate_transport` (what may travel) — the public surface is unchanged and re-exported.
- `_relocate_effects_transport` out of `_relocate_effects`, as a per-phase mixin, following the convention the other four phases already use.

## Tests

Real files and real byte counts under `tmp_path`. The `exec_fn` seam runs the rendered scripts through `subprocess`, so `wc`, `head` and `stat` actually execute and the numbers are the filesystem's. Nothing is mocked. The clock and the sleep **are** injected — they are the two things a test must not spend; the fake sleep advances the fake clock by exactly the interval requested, so the deadline arithmetic runs at full fidelity in no wall-clock time.

Covered, at minimum:

- a file that **grows** between the snapshot and the copy still verifies, because verification is against the recorded offset — and the growth does not reach the target
- a file whose final record is half-written snapshots to the **last newline**; the carried prefix ends in `\n` and verifies
- `target < sent` and `target > sent` produce **different** mismatch lines and different hints
- `source_stop` returns **UNKNOWN, not PASS**, when the file is still changing at the deadline — including on the `already stopped` path
- multiple transcripts + a valid marker → the marked file (which is deliberately *not* the newest in the fixture, so the two rules are distinguishable)
- multiple transcripts + no marker → the newest
- a marker naming an uncarried file → refuses, naming both the marker and every candidate
- zero transcripts → refuses (existing behaviour, unregressed)
- an agent whose session cannot be resolved **fails preflight** rather than passing and aborting later

New test files mirror source modules of the same stem, per PS-204 §2: `test__relocate_quiescence`, `test__relocate_session_choice`, `test__relocate_transport_verify`, `test__relocate_effects_source`. No orphans.

## Not in this PR

`~/.scitex/agent-container/bin/self-relocate.sh` piped `sac agents relocate` through `tail` and never inspected its exit status, so it sailed past a 422 that said *"do NOT continue the relocation"* in plain text. It is fixed on **ywata-note-win** (where it actually lives — not on compute-04, contrary to the brief): the relocate output goes to a log, `$?` is read directly with no pipeline in the way, and anything other than 0 stops the script. The original was displaced to `bin/.old/20260811T220323Z/`, not deleted. It only auto-rolls back when the log shows `past_no_return=False`; otherwise it refuses to start the agent on either host and says so.

https://claude.ai/code/session_01NcDz9gLL2Ct24FXqX2SF9T

---

## Second commit: F-35 — `memory/` was left behind, plus a regression it exposed

On 2026-08-11 a relocation that ran end to end to DONE logged
`transport: REFUSED memory — only conversation transcripts (.jsonl) travel; this is not one`
and landed an agent whose target `memory/` was **empty** while nine files and 20,014 bytes stayed on the source — two of them written that evening from the operator's own corrections. The transcript is the conversation; memory is the rules distilled from it. Carrying only the first moves an agent that can recall what was said and has lost every rule it wrote for itself.

- `memory/` joins the allowlist and travels **whole, by tar** — not down the snapshot path. A memory note is not an append-only log being flushed, so there is no last newline to cut at, and half of one is not a shorter version of it.
- `select_transferable` now returns transcripts and directories **apart**, because they are carried by different means. One merged list would have to be re-derived by every caller, and the first to get it wrong would snapshot a directory or carry a live log unbounded.
- Arrival is confirmed by comparing the **whole tree** on both hosts — a per-file check of the names you thought to name says nothing about the file you forgot.

**A regression this PR's first commit introduced, caught here.** Making `copy_transcripts` snapshot-bounded broke `_relocate_effects_standby._carry_spec`, which hands it a bare *directory* name — it would have raised `AttributeError` at the phase that carries the spec. **No test caught it: there is no test file for the standby module.** The two copies are now two functions, `copy_tree` (whole entries) and `copy_transcripts` (bounded by a recorded offset).

### Provenance, which the operator ruled is the actual requirement

> 「エージェントがどこから来たのかっていうのが分かって、そこを調査することができるならば、全く問題ない」
> — if the agent can tell where it came from and go and investigate there, there is no problem at all.

He called the merge policy 枝葉, a leaf. So `_relocate_provenance` writes **`RELOCATED-FROM.md`** into the target's project directory: source host, source absolute path, raw unix time (skew ruled 「そんなシビアじゃない」), session resumed, what was carried with the counts as carried — and **what was not, with the reason for each**.

That last section is the half that goes missing, and it is exactly how `memory/` was lost: the refusal existed, as one line in a run log nobody keeps. It now lands beside the conversation, where the agent that lost the thing is the one who reads it.

**No mtime merge, and that is a consequence rather than a shortcut.** When the destination already holds a project directory, the existing transport moves the **whole** directory aside to `.old/<stamp>/` before anything is written — so there is no live directory to merge into and no file can be orphaned by a `MEMORY.md` that lost its lines. The source wins completely (「ソースの方が普通大切だろう」), and the target's previous copy is intact, named in the record, and a deliberate merge away.

The provenance write is **never fatal**: by the time it runs the transcript is verified on the target, and refusing the phase over a missing note would abort a move that worked, leaving the agent stopped on one host and unstarted on the other. A failure to confirm it logs a WARNING saying the agent will not be able to read where it came from.

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_relocate_checks.py
+++ b/src/scitex_agent_container/_lifecycle/_relocate_checks.py
@@ -43,6 +43,8 @@ from __future__ import annotations
 from typing import Final
 
 from ._relocate_preflight_facts import Check, SourceFacts, TargetFacts
+from ._relocate_session_choice import CODE_UNKNOWN as SESSION_UNKNOWN
+from ._relocate_session_choice import choose_session
 
 __all__ = [
     "CHECK_BINDS",
@@ -55,6 +57,7 @@ __all__ = [
     "CHECK_RUNTIME",
     "CHECK_SAC_PRESENT",
     "CHECK_SCHEMA",
+    "CHECK_SESSION",
     "CHECK_SOURCE_WORK",
     "check_binds",
     "check_card_store",
@@ -66,6 +69,7 @@ __all__ = [
     "check_runtime",
     "check_sac_present",
     "check_schema",
+    "check_session_resolvable",
     "check_source_work",
 ]
 
@@ -80,6 +84,7 @@ CHECK_PORTS: Final = "ports_free"
 CHECK_HUB_FROM_TARGET: Final = "hub_reachable_from_target"
 CHECK_SAC_PRESENT: Final = "sac_present_on_target"
 CHECK_SOURCE_WORK: Final = "source_work_committed"
+CHECK_SESSION: Final = "session_resolvable"
 
 
 def _unobserved(name: str, what: str) -> Check:
@@ -321,6 +326,55 @@ def check_sac_present(facts: TargetFacts, to_host: str) -> Check:
             f"absolute path. The evidence: ssh {to_host} 'command -v sac' printed "
             f"nothing while {facts.sac_resolved_path} exists there"
         ),
+    )
+
+
+def check_session_resolvable(source: SourceFacts, agent: str) -> Check:
+    """Can the conversation to resume be NAMED, before anything is stopped?
+
+    This check exists because the answer used to be discovered three phases too
+    late. TARGET_STANDBY refuses without a session id, and TARGET_STANDBY runs
+    after SOURCE_STOP — so an agent whose session could not be identified was
+    taken down, had its transcript copied and verified, and then aborted with the
+    marker unwritten and nothing running anywhere.
+
+    Measured 2026-08-12 on ywata-note-win: ten agents, every one holding between
+    two and five transcripts, every one reporting ``GO — every check passed``, and
+    not one of them able to complete. The guard was ``len(files) == 1``. A check
+    that passes on ten agents that cannot proceed is not a check; the gap is as
+    much the bug as the guard was.
+    """
+    if source.transcripts is None:
+        return Check(
+            name=CHECK_SESSION,
+            ok=None,
+            detail=f"the transcripts for {agent} on the source were not listed",
+            hint=(
+                "list the source's project directory before deciding. Which session "
+                "travels is settled here or it is settled after the agent has been "
+                "stopped, and only one of those is recoverable without an operator"
+            ),
+        )
+    choice = choose_session(
+        agent=agent,
+        carried=[name for name, _ in source.transcripts],
+        marker=source.session_marker,
+        mtimes=dict(source.transcripts),
+    )
+    if choice.session is not None:
+        return Check(
+            name=CHECK_SESSION,
+            ok=True,
+            detail=(
+                f"session {choice.session} would travel, chosen by {choice.chosen_by} "
+                f"from {len(choice.candidates)} candidate(s)"
+            ),
+        )
+    return Check(
+        name=CHECK_SESSION,
+        ok=None if choice.code == SESSION_UNKNOWN else False,
+        detail=f"{choice.reason} ({choice.code})",
+        hint=choice.hint,
     )
 
 
--- a/src/scitex_agent_container/_lifecycle/_relocate_effects.py
+++ b/src/scitex_agent_container/_lifecycle/_relocate_effects.py
@@ -48,27 +48,24 @@ from ._relocate_effects_handover import HandoverEffects
 from ._relocate_effects_handshake import HandshakeEffects
 from ._relocate_effects_source import SourceEffects
 from ._relocate_effects_standby import StandbyEffects
+from ._relocate_effects_transport import TransportEffects
 from ._relocate_execute import PhaseEffects, StepResult
 from ._relocate_liveness import observe_running
 from ._relocate_move_aside import move_aside_destination
-from ._relocate_shell import Shell, resolved_path, shell_for
-from ._relocate_transcript_home import transcript_home_from_spec
-from ._relocate_transport import TranscriptFile, plan_transport, verify_arrival
-from ._relocate_transport_paths import derive_target_dir
-from ._relocate_transport_ssh import (
-    copy_transcripts,
-    ensure_dir,
-    list_transcript_dir,
-    measure_transcripts,
-    move_dir_aside,
-)
+from ._relocate_shell import Shell, shell_for
+from ._relocate_transport import TranscriptFile
+from ._relocate_transport_ssh import move_dir_aside
 
 __all__ = ["RelocateAdapters", "adapters_for", "build_effects"]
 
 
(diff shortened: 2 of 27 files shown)


ywatanabe1989:
**Local verification.**

`_lifecycle` + `cli_pkg`, after both commits, against this worktree's source:

```
4408 passed, 1 skipped, 16 deselected, 23 warnings in 178.92s (0:02:58)
EXIT=0
```

Adding `tests/develop` gives one failure, which is **pre-existing and not from this PR**:

```
ERRO: [§6a] env var 'CODEX_HOME' has no recognized prefix
ERRO: [§6a] env var 'LOG_PATH'   has no recognized prefix
```

Measured rather than assumed:

- the diff contains **zero** occurrences of either name;
- both live in `src/scitex_agent_container/_account/codex_account.py`, untouched here;
- running the same audit against the **`develop` checkout** in this container reports the same two errors.

Not fixed here, deliberately: `CODEX_HOME` is **Codex's own** environment variable, read to locate that tool's login — prefixing it would stop it finding the file. The remedy is the audit allowlist, a separate change with its own reasoning. CI is green on `develop` because the allowlist dictionaries the audit looks for (`<repo>/.scitex/dev/cli-audit-dict.yaml`, `~/.scitex/dev/cli-audit-dict.yaml`) are both reported **absent in this container**.

**What is NOT verified by test, and cannot be from here:** that two real hosts agree — the tar-over-ssh copy of `memory/`, the `head -c` transcript copy, and the base64 provenance write all run through `build_probe_argv`, so they need a second machine. The rendered commands are pinned; the round trip is what the canary run is for. This is the same limitation `_relocate_transport_ssh`'s own docstring states.

PR #983, opened 2026-08-12
feat(credstate): credentials become STATE — the database records references, never material

ywatanabe1989:
Credentials stop being facts about a filesystem and become rows: which account, which node is primary, which tier, when it expires, and a **locator** — a reference (`file:<abs path>` / `env:<VARNAME>`) that makes the material findable, refreshable and auditable **without being present**.

## The exposure question, argued (ADR-0023 §3)

The database records credential FACTS and never credential MATERIAL — not the secret, and **not a digest of it**. Four grounds:

1. **Storing bytes addresses none of the three incidents.** All three are *knowledge* failures. The token that killed eight subagents existed and was reachable; what was missing was the fact it was *unusable*. The Telegram material was intact on the old host; what was lost was the *knowledge that this was the working path*. The `~/.bashrc` token works fine; the defect is that nobody *declared* it. A design that adds exposure while fixing none of the motivating failures is a strictly worse trade.
2. **Blast radius follows replication, and these rows are built to replicate.** Four clones of the cards store exist now. A secret in a synced table is a secret in every store the row reaches, in each of their WALs — postgres never overwrites in place — and in every base backup, permanently. Rotation cannot retract it.
3. **It would silently repeal the two-tier invariant.** `assert_access_only` guards ONE rail (the ssh push). A table holding material opens a second, quieter rail: `assert_access_only` raises; `INSERT` does not.
4. **A verified push already exists.** Replacing `keepalive_push` (HTTP 200 verified, 0600, no-downgrade) with an unverified pull is a downgrade sold as a feature.

**Enforced, not conventional.** `_material.py` scans every row on the only path into the store — by field name *and* by value shape (provider prefixes, JWT, PEM, Telegram, conservative high-entropy) — and refuses the write. The refusal **never quotes the value**; that message reaches transcripts, and there is a test asserting it.

## Schema — sync columns from creation

Three tables, all carrying the five ADR-0022 §5.1 columns (`row_uuid`, `origin_node`, `revision`, `updated_at`, `deleted_at`).

| Table | Holds | Conflict class |
|---|---|---|
| `credential_descriptor` | what exists, who owns it | `shared_mutable` |
| `credential_placement` | where it should be | `shared_mutable` |
| `credential_observation` | what was actually seen | `log` |

Descriptor/placement are `shared_mutable` because primacy **handover** is exactly the edit that must never auto-merge: blind LWW would let a clock-skewed replica seize primacy and create two refreshers. Rule 4 applies — accept only a higher revision *from the same origin*, else DO NOTHING + divergence record. Observation is `log` with `origin_node` in the PK, so a collision provably means the identical row arrived twice.

`generation` (material version) is deliberately separate from `revision` (row version): conflating them makes a typo fix indistinguishable from a rotation.

## The base-table helper ADR-0022 §7.3 says is missing

`_contract.py` emits the five columns and **refuses** a `CREATE TABLE` that lacks them, mistypes them, makes the tombstone non-nullable, or (log class) omits `origin_node` from the PK. Fails closed — an unparseable DDL is rejected. Generic and credential-agnostic, so any consumer stream can use it.

## CR-001, made checkable

Today the refresh holder is **inferred** from whether a file happens to contain a field, so a second holder looks exactly like the first — an inference cannot contradict itself. Declaring `primary_node` makes it a fact that can be **wrong**:

- fleet half — `CR001_MULTIPLE_PRIMARIES_SQL`
- disk half — declared role vs observed refresh material → `EXTRA_REFRESHER` / `NO_REFRESHER`

## The materialize path

`sac creds status` / `check` / `declare`. Every fault carries a remedy and a non-zero exit. For a `primary_secret` on a non-primary node the remedy **refuses to materialize** and names what to mint instead — that refusal is the two-tier model holding, not a gap.

Observation is presence-only: never the value, never its validity, because probing whether a refresh token still works is a **write** (a rejected refresh clears the field).

## Real output, on this host

```
credential state on scitex-compute-04:
  [WARN ] anthropic-oauth:live: WORLD_READABLE
        mode 0664 — readable beyond its owner
        -> chmod 600 on the file named by 'file:/home/agent/.claude/.credentials.json'
  [FAULT] telegram-bot:slot-demo: ABSENT
        required here but environment variable CCT_BOT_TOKEN_DEMO_ABSENT is unset or empty
        -> materialize from primary 'ywata-note-win': sac accounts keepalive --to $(hostname)
```

Two genuine findings on first real run: the live artifact is **mode 0664**, and this node **holds refresh material** while fleet doctrine records compute-04 as stripped to access-only. (Measured from inside the container — see "not verified" below.)

## Incidental refactor

`cli_pkg/_main.py` was 521 lines against a 512 cap, so registering `creds` was blocked. Its one unrelated responsibility — cache-based shell-completion install, the only part reaching for `subprocess`/`os`/`~/.bashrc` — moved to `_main_completion.py`. Public API unchanged; `_main.py` is now 420 lines.

## Tests

`2749 passed, 1 skipped, 16 deselected` (cli_pkg + _credstate). The 155 `_credstate` tests include 17 run against the **real postgres on 55432**, which caught an `AmbiguousParameter` that a fake connection would have reported green.

## Not verified

- **Sync is not implemented** — the columns and rules exist so it *can* be; the primitive lives in `scitex_dev.state`, which does not exist yet.
- `_credstate/_store.py` is a **temporary home** written to ADR-0022 §4's shape so it lifts out unchanged.
- No production database is created; the DSN is never defaulted.
- The refresh-material finding was measured **from inside a container**, whose `$HOME` and mounts are not the bare host's. Whether the compute-04 *host* holds refresh material is a separate question this run cannot answer.
- Nothing was tested against a live credential refresh — deliberately. A failed refresh zeroes the file in place, so no probe of validity was performed anywhere.


ywatanabe1989:
**Branch updated to `develop`. Your `test_audit_all_clean` failure is this branch's own.**

Twelve PRs were failing `test_audit_all_clean` with the identical assertion text; the underlying rule differed per PR. Yours is:

```
[E] [PS-210 §2 dev-extras-incomplete]
pyproject.toml: `[state]` extra declares `psycopg` and tests import it
```

The `state = [...]` extra exists only on this branch — `develop`'s `pyproject.toml` has no such block — so this is introduced here and updating the branch will not clear it. The audit wants the extra completed so an install from it can actually run the tests that import `psycopg`.

ywatanabe1989:

PR #985, opened 2026-08-12
feat(auto-merge): a hold label the sweep honours, and merges that admit they were automated

ywatanabe1989:
## What this changes, in words

Two things, both in `.github/workflows/auto-merge-to-develop.yaml` — the cron sweep that merges green PRs into `develop` with `gh pr merge --admin`:

1. **You can now stop it.** Put a `hold` or `do-not-merge` label on a PR (or make it a draft) and the sweep refuses to merge it, and says so out loud in its log, on every tick. Remove the label and it merges again on the next tick. Nothing else to do.
2. **When it does merge, it says so on the PR first.** It posts a comment naming itself and listing why it merged, *before* merging. If that comment can't be posted, it doesn't merge.

Plus `CONTRIBUTING.md`, which claimed PRs land "reviewer-approved" — there is no reviewer — and now documents the sweep and the hold markers.

## Why

On 2026-08-12 a pull request that an agent had **deliberately held** — it relaxed a guard the operator had escalated, and the agent judged a human had to look first — was merged anyway at 00:29Z, unreviewed and uncommented. Two independent defects made that possible:

**The hold existed only in one agent's head.** Nothing in the repository knew a hold had been declared, so nothing in the repository could honour it. An intention that no mechanism reads is not a hold; it is a hope. The fix is a marker the automation actually reads.

**An automated merge was indistinguishable from a human one.** Every agent, every workflow and the operator act through the same GitHub account, so `merged_by` carries no information — it cannot separate "a watcher merged this the moment CI went green" from "a person read this diff". The agent whose hold was overridden could not find out who had overridden it either. The fix is a trace that does not depend on identity at all.

## A. The hold

* Marker is a **label** — `hold` or `do-not-merge`, listed in a `HOLD_LABELS` env var — because the person most likely to need this is the operator, and he must not have to edit YAML to stop a robot. Both labels now exist in the repo, so they are one click in the GitHub UI. **Draft** status counts too: it was already honoured before this change, but *silently*, which is half a mechanism.
* **Loud, every tick.** The refusal is a `::warning::` annotation naming the PR and which marker held it. Not once — every pass. A one-shot silent skip reproduces the original bug in a new place: the hold works, nobody can tell that it worked, and the next reader assumes it did not.
* **No noise on the PR itself.** The refusal lives in the run log, which is per-run and ephemeral. On a 15-minute tick, a hold that commented would add ~96 comments a day to the PR it is protecting.
* **Checked before CI state**, so a held PR reports as held even while it is red — "why didn't this merge" has one answer, not two.
* **Costs no merge budget.** `MAX_MERGES` is 1 per run; if a hold spent that budget, holding one PR would silently stall every PR behind it.
* **Unreadable labels count as held.** Same three-states-not-two rule the develop-health gate already follows: "I could not read it" is not "there is nothing there".

## B. The attribution

* `gh pr comment` runs **before** `gh pr merge`. Before, not after, so a merge can never exist without its explanation.
* The comment names the workflow, lists the conditions that let it merge (base, no hold marker, green non-advisory checks, `develop` settled and green, `mergeStateStatus`), states plainly that no human read the diff, explains **why `merged_by` cannot tell you that**, and tells the reader how to stop the next one.
* **If the comment fails, nothing merges.** The PR stays open, the failure is counted, and the run exits red. Merging anyway would trade a 15-minute delay for an untraceable merge — the exact defect the comment exists to close.
* The comment carries an invisible marker, so a merge refused by branch policy is not re-attributed on every subsequent tick.
* A `dry_run` dispatch stays side-effect free: the DRY_RUN branch is taken before the comment.

**Not done here, on purpose:** per-agent GitHub identities. That would fix attribution at the source, it touches credentials on every machine, and it is the operator's decision. No accounts created, no credentials touched.

## Verification

`bash -n` on the extracted step script and `yaml.safe_load` on the file both pass. **58 tests pass** — 41 new, plus the 17 pre-existing in `tests/develop/test_auto_merge_dispatch.py`, which still find exactly one merge step.

The hold tests are **not** string-matching the YAML: they extract the shipped hold blocks verbatim from the real workflow by their own anchors and **execute them under bash**, supplying only the label string `gh` would have returned. The decision under test is the decision that ships. The ordering properties (hold before merge, hold before budget, hold before CI, comment before merge, DRY_RUN before comment) are asserted against the same real file — a check that runs after the merge is not a gate.

It was also run for real: the modified step script was executed locally against this repository with `DRY_RUN=true`, using the live GitHub API, with a held PR present. See the PR comments for that output. No PR was merged, and no other agent's PR was touched.

## Known gap, reported not fixed

`.github/workflows/autobump-release-sweep.yaml` is a **second** `gh pr merge --admin` path. It merges only the `autobump/release-v*` bump PR that it creates itself, seconds after creating it, so a hold label is not reachable there by a human — and its PR body already names the automation, so it is not anonymous the way a swept merge was. Left alone deliberately: it is the release path, and the risk of touching it exceeds the value.

**This file is read from the default branch, so none of this takes effect until `develop` is promoted to `main`.**


ywatanabe1989:
**Merged by automation. No human read this diff.**

`auto-merge-to-develop` (`.github/workflows/auto-merge-to-develop.yaml` in this repo) is merging this pull request now — before you look for a reviewer, there wasn't one.

Why it merged:

- base branch is `develop` (the only base this sweep touches);
- no hold marker: not a draft, and none of the labels `hold do-not-merge`;
- every non-advisory check on the PR head is green (codecov and readthedocs are advisory and ignored);
- `develop`'s own head was settled and green when the sweep ran, so this lands on a base that was known good;
- `mergeStateStatus` = `CLEAN`.

This comment exists because the merging IDENTITY cannot tell you any of that. Every agent, workflow and human in this fleet acts through the same GitHub account, so `merged_by` on this PR is the same name whether a watcher merged it on a green check or a person merged it after reading it. If you are reading this months later: it was the watcher.

To stop an automated merge, add a `hold` (or `do-not-merge`) label, or convert the PR to a draft. The sweep refuses a held PR on every tick and says so in its log; removing the marker lets it merge again.

Sweep run: https://github.com/scitex-ai/scitex-agent-container/actions/runs/local-verification-of-pr-985

ywatanabe1989:
## About the comment above, and the evidence for this PR

**The attribution comment above was produced by a verification run of this PR's own change. Nothing was merged.** It is real output from the workflow's own code, posted through `gh pr comment` against the live API, kept here on purpose so the artifact is visible rather than described. Only the attribution block was executed — never the `gh pr merge` that follows it.

Everything below ran against the **live GitHub API**, driving the **shipped shell** extracted verbatim from the modified `auto-merge-to-develop.yaml`. No other agent's PR was held, merged, or written to.

### A hold is honoured, and stays honoured

`hold` label on this PR, three consecutive polls — exactly what the 15-minute cron does:

```
### comments on PR 985 BEFORE the three polls:
0
===== poll pass 1 =====
::warning::#985 is HELD (label:hold) — auto-merge refuses to merge it. This is not a one-off skip: it will be refused on every tick until the marker is removed, and it costs no merge budget. Remove the label (or mark the PR ready for review) and it merges normally on the next tick, with no other intervention.
--- merges(dry)=0 attrib_failures=0 ---
===== poll pass 2 =====
::warning::#985 is HELD (label:hold) — auto-merge refuses to merge it. ...
--- merges(dry)=0 attrib_failures=0 ---
===== poll pass 3 =====
::warning::#985 is HELD (label:hold) — auto-merge refuses to merge it. ...
--- merges(dry)=0 attrib_failures=0 ---
### comments on PR 985 AFTER the three polls:
0
```

Held on every pass, not once. The merge budget stays at 0, so a held PR delays nobody. Comment count went 0 → 0: the refusal is loud in the automation's log and silent on the PR, which is what keeps it from burying the PR it protects.

### Removing the label is sufficient on its own

```
### PR 985 labels + check rollup BEFORE removing the hold:
{"labels":["hold"],"not_green":0}
### removing the hold label:
### PR 985 labels AFTER:
[]
===== same loop, hold removed =====
DRY RUN: would post the automation-attribution comment on #985 and merge it -> develop (mergeStateStatus=BLOCKED)
--- merges(dry)=1 attrib_failures=0 ---
### comments on PR 985 after the unheld dry run (must still be 0 — DRY_RUN writes nothing):
0
```

Same loop, same PR, same green checks (`not_green: 0`). The only change is the label, and the PR goes from refused to merge candidate. Note the dry run wrote nothing: `workflow_dispatch` with `dry_run: true` stays side-effect free.

### The attribution posts once, not once per tick

```
### comments on 985 before: 0
===== attribution attempt 1 =====
https://github.com/scitex-ai/scitex-agent-container/pull/985#issuecomment-5260838741
===== attribution attempt 2 =====
#985 already carries this automation's attribution comment from an earlier tick — merging without posting a second one.
### comments on 985 after two attempts: 1
### comments carrying the attribution marker: 1
```

A merge refused by branch policy leaves the comment already posted; the next tick has to recognise it rather than stack another.

### The develop-health gate still fails safe

A full run of the modified step against the live repo stopped before touching any PR:

```
develop head: 8ee32a2c6ca9eba44cb8c92c1847483990d5252d
develop CI still in flight (12 running) — the next tick is the wait. Merging nothing.
```

Unchanged behaviour, confirmed still working after the edit.

### Static + unit

`yaml.safe_load` parses the workflow; `bash -n` on the extracted step passes. **58 tests pass** — 41 new, plus the 17 pre-existing in `tests/develop/test_auto_merge_dispatch.py`, which still find exactly one merge step.

### What is NOT verified here

The `gh pr merge` call itself was never executed, because the only honest way to test that is to merge a real PR. The merge line is unchanged by this PR; what changed is what runs *before* it.

Nor can anything here prove the behaviour on GitHub's runners until this file reaches `main` — the workflow is read from the default branch, so it takes effect only on promotion.

ywatanabe1989:
## Correction to my own evidence, and a separate pre-existing defect it exposed

Above I wrote that this PR's checks were green (`not_green: 0`) when the unheld run reported it as a merge candidate. **That number was wrong, and it was wrong for a reason worth reporting.** All nine checks were `QUEUED` at that moment — not one test had run.

The hold evidence stands unchanged: same PR, same rollup, same loop, and the only difference between "refused" and "merge candidate" was the label. But the *reason* the sweep considered it green is a **pre-existing bug in the greenness predicate**, which this PR does not touch and does not fix.

### The bug

```
[ .statusCheckRollup[] | select(...advisory...)
  | (.conclusion // .state // "")
  | select(. != "SUCCESS" and . != "NEUTRAL" and . != "SKIPPED" and . != "") ] | length
```

`gh` returns `conclusion` as an **empty string, not null**, for a CheckRun that is `QUEUED` or `IN_PROGRESS`. In jq `""` is truthy, so `.conclusion // .state` evaluates to `""` — and the trailing `. != ""` then excludes it. A check that has not started is counted as neither failing nor pending. It is counted as green.

Measured here, on this PR:

```
{"__typename":"CheckRun","conclusion":"","name":"rtd-sphinx-build-on-ubuntu-latest","status":"QUEUED",...}
{"__typename":"CheckRun","conclusion":"","name":"import-smoke-on-ubuntu-py3-12","status":"QUEUED",...}
{"__typename":"CheckRun","conclusion":"","name":"ruff-on-ubuntu-latest","status":"QUEUED",...}

=== what the shipped pending filter counts ===
0
```

Nine queued checks, zero counted. A PR pushed shortly before a cron tick can therefore be merged with `gh pr merge --admin` — which bypasses exactly the `BLOCKED` branch protection that would otherwise have held it — before any of its CI has run. This is the "a subtle bug merges a red PR" case this workflow's own header warns about.

The develop-health gate higher in the same file gets this right: it reads `.status != "completed"` into a separate `dev_busy` count and waits. The per-PR filter needs the same three states — completed-and-good is green, completed-and-bad is red, **not completed is pending**.

### Why it is not fixed here

This PR is about hold markers and merge attribution. The greenness predicate is the single highest-risk line in the file, and folding a change to it into this one would mean shipping the two together and reviewing neither. Filed separately as `auto-merge-queued-checks-read-as-green`.

I would rather report this and be wrong about the priority than fix it quietly inside a PR nobody expected it in.

PR #987, opened 2026-08-12
feat(cct-rail): an agent that comes up mute has to say so, over somebody else's voice

ywatanabe1989:
## The failure

`prune_tokenless_telegrammer_mcp` removes the `claude-code-telegrammer` MCP
server when no `CCT_BOT_TOKEN_<SLOT>` resolves. That is correct and deliberate
(operator ruling, card `sac-omit-telegram-mcp-when-no-cct-bot-token-20260702`).

But it removes the rail in **both directions** at the one moment nothing can
report it. The agent starts perfectly, reports healthy, and cannot even
self-diagnose — `health` is a tool on the server that just went away. The
2026-08-12 outage was found by the operator noticing silence.

## Measured, not assumed

`sac agents cct-audit` run on compute-04 **through the listen daemon's own
environment** (the vantage point a daemon-started agent actually has):

```
81 agent(s) considered — 66 DOWN, 0 UNKNOWN
```

The "did you mean" column named four live mismatches, **two of which nobody had
reported**:

| agent | derives | pool has |
|---|---|---|
| `neurovista` | `NEUROVISTA` | `PAPER_NEUROVISTA` |
| `neurovista-paper-writer` | `NEUROVISTA_PAPER_WRITER` | `PAPER_NEUROVISTA_WRITER` (word order) |
| `scitex-clew` | `SCITEX_CLEW`, `CLEW` | `PAPER_SCITEX_CLEW` (**new**) |
| `spartan-dev` | `SPARTAN_DEV` | `DEV` (**new**) |

Both neurovista agents are real registered specs on ywata-note-win, both
declare the channel, neither carries a `CCT_BOT_TOKEN_SLOT` override. Genuine
defects.

`_slot_candidates` is untouched and no pool key is renamed — which side should
move is the operator's decision on
`sac-cct-token-slot-mismatch-and-env-fold-20260812`. Near misses are shown to a
**human** and never acted on; resolving through them is how one agent takes
another's bot.

## Refuse to start, or start and shout?

**Start and shout.** Refusing would refuse 66 of 81 agents — the channel request
is inherited from the three spec templates as scaffolding. Telegram is a comms
rail, not a boot dependency (already an operator ruling: this log was demoted
ERROR→WARNING because a brand-new agent has no bot by definition). And refusing
makes the silence *worse*: a stranded agent cannot do its non-Telegram work and
the one process that could report the problem is gone.

## Where the alarm lands

Not a log line — that was tried and measured (2026-08-10: four agents mute
behind one INFO line each while the operator concluded they were ignoring him).
Two existing sac rails, five sibling alarm modules already use them:

- **The record** — `emit_subject_verdicts` under subsystem `cct-rail`. Durable,
  sac-owned, three-valued at the source, transition-tracked.
- **The push** — `push_to_lead(kind="blocker")` (ADR-0013). **This is the
  load-bearing choice: it reaches the operator through the LEAD's Telegram — a
  different agent with a different bot. A mute agent shouts with somebody
  else's voice.** The shout cannot ride the rail that is broken.

A board card was the other candidate and was rejected: sac does not depend on
scitex-cards, and the fix for "the operator was not told" must not itself add a
cross-package dependency that can be absent exactly when it is needed.

The **push** is gated on evidence somebody meant this agent to have a bot (a
declared slot, a near miss, or a rail that worked here before). Paging all 66
would rebuild the ignored alert channel the prune was written to remove. Every
alarming verdict is still **recorded**, and `cct-audit` lists all of them.

## Three-valued

`_secret_pool.read_pool` now reports whether a MISS is **conclusive**. A read
that sourced no secret FILE holds only the launching process env — which can
prove a slot present but never absent. That flag is the outage's root cause in
one bit; the operator's own correction is the spec:

> 「04 にトークンが無い」と私は言ったが誤り。**起動プロセスに無かった**が正しい。この区別がバグそのもの。

A pool that reads clean but holds **no `CCT_BOT_TOKEN_*` at all** is likewise
UNKNOWN — 80 confident false alarms are indistinguishable from a broken fleet.

## The fix is named in every message

```yaml
spec:
  apptainer:
    env:
      CCT_BOT_TOKEN_SLOT: <SLOT>
```

Precedence #2, and the only route that survives a relocation — the `.envrc`-folded
`$HOME/.env` does not travel, which is the other half of the 2026-08-12 root cause.

## Secrets

No token value is read, logged, or transmitted anywhere on this path. Presence
only (`bool(str(v).strip())`); slot NAMES and pool source PATHS at most. Tests
assert the absence of a value-shaped string in the verdict `repr`, the lead
push, the JSON output and the console table.

## Tests

```
81 passed in 0.85s
```

Wide regression over `_lifecycle`, `runtimes`, `cli_pkg`, `_events` and the
cross-package import gate: **6395 passed**. 3 failures + 19 errors in
`test__sdk_common.py` / `test__mcp_spec_env.py` are **pre-existing** — verified
byte-identical against the base checkout with none of these changes applied.
Not touched here; they deserve their own PR.

## Also in this PR

`_start_supervision.py` is a straight extraction of the two post-start daemon
threads from `_start.py`, which the new call site would otherwise push past the
512-line cap — the same split that already produced `_start_announce` and
`_start_preflight`.

The docs land as a new skill leaf, `23_telegram-rail-verdict.md`, beside its
companion the way `26_credentials-*` already sit together. Appending to
`23_telegram-integration.md` breached the SK-401 10240-byte budget, and
trimming it to fit left 12 bytes of headroom — not a margin. Split, they are
8607 and 7125 bytes. The stale claim in skill 23 that slot candidates come from
the workdir basename is also corrected; that behaviour was removed in 2026-07
as a bot-theft fix.

`cct-audit` takes `--agents-dir` so a caller states which spec tree it means
(useful for a peer's synced tree) — and so the tests say it directly instead of
intercepting `$SCITEX_DIR` and hoping every layer that resolves the state root
agrees.

Card: `sac-cct-rail-loud-when-no-slot-resolves-20260812`

https://claude.ai/code/session_01NcDz9gLL2Ct24FXqX2SF9T


ywatanabe1989:
**Unstuck: this PR was `DIRTY`, so it could not run CI at all** — one green `CLAssistant` and zero test jobs, which is easy to misread as "queued behind a busy pool". Merged `origin/develop` in (merge, not rebase — force-push is barred by hook).

Single conflict, `CHANGELOG.md`: this branch's `Unreleased / Added` entry (the cct-rail mute-agent alarm) collided with the entries `develop` has accumulated since the branch point. **Both sides kept in full**, under the same `Added` heading, this branch's entry first. No code was touched and no behaviour changed.

Verified after pushing: `mergeStateStatus` `DIRTY → BLOCKED`, `mergeable` `CONFLICTING → MERGEABLE`, check count 1 → 8, with `ruff`, `rtd-sphinx-build`, `import-smoke`, `quality-audit` and `no-hosted-runners-guard` already green.

PR #988, opened 2026-08-12
feat(relocate-preflight): ask whether the spec HOLDS on the target, and say what would fix it

ywatanabe1989:
## What this changes

Relocation's `preflight` phase now answers one question properly — **does this
agent's spec actually hold on the TARGET host?** — and when it does not, the
refusal says what is missing, where it was checked, and what would fix it.

Everything from `source_drain` onward has already begun taking the agent off the
source, so a spec incompatibility found at `target_standby` is an outage while the
same one found at `preflight` is a clean refusal that costs nobody anything. All
of this is therefore in `preflight`, and none of it is spread into later phases.

Twelve checks become fifteen, and the refusal becomes a work list.

### Three checks that did not exist

- **`workdir_exists_on_target`** — `spec.workdir` becomes apptainer's `--pwd`.
  There is no `spec.repo` (the validator rejects the key); the workdir *is* the
  checkout. An absent one fails at boot, which under relocation means after the
  source has been stopped. Measured on a real host during this work:
  `/home/ywatanabe/proj/scitex-os` does not exist on scitex-compute-04, and
  nothing before this change would have said so.
- **`card_store_dsn_correct`** — judges the DSN as a string, before anything
  answers it. `5432` is wrong on every host in this fleet with no exceptions, and
  a *port-less* DSN is the same bug wearing no number, because libpq defaults it
  to 5432. This has to be caught by reading rather than by dialling: something
  answers on 5432 nearly everywhere, so the existing reachability check passes
  while the agent writes its cards into a database nobody reads. Confirmed on
  scitex-compute-04: `55432` is listening, `5432` is not listening at all.
- **`groups_resolvable_on_target`** — asks the target's own sac what it makes of
  this spec's `metadata.labels.groups`. An agent that arrives holding groups the
  target cannot resolve is refused by every group-gated call it makes.

### A missing bind is not one problem

Fifteen fleet specs bind paths absent on scitex-compute-04 and every one of those
specs is *correct for the machine it currently pins*. Nine are Spartan agents
binding shared cluster storage a workstation cannot provide at all; six are laptop
agents binding a dataset and a local checkout that exist on one machine because
that machine made them. Printed as "path not found" those are indistinguishable,
and the old hint — "remove or re-point these binds" — is right for one of them.

Each missing path is now classified and the hint split by what you have to do:

- **provision on the target** — host infrastructure (`/data`, `/mnt`, `/gpfs`, …).
  If the filesystem does not exist there at all, the hint says so: this spec
  belongs on a host that has it, and re-pointing the bind produces a *different
  agent*, not a relocated one.
- **must travel with the agent** — anything under or beside the agent's own
  workdir, and any dataset directory. A relocation carries the spec and the
  transcript and nothing else.
- **provision there, never copy** — credential paths (`.ssh`, `.config/gh`,
  `accounts/*/.credentials.json`, `.pgpass`). Decided first, precisely so key
  material never attracts "move it with the agent".

Where the path's shape cannot settle it the answer is `unclassified` and the hint
states both possibilities. A confident wrong category sends somebody to provision
a directory that should have travelled.

### Refused, or unable to ask?

Nine relocation probes were refused 403 on 2026-08-11 by the *local* listen daemon
being asked to broker the ssh — not by any target. That status was being flattened
into free text, losing the one bit that separates "I was refused before I could
ask" from "the target answered no". `ProbeTransportError` now keeps it, and the
message says in words that the refusal is about this container's authorization and
that **nothing about the target was measured**. A separate upgrade of those hosts
is in flight and is not duplicated here.

### Every problem in one pass, ordered by what to do

No check is skipped because an earlier one failed; all fifteen always run. Two
things then make a complete list readable:

- **Root causes are stated once.** An unreachable target turns eleven checks
  UNKNOWN, and each carries the *same* probe-error text — so identical reasons are
  recognised as one cause and printed once, with the blocked checks named under
  it. They still block; they are simply not eleven tasks. (Verified live: eight
  unknowns collapsed to one `ONE ROOT CAUSE` line.)
- **The rest are ordered by action** — provision on the target, must travel with
  the agent, correct the spec, decide, go and measure — rather than by check
  index, which is the code's structure leaking into the operator's afternoon.

### Several agents in one read-only sweep

`sac agents relocate a b c --to host` preflights each and prints a combined
summary, so the shape of a queued batch is visible before anything is touched.
`--no-dry-run` still takes exactly one name and refuses a batch outright: the
journal that makes a relocation resumable is per-agent, so one interruption would
leave several agents half-moved.

## The three-valued rule

Unchanged in spirit and now enforced per check: `pass` / `fail` / `unknown`, where
`unknown` blocks and is never reported as either neighbour. What is new is that
**the rollup policy is named at the aggregation site** — `UNKNOWN_BLOCKS_RELOCATION`
— rather than baked into each predicate. Whether an unknown refuses is a property
of *relocation*, not of the verdict type: a dashboard may paint the same unknown
amber and be right to.

The per-check record stays `{name, ok, detail, hint}` with a three-valued `ok` and
no boolean-plus-flag anywhere, so adopting the fleet-wide status type being defined
in scitex-dev is a rename rather than a rewrite.

Tests include a case **per check** where the probe itself failed, asserting the
verdict is `unknown`, that it is not counted as a failure, and that the overall
preflight refuses — plus a guard that fails if a new check is added without one.

## Structure

Five new modules, three of them extractions forced by the 512-line limit and worth
doing on their own terms:

| module | what it owns |
|---|---|
| `_relocate_bind_kind` | classify a missing path; pure path shapes, no I/O |
| `_relocate_checks_spec` | the three new predicates |
| `_relocate_plan` | root-cause grouping and the action-ordered work list |
| `_relocate_spec_reads` | where in a spec each fact lives |
| `cli_pkg/_relocate_preflight_one` | one agent's preflight as a *value*, which is what makes a sweep possible |

Preflight remains read-only: it reads a spec file, reads the state db, and runs one
batched read-only probe. It creates, moves and modifies nothing on either host.

### DIFF
--- a/docs/relocate.md
+++ b/docs/relocate.md
@@ -54,21 +54,36 @@ first one, so you do not have to run it N times to find N problems.
 sac agents relocate <name> --to <host> --dry-run
 ```
 
-### The eleven checks
+Several names preflight together and print a combined summary, so the shape of a
+queued batch is visible before anything is touched:
 
-| check | what it catches | the 2026-08-07 instance |
+```bash
+sac agents relocate scitex-dev scitex-hpc scitex-db --to scitex-compute-04
+```
+
+That form is read-only by construction. `--no-dry-run` takes exactly one name:
+the journal that makes a relocation resumable is per-agent, and a batch
+interrupted halfway would leave several agents half-moved across two hosts.
+
+### The fifteen checks
+
+| check | what it catches | the instance |
 |---|---|---|
 | `target_reachable` | host does not answer | — nothing else in the report means anything until this passes |
 | `image_present` | SIF absent on the target | a missing image fails at boot, **after** the lease has moved |
 | `binds_exist_on_target` | bind source paths absent there | the spec bound `/mnt/c`, a Windows drive that does not exist on the nas |
-| `card_store_reachable` | agent cannot reach its board | `SCITEX_CARDS_DB` is port 5432 here, 5442 there |
+| `workdir_exists_on_target` | `spec.workdir` absent there | it becomes apptainer's `--pwd`; there is no `spec.repo`, the workdir **is** the checkout |
+| `card_store_dsn_correct` | the DSN itself is wrong | `5432` is wrong on every host with no exceptions, and a port-less DSN means the same thing to libpq |
+| `card_store_reachable` | agent cannot reach its board | the fleet's endpoint is `55432`; an agent that cannot reach its board runs and records nothing |
 | `credentials_valid` | **validity, not presence** | the nas had a file expired 2026-05-23 with an empty `refreshToken`, and sac loaded it *in preference to* the good one — every turn 401'd while `sac agents health` still said healthy |
 | `runtime_supported` | target's sac rejects the runtime | nas's sac 0.21.9 rejects `tui` |
 | `spec_schema_accepted` | target's validator rejects a key | a top-level `provider:` key, same older validator |
 | `ports_free` | port already bound there | reassign in the spec before moving |
+| `groups_resolvable_on_target` | the target's sac cannot read this spec's group labels | three hosts resolve `[]` for every agent regardless of `spec.yaml`; nine relocation probes were refused 403 by exactly that on 2026-08-11 |
 | `hub_reachable_from_target` | hub unreachable **from there** | the nas's services bind `127.0.0.1`, so reaching them from HERE proves nothing about THERE |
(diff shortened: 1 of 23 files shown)


ywatanabe1989:
Superseded by #989, which carries the identical commit rebased onto the current develop. This one was opened from a base that had moved and GitHub reported it CONFLICTING; the branch was replaced rather than force-pushed.

PR #990, opened 2026-08-12
fix(auto-merge): a queued check is not a green check — and the sweep stops using --admin

ywatanabe1989:
> ## ⚠️ Approving this PR does not disarm anything
>
> **GitHub runs a scheduled workflow from the DEFAULT branch.** The copy actually executing is `main`'s — and `main` and `develop` currently hold **byte-identical** copies of this file. So a fix merged to `develop` would look complete while changing nothing: the sweep would go on running the buggy predicate out of `main`.
>
> This is the **fifth distinct instance tonight of merged-not-live**, and the first where the *fix for a hazard* would itself have been inert.
>
> The sweep is **disabled** right now, and stays disabled. The sequence is:
> **#985 → this PR → promote `develop` to `main` → only then re-enable.**
> Nothing is safe to re-enable at any earlier point in that chain.

---

## The defect

`.github/workflows/auto-merge-to-develop.yaml` decided whether a PR was green with:

```jq
| (.conclusion // .state // "")
| select(. != "SUCCESS" and . != "NEUTRAL" and . != "SKIPPED" and . != "")
```

`gh` reports `conclusion` as an **empty string** — not null, not absent — for a check that is `QUEUED` or `IN_PROGRESS`. So `select(. != "")` dropped exactly the checks meaning *we do not know yet*, and the pending count came back **0**.

Measured, on a verbatim capture of this repo's PR #985 while it carried 8 queued checks and 1 finished one (the fixture added here):

```
OLD filter answered pending = 0        # i.e. "fully green"
```

Combined with `gh pr merge --admin`, that merged unverified code **past the branch protection that would have refused it**. With the self-hosted pool saturated, the window is close to permanent rather than rare.

## It already fired — 6 early merges, 3 of them red

A read-only audit of **186** PRs merged into `develop` by the sweep (back to 2026-05-28; the same buggy filter was on `main` from at least 2026-06-05):

| verdict | count |
|---|---|
| `verified-at-merge` | 179 |
| `merged-early` | **6** — #316, #333, #334, #348, #389, #926 |
| `cannot-tell` | 1 — #374 |

**Three of the six landed RED**: #316, #333, #334 each had all three `pytest-matrix` legs still in progress at merge time, and all three **failed** minutes afterwards.

The cleanest single demonstration is **#926** (2026-08-09): `no-hosted-runners-guard-on-self-hosted` sat `QUEUED` for **46 minutes** waiting for the `scitex-ci` runner. The sweep's own log for that tick shows `PR #884 not green (1) — skip` and then `MERGED #926 -> develop` — with **no** "not green" line for #926. The check completed 4–5 s *after* `mergedAt`. It happened to pass.

`#374` is kept as `cannot-tell` rather than rounded to fine: its check's `completed_at` equals `mergedAt` **to the second**, and sub-second ordering is not recoverable from any API. Rounding one row to "fine" in a report about a bug that rounds unknown to fine would be its own joke.

Nothing is reverted here, and nothing should be on my say-so — a revert of an integrated change is more dangerous than the original early merge, and some of those PRs were fine on their merits regardless of when the checks finished.

## The fix — three states, named

A check is **passing**, **failing**, or **not yet known**. The old code had two buckets for three states, so the third fell into "green". Now all three are counted separately and the merge requires `failing` *and* `not-yet-known` to both be zero.

| reported state | bucket | why |
|---|---|---|
| `SUCCESS` | passing | |
| `NEUTRAL` | passing | GitHub's own "I ran, I am not blocking you"; branch protection counts it satisfied |
| `SKIPPED` | passing | see below — this one got a decision, not a default |
| `FAILURE` `ERROR` `TIMED_OUT` `ACTION_REQUIRED` `STARTUP_FAILURE` | failing | reported a negative verdict |
| `CANCELLED` `STALE` | **not yet known** | ran, but produced no verdict to trust. Not "failing" (nothing shown wrong), not "passing" (nothing shown right) |
| any `status` other than `COMPLETED` (`QUEUED`, `IN_PROGRESS`, `WAITING`, `REQUESTED`, `PENDING`) | **not yet known** | ← the bug |
| `COMPLETED` with an empty conclusion | **not yet known** | should not happen; we do not guess |
| `PENDING` / `EXPECTED` (StatusContext) | **not yet known** | |
| **anything unrecognised** | **not yet known** | the default arm — a state GitHub invents after this was written blocks a merge instead of passing one |
| **no non-advisory checks at all** | refuse, loudly | absence of evidence is not evidence of green |

### Why `SKIPPED` passes

Two reasons, and it is worth stating because `SKIPPED` genuinely means two different things.

1. GitHub's own branch protection treats a skipped required check as **satisfied**. Blocking on it would make this sweep **stricter than the protection it bypasses**, and would deadlock every PR carrying a legitimately inapplicable matrix leg. A sweep that never merges is its own outage — this file exists because of one (38 PRs stranded for 8 days).
2. The frightening reading — *"it silently did not happen"* — **does not arrive as `SKIPPED`**. A workflow that never ran contributes **no entry at all** to `statusCheckRollup`. That is *absence*, and absence now has its own explicit refusal. A `SKIPPED` entry is an affirmative report from CI that the leg did not apply.

**Revisit this** the day a *required* job becomes path-filtered: `skipped` would then start meaning "did not happen", and that arm would be wrong. Said in the workflow comment too.

### Why the classification is shell and not jq — this is not a style choice

The runner has neither `gh` nor `jq` (see the file's own header; the workflow bootstraps `gh` into `RUNNER_TEMP` for its own use). **A decision expressed in a language the executing machine cannot run can only be tested by re-implementing it — and then the test and the code are two guesses at the same thing, agreeing or disagreeing for reasons unrelated to whether either is right.** So the decision is shell, which any machine here can execute, and the tests below run *that*. jq's remaining job is a pure projection of four raw fields, interpreting nothing.

### Reporting

The verdict is logged as **counts per state**, always — not a table of checks. An empty conclusion renders as a blank cell and a blank cell reads to a human as "nothing wrong": the same misreading in the report that the filter made in the code. A check that reports no state at all is printed as `NO STATE REPORTED`, never as nothing.

## `--admin` is dropped — argued, not assumed

`--admin` is what turned this from "merges early" into "merges past the only thing that would have caught it". GitHub *would* have refused those merges.

Measured (`gh api repos/<repo>/branches/develop/protection`, 2026-08-12):

```
required_status_checks.contexts = [pytest-matrix-on-ubuntu-py3.11,
                                   pytest-matrix-on-ubuntu-py3.13]
required_approving_review_count = 0      <- no human review required
strict                          = false  <- no up-to-date/rebase demand
enforce_admins                  = false
```

So dropping it **costs a genuinely green PR nothing**: no review requirement to satisfy, no rebase demand, and both required contexts do report on every PR here — i.e. there is no never-reporting required check, which is the one honest reason to keep the flag. What it buys is independence: the sweep's predicate and GitHub's protection must now *agree* before anything lands, so one bad predicate is no longer sufficient to merge.

If anyone restores it, the workflow comment asks them to name the required check that forced it. A test asserts its absence with the same message.

## Testing — the shipped shell, executed

`tests/integration/test_ci_auto_merge_greenness.py` lifts the decision block **verbatim out of the workflow by its own anchors** and runs it under `bash`, exactly as #985's hold tests do. Only `rollup` is supplied — the value `gh` would have returned. Two fixtures are verbatim `gh pr view <n> --json statusCheckRollup` captures (PR #985 queued, merged PR #964 green), and a jq test asserts the shipped projection turns those payloads into the rows the decision tests use, so the two cannot drift.

**63 tests, all passing.** #985's 41 hold/attribution tests still pass unchanged.

**Those tests earned their keep before they ever ran in CI, and this is the argument for executing the shipped block rather than asserting on a YAML string.** They caught a real defect in the first draft of this fix: `IFS=$'\t' read -r a b c d` **collapses runs of tabs**, because tab is IFS *whitespace*, so an empty `conclusion` shifted every column left and a green `StatusContext` was read back as `status=SUCCESS`. A test that asserted against the YAML text would have shipped that — the same class of bug this PR exists to remove, reintroduced by the thing meant to prevent it. The four fields are now cut out explicitly.

### Live verification, through `gh`'s own embedded jq

| subject | shipped code | old filter, same PR, same moment |
|---|---|---|
| this PR with 8 queued checks | `1 passing, 0 failing, 8 not yet known` → **refused** | `pending = 0` → **would have merged** |
| merged PR #964, fully green | `10 passing, 0 failing, 0 not yet known` → **merges** | `pending = 0` → merges |

Full output in a comment below.

### Mutation check

Each mutation applied to the real file, suite re-run, file restored:

| mutation | result |
|---|---|
| baseline (shipped) | **63 passed** |
| **M1** — non-`COMPLETED` status returns `pass` (*the original bug*) | **16 failed / 47 passed** — every failure is queued/in-progress/waiting/requested/pending; `SKIPPED`, `NEUTRAL`, `FAILURE`, all-green and the projection tests all still pass |
| **M2** — the unrecognised catch-all returns `pass` | **10 failed / 53 passed** — exactly the empty-conclusion / `PENDING` / `EXPECTED` / `WORMHOLE_2031` family |
| **M3** — the literal pre-fix predicate and `--admin` restored | **61 failed / 2 passed** — the tests cannot pass vacuously against the old code |
| restored | **63 passed**, tree clean |

## Coordination with #985

**This is stacked on #985** and contains its two commits. #985 inserts the hold-marker block immediately *above* the predicate and the attribution block immediately *below* it, and rewrites the merge line I also touch — a `develop`-based edit would have collided there, and a hand-resolved conflict in exactly the code that decides whether to merge unverified code is not a risk worth taking. **Merge #985 first**; if it is closed rather than merged, this needs a rebase onto `develop`.

One small change inside #985's own text: the attribution comment now records the counts at merge time (`N passing, 0 failing, 0 not yet known`) and states that a queued check blocks rather than passes — the audit trail is the right place for the number the incident turned on.

## One loose thread, carded separately

`rtd-sphinx-build-on-ubuntu-latest` does **not** match the `codecov|readthedocs` advisory filter — its name contains neither string — so it blocks merges today. That errs safe, but only by accident: it is a gate nobody chose. Left alone in this PR and tracked separately rather than fixed in passing.


ywatanabe1989:
## Live verification — the shipped block, run against real PRs through `gh`'s own jq

Not the `jq` binary and not a fixture: the decision block was extracted verbatim from the workflow (137 lines, `gh pr view --jq` call included) and executed under `bash` against live PRs, so the engine evaluating the filter is the same embedded jq the runner would use (`gh 2.97.0`).

### 1. A real PR with queued checks — this PR, minutes after it was opened

```
  #990 NOT-YET-KNOWN rtd-sphinx-build-on-ubuntu-latest (status=QUEUED)
  #990 NOT-YET-KNOWN import-smoke-on-ubuntu-py3-12 (status=QUEUED)
  #990 NOT-YET-KNOWN ruff-on-ubuntu-latest (status=QUEUED)
  #990 NOT-YET-KNOWN no-hosted-runners-guard-on-self-hosted (status=QUEUED)
  #990 NOT-YET-KNOWN scitex-dev-quality-audit-on-ubuntu-latest (status=QUEUED)
  #990 NOT-YET-KNOWN pytest-matrix-on-ubuntu-py3.11 (status=QUEUED)
  #990 NOT-YET-KNOWN pytest-matrix-on-ubuntu-py3.12 (status=QUEUED)
  #990 NOT-YET-KNOWN pytest-matrix-on-ubuntu-py3.13 (status=QUEUED)
PR #990 checks: 1 passing, 0 failing, 8 not yet known (of 9 non-advisory; codecov/readthedocs excluded).
PR #990 is not green — skip (0 failing, 8 not yet known). NOT-YET-KNOWN IS A REFUSAL, NOT A PASS: a queued check has tested nothing.
```

**The same PR, at the same moment, through the same `gh`, with the OLD filter:**

```
OLD pending count = 0
  -> the OLD code would have called PR #990 GREEN and merged it.
```

That is the defect reproduced live and then refused live, side by side.

### 2. A real fully-green PR — merged #964

```
PR #964 checks: 10 passing, 0 failing, 0 not yet known (of 10 non-advisory; codecov/readthedocs excluded).
REACHED_MERGE  <-- the sweep WOULD have merged this
```

The fix does not stall the sweep. A green PR still merges.

### Note on the counts

`1 passing` on this PR is `CLAssistant`, which had finished. Eight legs were still queued — including both contexts `develop`'s branch protection requires (`pytest-matrix-on-ubuntu-py3.11` and `-py3.13`), which is precisely why dropping `--admin` gives a second, independent refusal for this exact situation.

PR #995, opened 2026-08-12
feat(hooks): deploy sac's own hook assets on every start, so a merged fix reaches agents

ywatanabe1989:
## The defect

`_baseline_assets/` holds the canonical, CI-self-tested hook implementations. The copies agents actually run came from the operator's dotfiles tree (`_shared/to_home/.claude/hooks/pre-tool-use/`), which is a **hand-maintained mirror** of that directory. Nothing connected the two, so merging a hook fix into sac armed nothing until a human remembered to re-copy it — and nothing anywhere reported that they had not.

This is the same shape as the bugs those hooks catch: a rule that exists, a fix that exists, and nothing that fires.

### Measured before the fix (compute-04)

| Asset | State |
|---|---|
| `enforce_telegram_use_lists.sh` | deployed copy 601 bytes behind the repo |
| `hpc_login_whitelist_{core,policy}.py`, `enforce_hpc_login_node_whitelist.sh` | all three behind — the deployed pair still passes the pre-rename `LOG_PATH`, not the namespaced `SCITEX_AGENT_CONTAINER_HOOK_LOG_PATH` that `test_hook_audit_log_env_contract.py` pins |
| `claude_worktree_hooks`, `git_identity_hooks`, `heavy_job_hooks` | **never copied at all** — every file absent |

## The fix — a missing layer, not a new mechanism

`deploy_to_home` already runs on **every start**. The packaged assets are now deployed as part of it: after the `to_home` walk, so the canonical version wins over the stale mirror; and before `_apply_host_merge_with_drift_guard`, so the real files are visible to its agent-layer-wins check (a same-named host hook is then correctly skipped rather than double-armed).

No packaging change was needed — `_baseline_assets` already ships in the wheel via `[tool.hatch.build.targets.wheel.force-include]`.

### Destinations are read, not assumed

Four families arm from `pre-tool-use`; `claude_worktree_hooks` arms from `.claude/hooks/claude_worktree_hooks/`. A flat deploy would have put three scripts at paths the registration does not point at — deployed and still dead. Each asset's destination is therefore parsed out of **its own family's settings fragment**, the same file that arms it, so registration and deployment cannot disagree. The parse is a regex over the fragment's raw text because the fragments use at least two different JSON shapes.

## Safety

A hook sits on the send path of the operator's only channel. An agent left with a broken hook cannot send, and therefore cannot report that it cannot send.

- **Atomic** — temp file + `chmod` + `os.replace`. A failed write aborts before the destination is touched; the old hook keeps running.
- **Validated** — a shell asset failing `bash -n` is refused, never deployed. Only files whose content actually changed are checked, so the steady state spawns no subprocesses.
- **Fail-open** — every error is caught per-file and recorded, never raised. A deploy failure can never be why an agent fails to start. One unwritable directory does not stop the others.
- **Loud once** — failures log at ERROR naming the file and the consequence.
- **Nothing deleted** — a replaced file is displaced to `.old/<timestamp>/`.

## Verification

Presence assertions would be vacuous on a machine where the hook is already present. The suite starts from a deployed copy that **differs** from the packaged asset and `pytest.fail`s if that precondition is not actually false. It then asserts the bytes converge **and** that the landed file executes: rc=2 on a bare `#162`, rc=0 on a compliant message, and its own `--self-test` green.

**Mutation-checked** (29 tests):

| Mutation | Result |
|---|---|
| remove the `deploy_baseline_hook_assets` call from `deploy_to_home` | exactly the 2 `TestDeployToHomeWiring` tests fail |
| flatten all destinations back to `pre-tool-use` | exactly the 3 placement tests fail |
| restored | 29 passed |

### No hook is weakened

The one drift that looked like a host-local fix (`enforce_telegram_use_lists.sh`, whose deployed header reads `rule 3 + slash-enum gap fix`) was probed behaviourally rather than assumed: both versions block slash-enum (rc=2), block comma-enum (rc=2), allow URLs and POSIX paths (rc=0), and both self-tests report `pass=13 fail=0`. The difference is comments and shell formatting only.

## Follow-up, deliberately not in this PR

The three never-copied families will now have their **scripts** deployed, but arming them means splicing their `settings.local.json.fragment.json` into the settings cascade. That turns on guards that have never run fleet-wide (e.g. `enforce_commit_author_allowlist.sh`), which is a behaviour change with real blast radius and deserves its own PR. `TestArmedHooksLandWhereTheyAreArmed` already pins that whatever *is* armed lands where it is armed.


ywatanabe1989:
**Branch updated to `develop`; 5 of 6 failures were inherited, 1 is yours.**

This PR's last run showed 6 failures. Five of them belonged to `develop` at the time that run was created, not to this branch:

- `test_audit_all_clean` → `[PS-226 §1 job-name-not-hyphenated]`, since declared as a deferral on develop by b70feb3 (#1009)
- `test__dev_jobs.py` ×3 → fixed on develop by c10eed0 (#1007)
- `test_host_group.py::test_library_warning_reaches_the_merged_cli_runner_output` → fixed on develop by b9bccd2 (#1004)

I ran update-branch. **Measured after the update: 6 failures → 1.** The remaining one is this branch's own:

```
FAILED tests/integration/test_baseline_hook_asset_deploy.py::TestDeployedHookActuallyRuns::test_deployed_hook_allows_a_compliant_message - assert 2 == 0
```

The deployed hook exits 2 on a message the test stages as compliant. Setup log shows all 15 assets deploying fine, so this is the hook's verdict, not a deploy failure. Worth checking against 374e67c (#1017), which changed the `#NNN` telegram rule — it is now in your base and did **not** resolve this.

(Posted by an automated PR-backlog sweep. No other change was made to this branch.)

ywatanabe1989:

PR #996, opened 2026-08-12
fix(pg): a new card-store cluster must be born closed on the unix socket

ywatanabe1989:
## The hole

`~/.scitex/pg/setup_cluster.sh` ran `initdb --auth-local=trust`. Every agent
container binds `/home/ywatanabe`, the socket is `srwxrwxrwx`, every agent runs
as uid 1000, and the login role is the bootstrap superuser — so **any agent
could connect as any role with no password and drop the board**.

Measured open tonight on **all four** compute hosts. compute-01 and -02 had
been reported as already using `scram-sha-256`; that was true of their `host`
(TCP) lines only — both `local` lines were `trust`.

On **compute-03** it was strictly worse: `scitex_cards` is not superuser there,
but `ywatanabe` is (`rolsuper=t, createrole=t, bypassrls=t`), and it was
reachable over the socket with no credential at all — unauthenticated privilege
escalation, demonstrated and then re-probed as closed.

The `initdb` call is guarded by PGDATA absence and `start_pg.sh` never rewrites
`pg_hba.conf`, so closing the running clusters did **not** close the script.
Every newly provisioned cluster was born with the hole again.

## Why one word was not enough

`--auth-local=scram-sha-256` on its own locks every socket user out. The
`.pgpass` half carries two traps, both hit for real:

* **libpq rewrites a socket host to `localhost` only when the socket directory
  equals its compiled-in default.** This cluster's is not, so the lookup key is
  the **literal path**. A `localhost` line pre-flights fine over TCP and still
  fails on the socket — how compute-04 broke.
* **A db-specific entry does not cover a connection to another database.**
  compute-03 carried `db=scitex_cards` lines and the admin path to `db=postgres`
  failed with `no password supplied` — which is how this PR's dedupe rule got
  its wildcard requirement.

## What changed

`scripts/pg/setup_cluster.sh` (version-controlled; the loose copy at
`~/.scitex/pg/setup_cluster.sh` was displaced to `.old/<ts>/` and replaced with
this one, byte-identical):

* `--auth-local=scram-sha-256`
* provisions `.pgpass` for all three keys — socket path, `localhost`,
  `127.0.0.1` — with a **wildcard database**, append-only, mode 0600
* sources the password from the canonical `:55432` entry instead of the
  ADR-0022-banned `:5432`, which simply does not exist on hosts provisioned
  after the ban (the old lookup failed outright there); mints one when absent
* **refuses to hand back a cluster** that still carries a `local … trust` rule
* parameterised via `SCITEX_PG_{ROOT,SIF,PORT,ROLE}` / `SCITEX_PGPASS` so it is
  testable; defaults preserve existing behaviour
* a password is never echoed, logged, or passed on a command line

## Verification

`tests/integration/test_pg_cluster_provisioning.py` stands up a **real**
throwaway cluster (no mocks — the failure mode is a libpq lookup-key rule that
no amount of reading the file can confirm) and asserts both halves plus a
wrong-password mutation control.

**The test was verified to go red.** With `--auth-local=trust` restored:

* guard intact → 12 errors (the script refuses to return an open cluster)
* guard neutered → **5 failed, 7 passed**, including
  `AssertionError: a wrong password was accepted over the socket` and
  `a 'local ... trust' rule survived provisioning`
* restored → **12 passed**

Gated by capability (apptainer + SIF), deliberately **not** by the `integration`
marker, which the default `addopts` deselects — a security regression test that
never runs is not a test.

## Fleet state after the change

Applied by **reload only, never restart**, each flip gated on a pre-flight that
proved the `.pgpass` key resolves over the socket first, and validated through
`pg_hba_file_rules` before going live. `pg_hba.conf` and `.pgpass` backed up to
`~/.scitex/pg/.old/<ts>/` on every host; nothing deleted.

| host | `local all` / `local replication` | no password | with `.pgpass` | wrong password | board |
|---|---|---|---|---|---|
| compute-01 | scram-sha-256 | refused | connects | rejected | reads; write proven (rolled back) |
| compute-02 | scram-sha-256 | refused | connects | rejected | reads; write proven (rolled back) |
| compute-03 | scram-sha-256 | refused | connects | rejected | 16 tables, 3719 rows; write proven |
| compute-04 | scram-sha-256 | refused | connects | rejected | verified from inside a container |

Role privileges were not touched.


ywatanabe1989:
**Branch updated to `develop`; 4 of 5 failures were inherited, 1 is yours.**

`test_audit_all_clean` on this PR was reporting **two unrelated things at once**, which is why the check name alone was misleading. The `[PS-226 §1 job-name-not-hyphenated]` half belonged to develop and is now declared as a deferral by b70feb3 (#1009). The `test__dev_jobs.py` ×3 and `test_host_group.py` failures were likewise develop's, fixed by c10eed0 (#1007) and b9bccd2 (#1004).

I ran update-branch. **Measured after the update: 5 failures → 1**, and the survivor is entirely this branch's own:

```
[PA-307 §3 test-quality] tests/integration/test_pg_cluster_provisioning.py:85:  STX-TQ004
[PA-307 §3 test-quality] tests/integration/test_pg_cluster_provisioning.py:150: STX-TQ004
[PA-307 §3 test-quality] tests/integration/test_pg_cluster_provisioning.py:165: STX-TQ004
  Fixture is scope="session"/"module" AND mutates state inside it
```

Three session/module-scoped fixtures that mutate state. `test_pg_cluster_provisioning.py` is new on this branch, so these are new violations; narrowing the scope (or moving the mutation out of the fixture) should clear the audit.

(Posted by an automated PR-backlog sweep. No other change was made to this branch.)

ywatanabe1989:

PR #998, opened 2026-08-12
ci(nightly): give the SIF nightly an alarm, because it had none

ywatanabe1989:
# The nightly had no alarm. That is why it was red for seven nights and nobody knew.

Measured 2026-08-12. `main` @`5532e6b` failed **this workflow's** nightly cron on
Aug 5, 6, 7, 8, 9, 10 and 11 — every leg, **zero tests executed**, each time
dying at:

```
mkdir: cannot create directory ‘/data’: Permission denied
##[error]Process completed with exit code 1.
```

Nobody was told. The workflow had three steps — run, upload coverage, clean
scratch — and **not one line of failure reporting anywhere in it**.

A drift detector whose alarm does not exist is not a weak detector; it is a dead
one, and its seven-day silence is evidence about the detector rather than about
the code.

## Why the alarm goes in before the 49 diagnoses

A fleet survey the same night, across the 22 repos carrying this SIF shim:

| | |
|---|---:|
| repos with ≥1 nightly whose latest run failed | **20 of 22** |
| failing repo/workflow nightly pairs | 52 |
| fully green | 2 |

Fixing causes into a system where red reaches nobody buys another seven days of
silence at the next regression. The detector has to work before its findings are
worth producing.

## What lands

**The sibling of `report-a-red-night` in `nightly-python-matrix-on-github-hosted.yml`,
deliberately.** Same job name, same one-issue-per-fault discipline, same hosted
runner. That workflow is already this repo's answer to *"a nightly nobody reads is
not a test"* — this is the half that was missing. **No third notification dialect
is introduced.**

Two deliberate differences from that sibling:

1. **`always()`, not `failure()`.** On a *green* night the job still runs and makes
   a read-only call against the same API the alarm needs. A revoked token,
   disabled issues or a dropped scope then turns the job red on a quiet night,
   instead of being discovered by the first regression that goes unreported.
   **That is what makes silence mean green: it is checked, not assumed.**

2. **`set -euo pipefail`, with the reporting call last in its branch.** The sibling
   runs `gh issue comment` and then `echo`, under `set -uo pipefail` with **no
   `-e`** — so a comment that fails is masked by the echo's exit 0 and the step
   goes green having reported nothing. That is the original defect wearing a fix.
   Fixed here; **the sibling needs the same treatment in its own change.**

Plus an explicit post-condition: *"the command exited 0"* and *"an open issue now
exists"* are different claims, and only the second is the alarm having worked.

## It must be able to fire, and now it can be made to

`selftest_force_red=true` on `workflow_dispatch` fails the test job **ahead of
checkout**, so the entire chain — real failure, real `needs` propagation, real
`if:` evaluation, real alarm — is rehearsed in seconds rather than by waiting for
a real red night. An alarm that has never fired is indistinguishable from one
that cannot.

It cannot fire by accident: `inputs` is null on push/pull_request/schedule.

## Why the alarm is on a hosted runner

New allowlist entry, and a **different argument** from the two already there
(cost, security): **fate-independence.** Every failure this job exists to announce
is a failure *of the self-hosted pool* — a wedged runner, a saturated queue, a
broken SIF, a shim that dies before a test runs. Put the alarm on that pool and
the outages it most needs to report are exactly the ones that silence it.

**The entry is recorded `approved_by: UNRATIFIED — proposed by an agent, awaiting
operator sign-off`,** not stamped `operator` like its neighbours. No human has
ruled on it. The guard does not machine-check that field, which is precisely why
it must not be used to manufacture consent. Ratify or reject it in review.

## Verification

```
tests/integration/test_ci_nightly_alarm.py            16 passed  (4 are mutation controls)
guard + matrix-shape + tmpdir + concurrency suites   158 passed
guard mutation — allowlist entry removed             SAC-CI001, rc 1
guard baseline — entry present                       rc 0
```

The guard mutation is the one that matters: with my allowlist entry removed the
guard **fails and names the exact job**, so the entry is load-bearing rather than
decorative.

`test_ci_nightly_alarm.py` encodes the rule **mechanically**, not as a name check:
any workflow that runs the suite on a `schedule` must have a job depending on it
with `issues: write`. Adding a third nightly without an alarm turns it RED instead
of quietly recreating this hole. Four mutation controls strip the alarm from an
in-memory copy and assert the rule notices.

## What I could not verify at the time of opening

The end-to-end dispatch (`selftest_force_red=true`, run
[31559141606](https://github.com/scitex-ai/scitex-agent-container/actions/runs/31559141606))
was still in flight. **I will post the result — including the issue it files — as
a comment.** Until that lands, this PR's claim that the alarm fires rests on
static assertions, which is the exact standard this PR argues is insufficient.

## The other 16 repos are a separate lever

Of the 22 shim repos, **16 are callers** of `scitex-ai/.github`'s reusable
`pytest-matrix.yml`; only 6 (including this one) carry local copies. So for the
majority the alarm belongs in that shared reusable workflow, in one edit — but
that is a different repository and is not mine to touch. Flagged for whoever owns
it, with the adoption measurement above as the argument.


ywatanabe1989:
**This PR is `DIRTY`/`CONFLICTING`, so it could not be updated and its CI has not re-run.** The failing checks shown are from 2026-08-12 03:09 and are stale.

```
.github/workflows/pytest-matrix-on-ubuntu-py3-11-3-12-3-13.yml
```

Once that is resolved, the five failures on the stale run should all clear on their own — every one of them belonged to `develop` at the time, not to this branch:

- `test_audit_all_clean` → `[PS-226 §1 job-name-not-hyphenated]`, declared as a deferral on develop by b70feb3 (#1009)
- `test__dev_jobs.py` ×3 → fixed by c10eed0 (#1007)
- `test_host_group.py::test_library_warning_reaches_the_merged_cli_runner_output` → fixed by b9bccd2 (#1004)

The same five were confirmed to clear by update-branch on the sibling PRs in this batch.

Also worth a look: #999 covers closely related ground (the hosted-nightly alarm) and is not conflicted.

(Posted by an automated PR-backlog sweep. Nothing was changed on this branch — resolving the conflict needs its author.)

ywatanabe1989:

PR #999, opened 2026-08-12
fix(nightly-alarm): the alarm could report nothing and still go green

ywatanabe1989:
# The nightly alarm could report nothing and still go green

`report-a-red-night` ran under `set -uo pipefail` — **no `-e`** — with:

```sh
gh issue comment "$num" --repo "$REPO" --body "$body"
echo "commented on existing issue #$num"
```

Without `-e` the step's exit status is the **last** command's. So a comment that
*failed* — revoked token, Issues disabled, rate limit, a `$num` that no longer
exists — was masked by the echo's exit 0, and the step went **green having
reported nothing**.

That is the original disease living inside the thing meant to be the cure: a
nightly alarm that cannot report, reporting success. And the run it guards is by
definition the one nobody reads, so **nothing downstream would ever have
contradicted it**.

The `else` branch was accidentally safe — a bare `gh issue create` happens to be
last — which is exactly the kind of half-working that survives review.

## Two changes

1. **`set -euo pipefail`**, so a failed report fails the step.
2. **A post-condition**: after reporting, assert an *open* issue with that title
   exists. `set -e` proves *"the command exited 0"*; only this proves *"an issue
   now exists"*, and those are different claims.

## Provenance

Found while adding the **missing** sibling alarm to the SIF nightly (#998) —
that workflow had no reporting at all and was silently red for seven consecutive
nights.

Deliberately **its own change** rather than folded into #998: a notify fix riding
along inside a notify PR is how a reviewer misses it.

## Verification

```
guard + concurrency + matrix-shape suites   80 passed
```

Note this fix is **not** covered by an execution test — the same gap it exists to
close. #998 adds `test_ci_nightly_alarm.py`, which asserts `set -euo pipefail`
and the post-condition on the SIF alarm; once both land, extending those
assertions to cover this workflow too is the natural follow-up.


ywatanabe1989:
**Branch updated to `develop`: 5 failures → 1. The survivor is not this PR's fault — a sub-audit failed silently.**

Four of the five were develop's, not this branch's, and are gone: `test__dev_jobs.py` ×3 (fixed by c10eed0, #1007) and `test_host_group.py` (fixed by b9bccd2, #1004). The `PS-226` half of the audit is now covered by the deferral declared in b70feb3 (#1009).

`test_audit_all_clean` still fails, and this one is worth a careful look, because **the audit reports no violations at all**:

```
=== audit-skills ===       SUCC: no skills violations
=== audit-python-apis ===  SUCC: no Python API violations
=== audit-project ===      0 error(s), 98 warning(s), 1 info
=== audit-cli ===          0 error(s), 3 warning(s)
summary: 0 unmasked error(s) (+3 warning/info finding(s)), 0 masked by skip-rules
```

Zero errors — and `audit-all` still exits 1.

The reason is `audit-mcp-tools`. On this run it printed **only** its opening line and then stopped:

```
=== audit-mcp-tools ===
INFO: scitex-agent-container: auditing ... (branch HEAD, HEAD 8729c03f, via explicit)
```

Every other PR run in this batch (#952, #995, #996, #1010) shows that section completing normally:

```
INFO: ... §6 MCP-parity exempt by config ([tool.scitex_dev] mcp_parity_exempt = true) ...
SUCC: ... no CLI convention violations ...
```

So the sub-audit bailed between reading the repo and reaching the exemption check, emitted no finding, and still contributed a nonzero exit.

**It cannot be caused by this branch.** The full diff against the merge-base is one file:

```
.github/workflows/nightly-python-matrix-on-github-hosted.yml
```

No Python, no MCP surface, no tests. A workflow YAML cannot break the MCP-tools audit.

Corroborating: this job ran 354s against a typical ~140s, and the same run warned `§10 import-budget SKIPPED — bare-interpreter baseline 444ms exceeds the 100ms sanity bound`. The node was heavily loaded, consistent with an import/introspection step giving up.

Suggested next step: just re-run the job. If it recurs, the real bug is in `scitex-dev`'s `audit-mcp-tools`, which should not be able to exit nonzero while printing no finding and reporting `0 unmasked error(s)`.

ywatanabe1989:
**Follow-up — confirmed transient. Same commit, two lanes, opposite verdicts.**

```
pytest-matrix-on-ubuntu-py3.13: pass
pytest-matrix-on-ubuntu-py3.11: fail
```

Both ran against the same head. A code defect cannot pass on one interpreter and fail on the other *for a repo-wide audit that does not depend on the interpreter* — and this branch's entire diff is one workflow YAML, with no Python, no MCP surface and no tests in it.

So the earlier `audit-mcp-tools` bail (opening `INFO` line, then nothing, no `SUCC`, no finding, `0 unmasked error(s)`, exit 1) was a transient failure on a loaded runner, not something this PR introduced.

**This PR needs no code change.** Re-running the py3.11 job should turn it green. I tried to kick it, but `gh run rerun --failed` currently refuses with *"This workflow is already running"* — worth retrying once run 31628889495 finishes.

The underlying bug is still real and belongs in `scitex-dev`: `audit-mcp-tools` should not be able to exit nonzero while printing no finding and reporting zero errors. In that state it is indistinguishable from a genuine violation, which is exactly how it cost this PR a red gate.

(Posted by an automated PR-backlog sweep.)

PR #1002, opened 2026-08-12
fix(listen): a zero on the inbox must name its cause — deaf vs not-running

ywatanabe1989:
## The measurement first

I was asked to confirm a finding that 9 of 15 registered agents were **running but deaf**, and that "reach decays with uptime, and it decays silently."

**The 9-of-15 count replicates exactly. The conclusion drawn from it does not.**

Those 9 agents are not deaf. **They are stopped.** Three independent instruments agree:

| instrument | reading |
|---|---|
| `tmux ls` on the host | exactly 6 sessions — the 6 with `inbox_subscribers >= 1`, and no others |
| host process table | `sac mcp channel --name X` running for exactly those same 6 |
| `state.db` `instances` | the 9 carry `ended_at=2026-08-11T17:54:26Z`, `exit_reason=crashed`; only 6 rows have `ended_at IS NULL` |

The registry PID is *not* an instrument here — it reports DEAD for all 15, including agents holding live tmux sessions, because the recorded pid lives in another namespace. `_verdict_tmux.py` says exactly this in its docstring; it is now also measured.

**There is no decay, and there is a positive control for that.** `sac listen` restarted at `2026-08-12T00:27:22Z` — *after* five of the six live agents booted (22:39–22:50Z). All five are still subscribed. Their SSE consumers re-subscribed across a broker restart with no agent restart, which is precisely the invariant `_mcp/_channel_sse.py` claims to hold. The apparent "restarted after 17:59" correlation was `started_at` standing in for *has this row's process survived* — the old rows are stale rows.

This is the third time this fleet has read a wall of zeros as deafness. On 2026-07-14 three agents did it simultaneously and escalated a P0 that did not exist; `_lifecycle/inbox_probe.py` documents that incident at length and warns that a zero is confounded. The warning did not work, because it lives in a docstring and the zero is published on the surface an agent reads *while in a hurry to hand over work*.

## What this PR changes

A zero now arrives with its cause attached, derived by pairing the subscriber count with the one instrument independent of the broker's own bookkeeping — the host's tmux table.

`fault` is published on every `GET /agents` row, on `GET /agents/<name>/status`, and in `sac agents health`:

- **`deaf_inbox`** — a live session observed **and** 0 subscribers. The state that was previously unnameable, and the one worth alarming on: green at every surface, work routed to it, work evaporates.
- **`not_running`** — the registry row outlived the process. Not a delivery fault at all; there is nothing to deliver *to*.
- **`null`** — healthy, or a reading nobody could take.

Discipline inherited from `_reachability.py` and not relaxed:

- **Only a positive observation convicts.** An untakeable snapshot (wedged tmux, or the container-blindness trap where an empty result is a namespace boundary and not an empty fleet) yields no fault. An agent whose spec does not declare the `tui` runtime is never convicted on a missing tmux session it was never going to have — that would be the old pidfile probe's bug run backwards.
- **A live subscriber outranks a missing session.** When the two instruments disagree, the fault is neither: a subscribed agent is by definition not deaf and demonstrably not gone.
- **A fault is a report, never a trigger.** Nothing here is wired to a restart — least of all `deaf_inbox`, whose subject is by definition a healthy session a restart would destroy.

### Two consequences fixed with it

**`a2a_send` no longer tells a sender to wait for a reconnect that has no process to happen in.** A 0-subscriber send raised `no_subscriber_error`, whose remedy says — correctly, for a live agent with a detached adapter — *"NOT LOST … do NOT re-send … it replays on their next connect"*. For a stopped agent that is inverted: no session exists to reconnect, so the row sits in `channel_events` until someone starts the agent and the sender waits forever behind reassuring text. Same shape as the `sac-04` incident one layer over — and on 2026-08-12 it applied to 9 of the 15 registered agents. The send path now raises a distinct `target_not_running` with the opposite remedy. It still does **not** tell the caller to start the target: whether a stopped agent should be running is an operator decision, not a side effect of someone wanting to message it.

**`sac agents health` no longer asserts a process state it never observed.** Its unreachable branch ended with *"The process is up; its inbox adapter is not attached"* — a claim the command had not made and could not make. It was wrong for 9 of 15 agents. It now reports which zero it is, and says so plainly when the cause is unconfirmed.

`sac agents list` needed nothing: it already probes tmux per row and reported all nine as `stopped`. The two surfaces disagreed, and only the peer-facing one was wrong.

## Verification

- `6134 passed, 13 skipped` across `_listen`, `_mcp`, `_lifecycle`, `cli_pkg` (PYTHONPATH pinned to this worktree's `src/`).
- 38 new tests: the classifier rules driven through injected snapshots and a hand-rolled runtime resolver (no mocks, no monkeypatch), the send-path projections, and route-level wiring on the existing real-Starlette harness.
- **End-to-end against the live fleet**, real tmux snapshot, no agent touched: 9 → `not_running`, 6 → `null`, **0 → `deaf_inbox`**. The new instrument's first act is to report that the alarm it was built to raise has no subjects on this host right now.

## Not done here, filed instead

The 9 agents crashed together at `2026-08-11T17:54:26Z` and the `scitex-agent-container-fleet-reconcile.timer` — documented as *"restarts agents whose tmux session is GONE"*, running every 5 minutes — did not resurrect any of them in ~9 hours. Separately, sac's own code notes that spec `restart: {policy: on-failure}` was never enforced. Filed as card `reconcile-timer-did-not-restart-9-agents`; restarting nine agents unasked overnight is the operator's call, not mine.

**Agents touched by this work: none.** Nothing was started, stopped, or restarted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,69 @@ versioning follows [SemVer](https://semver.org/).
 
 ### Added
 
+- **A zero on the inbox now names its cause: `deaf_inbox` vs `not_running`.**
+  `inbox_subscribers: 0` has always been confounded — it means a detached
+  inbox adapter *or* an agent that is not running at all — and
+  `_listen/_reachability.py` has warned about that in its own docstring since
+  it was written. Every consumer since has had to *remember* the warning, on
+  the surface an agent consults right before handing over work, and the record
+  says they do not: 2026-07-14, three agents read a wall of zeros as a deaf
+  fleet and escalated a P0 that did not exist (every agent in their lists was
+  simply stopped); 2026-08-12, a peer read the same zeros as evidence that
+  "reach decays with uptime".
+
+  Measured on the fleet host that morning: **15 registry rows, 9 reporting
+  `unreachable`, and all 9 had no tmux session, no `sac mcp channel` process,
+  and an `instances` row stamped `ended_at=2026-08-11T17:54:26Z`,
+  `exit_reason=crashed`.** Not one of them was deaf. They were dead, and their
+  registry rows had outlived them — still advertising a pid, a port and a
+  `turn_url`, with no field anywhere saying otherwise.
+
+  So the zero is now paired with the one instrument that is independent of the
+  broker's own bookkeeping — the host's tmux table — and the result is NAMED on
+  every `GET /agents` row, on `GET /agents/<name>/status`, and in
+  `sac agents health`:
+
+  - `fault: "deaf_inbox"` — a live session observed AND 0 subscribers. The
+    state that was previously unnameable, and the one worth alarming on: green
+    at every surface, work routed to it, work evaporates.
+  - `fault: "not_running"` — the row outlived the process. Not a delivery
+    fault at all; there is nothing to deliver *to*.
+  - `fault: null` — healthy, or a reading nobody could take.
+
+  Only a POSITIVE observation convicts: a snapshot we could not take (a wedged
+  tmux, or the container-blindness trap where an empty result is a namespace
+  boundary rather than an empty fleet) yields no fault, and an agent whose spec
+  does not declare the `tui` runtime is never convicted on a missing tmux
+  session it was never going to have. The overlay is a REPORT and is wired to
+  nothing destructive — least of all `deaf_inbox`, whose subject is by
+  definition a healthy session a restart would destroy.
+
+### Fixed
+
+- **`a2a_send` no longer tells a sender to wait for a reconnect that has no
+  process to happen in.** A 0-subscriber send raised `no_subscriber_error`,
+  whose remedy says — correctly, for a live agent with a detached adapter —
+  "NOT LOST … do NOT re-send … it replays on their next connect". For a
+  *stopped* agent that advice is inverted: no session exists to reconnect, so
+  the row sits in `channel_events` until someone deliberately starts the agent,
+  and the sender waits forever behind reassuring text. This is the same shape
+  as the `sac-04` incident (a name that was never registered, queued all day
+  behind the same "it's queued" advice) one layer over — and on 2026-08-12 it
+  applied to 9 of the 15 registered agents on the host.
+
+  The send path now reads the `fault` the listen route publishes and raises a
+  distinct `target_not_running` failure with the opposite remedy. It still does
+  **not** tell the caller to start the target: whether a stopped agent should be
+  running is an operator decision, not a side effect of someone wanting to
+  message it.
+
+- **`sac agents health` no longer asserts a process state it never observed.**
+  The unreachable branch ended with "The process is up; its inbox adapter is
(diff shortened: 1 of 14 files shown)


ywatanabe1989:
## Correction to my own evidence: that timestamp is a sweep artifact, not a death time

The PR's argument is unaffected — the nine agents are dead, and the three instruments still agree. But one line of my supporting evidence would mislead whoever reads it next, so I am correcting it in place rather than leaving it.

I wrote that the nine carry `ended_at=2026-08-11T17:54:26Z, exit_reason=crashed`. That is literally true of the rows and it is **not a death time**. In `_state/state_db_gc.py`, `now_ts = now_iso()` is computed **once, before the loop**, and then written to every row the sweep reaps:

```python
now_ts = now_iso()          # computed ONCE
...
for row in rows:
    try:
        os.kill(pid, 0)
    except (OSError, ProcessLookupError):
        conn.execute(
            "UPDATE instances SET ended_at=?, exit_reason='crashed' WHERE id=?",
            (now_ts, row["id"]),   # same now_ts for every row
        )
```

So `exit_reason='crashed'` means only *"`os.kill(pid,0)` raised ESRCH at sweep time"*. It carries no cause and no time of death. **An identical second across N rows is the expected output of one sweep call — not evidence of a simultaneous kill.** I read it as the latter.

Two corrections follow:

1. **It is eleven agents, not nine.** `select ... where ended_at='2026-08-11T17:54:26Z'` returns 11 rows: the nine plus `dotfiles` and `scitex-cards`, which read as "survivors" only because they were manually revived at 22:49–22:50Z. Every sac agent alive on this host died in one event.

2. **They died ~10h46m earlier, at ~2026-08-11T07:08:00Z**, when tmux server pid `2621557` went away and took every pane with it. Eleven systemd `tmux-spawn-<uuid>.scope` units logged `Consumed ... CPU time` (scope release = last process exited) inside a 2-second window, each scope's pane pid matching an `instances.pid`, all children of that one server. It is gone; today's server is pid `2173676`.

Ruled out by measurement: OOM (zero lines in `journalctl -k`/`dmesg -T`), reboot (`uptime -s` = 2026-08-08), user-session teardown (`Linger=yes`, and `systemd --user` survived to log the teardown), graceful stop (would write `exit_reason='stopped'`). **Not determined:** what killed the tmux server — it is not a systemd unit, so its death is not journaled.

### Why this belongs on this PR

It is the same failure this PR exists to fix, one layer down. `inbox_subscribers: 0` was a number whose *cause* was unreadable, so callers supplied one. `ended_at` on a `crashed` row is a number whose *meaning* is unreadable, so I supplied one. In both cases the field is honest about what it stores and silent about what it means, and the reader fills the gap with the most available story.

Worth a follow-up (not this PR): `exit_reason='crashed'` would carry its own caveat if it were named for what it observes — something like `pid_absent_at_sweep` — and the sweep could stamp `ended_at` as the observation time it is, distinct from a death time it never had.

PR #1005, opened 2026-08-12
feat(ci-rail): a push writes a card, CI's verdict writes back, the agent hears it

ywatanabe1989:
## The rail the operator asked for

> プッシュに連動して、必ずフックでカードにその情報を書かないといけない…GitHubから信号が帰ってきたときに、フックでカードに書き込む、それがエージェントに通知が行く

Implements **ADR-0024**, piloted on this repo and no other.

```
push  -> .githooks/pre-push registers the commit on a card
CI    -> a job on the control-plane host writes the verdict to the SAME card
card  -> the verdict is delivered into the pushing agent's live session
```

The card id is derived from `(repo, head sha)` on both sides, which is why no
channel is needed between a git hook and a runner job that starts minutes later.

## Why a runner step and not a webhook

`sac listen` binds `127.0.0.1` only, by policy — there is no public route for
GitHub to POST to. But a self-hosted runner already long-polls **outward** to
GitHub and then executes on our hardware, where loopback is reachable and the
bearer is a local file. The GitHub→host path exists and is already trusted.
That is nearer 「フック」 than the two existing pollers: it fires *on* the event
rather than noticing it up to 300 s later, and it drops the `gh`-auth
dependency that made both fail closed. Neither has ever delivered a verdict on
this host — `verdict_delivered` has never even been created.

## The one thing ADR-0024 got wrong, measured

The ADR reasoned "the self-hosted runners execute on this host". They do not.
`vars.CI_RUNS_ON` is `scitex-org-cpu` — **four runners on four machines**, and
only `scitex-04-org-cpu-01` sits on the box running `sac listen` and the card
store. The premise holds for 25% of jobs; on the other 75% both
`127.0.0.1:7878` and `127.0.0.1:55432` are a *different* machine's daemon and a
*different* postgres — recorded nowhere, delivered to nobody, nothing raised.

So the verdict job is pinned to a `sac-control-plane` label carried by that one
runner, and `test_verdict_job_is_pinned_to_the_control_plane_host` recomputes
the pin from the workflow rather than trusting it. The cost is one queue
admission (median ~706 s here); the ADR's cheaper in-job placement was only
available under the false premise.

## Three silent failures this refuses to reproduce

| Failure | What this does |
|---|---|
| **Delivering to a deaf agent.** 9 of 15 registered agents hold zero inbox subscribers; a message lands on an empty bus and raises nothing. | Recipient resolution *prefers a reachable agent*. sac's own `resolve_owner` takes the first spec in sorted order — for this repo, deterministically the agent with no subscriber since 2026-08-10. A zero delivered-count is reported as a **recipient** fault, not a rail fault, since the event is persisted and replays. |
| **Writing to a store nobody reads.** `:5442` and `:55432` both answer, both report `store_uuid 1d55dd6e-…`, and hold **3496 vs 3793** cards — two divergent databases wearing one identity. An unset DSN silently picks a local file. | DSN passed explicitly, backend re-checked, resolved target printed so a run is auditable after the fact. |
| **Saying only "Red".** Seven consecutive red nights reached nobody; arriving without naming the failing test solves half of that. | A red verdict quotes the first failure lines from the **failed jobs'** logs — the per-job endpoint, because this job is itself running and the whole-run archive is still empty and greps exactly like a clean one. Every unreadable-log path says so explicitly. It reports the verdict and quotes the log; it does not assert a cause it cannot know. |

A red gate files the card `failed`, never `blocked`. A `blocked` card must name
its gate, none of this store's blockers describes a red suite, and a gate-less
blocked card nudges nobody and leaves the runnable count — which is how 21
operator decisions sat invisible for weeks.

## Also fixed

- `.githooks/pre-push` resolves ruff via `uv` when no venv ships it. The agent
  containers ship **no** ruff at any searched path, so this gate could not run
  in the environment most pushes come from — it either blocked every push or
  taught everyone to reach for `--no-verify`, disarming the protected-branch
  guard in the same stroke. Still strict; same ruff, same rules.
- `pyproject.toml` `library_dirs` held `".github/ci"`, but `is_script()` tests
  `lib_dir in path.parts`, so a value containing a separator can never match.
  It looked configured and exempted nothing.

## Verification

The push half fired on this very branch: a real `pre-push` hook recorded commit
`b2745405` on card `ci-scitex-agent-container-b27454051ade`. The delivery half
was proven end to end before this PR — a verdict reached a running agent's live
session, confirmed by that agent independently.

**This PR's own gate run is the acceptance test for the CI half.**

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -71,13 +71,40 @@ if [ -z "$RUFF" ] && command -v ruff >/dev/null 2>&1; then
   RUFF="$(command -v ruff)"
 fi
 
+# LAST RESORT: let uv fetch ruff on demand. Measured 2026-08-12 — the agent
+# containers ship NO ruff at any of the paths above and none on PATH, so this
+# gate could not run in the very environment most pushes come from. A gate
+# that cannot run has exactly two outcomes, and both are bad: it blocks every
+# push, or it teaches everyone to reach for --no-verify, which disarms the
+# protected-branch guard above at the same time. Resolving ruff through uv
+# keeps the check STRICT (same ruff, same rules) instead of trading strictness
+# for availability. Declared as an array because it is a multi-word command.
+RUFF_ARGV=()
+if [ -n "$RUFF" ]; then
+  RUFF_ARGV=("$RUFF")
+else
+  for uv_bin in "${SAC_PRE_PUSH_UV:-}" "/uvwork/bin/uv" "$HOME/.local/bin/uv"; do
+    if [ -n "$uv_bin" ] && [ -x "$uv_bin" ]; then
+      RUFF_ARGV=("$uv_bin" tool run ruff)
+      RUFF="$uv_bin tool run ruff"
+      break
+    fi
+  done
+  if [ -z "$RUFF" ] && command -v uv >/dev/null 2>&1; then
+    RUFF_ARGV=("$(command -v uv)" tool run ruff)
+    RUFF="uv tool run ruff"
+  fi
+fi
+
 if [ -z "$RUFF" ]; then
   # Loud failure — never silently skip the gate (handoff Q3). If a
   # genuine emergency requires bypass, --no-verify is documented as
   # the explicit override and shows up in the audit log.
   echo "PRE-PUSH BLOCKED: ruff binary not found." >&2
   echo "  Searched: \$SAC_PRE_PUSH_RUFF, .venv/bin/ruff, /opt/venv-agent/bin/ruff," >&2
-  echo "           /opt/venv-sac/bin/ruff, \$HOME/.local/bin/ruff, PATH." >&2
+  echo "           /opt/venv-sac/bin/ruff, \$HOME/.local/bin/ruff, PATH," >&2
+  echo "           and 'uv tool run ruff' (\$SAC_PRE_PUSH_UV, /uvwork/bin/uv," >&2
+  echo "           \$HOME/.local/bin/uv, PATH)." >&2
   echo "  Install ruff via: pip install -e .[dev]" >&2
   echo "  Or point the hook at an existing ruff: export SAC_PRE_PUSH_RUFF=/path/to/ruff" >&2
   echo "  Or bypass once (emergency only): git push --no-verify" >&2
@@ -88,7 +115,7 @@ fi
 # the unused-import slip-through that motivated this hook.
 # --extend-per-file-ignores excludes __init__.py re-export shims (the
 # standard idiom — imports there *are* the public surface, not dead code).
-if ! "$RUFF" check \
+if ! "${RUFF_ARGV[@]}" check \
       --select F401,F811 \
       --extend-per-file-ignores "**/__init__.py:F401" \
       src/ tests/; then
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -30,7 +30,16 @@ cd "$(git rev-parse --show-toplevel)"
 #
 # git feeds pre-push one line per ref on stdin:
 #   <local ref> <local sha1> <remote ref> <remote sha1>
+#
+# READ IT ONCE, INTO A VARIABLE. stdin is a stream: whoever consumes it
+# first consumes it for everyone. The branch guard below and the card
+# record at the bottom of this file both need these refs, and before
+# ADR-0024 the guard's `while read` swallowed them. Capturing up front is
+# what lets a second consumer exist at all.
+PUSHED_REFS="$(cat)"
+
 while read -r _local_ref _local_sha remote_ref _remote_sha; do
+  [ -n "${remote_ref:-}" ] || continue
   case "${remote_ref##refs/heads/}" in
     main|master)
       echo "PRE-PUSH BLOCKED: refusing to push directly to '${remote_ref##refs/heads/}'." >&2
@@ -39,7 +48,7 @@ while read -r _local_ref _local_sha remote_ref _remote_sha; do
       exit 1
       ;;
   esac
-done
+done <<< "$PUSHED_REFS"
 
 # Locate ruff. Search order matches the venvs the SciTeX harness
 # typically installs into; falls back to whatever ruff is on PATH.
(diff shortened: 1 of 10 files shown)


ywatanabe1989:
## The pinning tradeoff, stated explicitly

Review raised that pinning the verdict job to `sac-control-plane` makes the
rail exactly as available as one machine, and that a pinned job which cannot
be scheduled **queues into silence** — indistinguishable from a saturated pool
at this repo's median queue of 289 s (p90 902 s). That is the correct concern
and it is now stated in the workflow itself rather than left implicit.

### On the "runner is OFFLINE" report — corrected, with evidence

A review measurement read `scitex-04-org-cpu-01` as `OFFLINE`, concluding the
pin would lose every verdict. Re-measured two independent ways:

| Source | Reading |
|---|---|
| `GET /orgs/scitex-ai/actions/runners` (re-read) | `status=online busy=true` |
| Process table on the host | `Runner.Listener` up **4 h 48 m**; `Runner.Worker spawnclient` started 2 s before the check |

A dead runner cannot accept a job. The `OFFLINE` reading was transient — a
reconnect blip, not a dead runner. Recording it because the opposite error
(trusting one green reading) is the same mistake in the other direction, and
this fleet has now made both in one night.

### The trade, as taken

- **Latency** — a separate job pays one queue admission. ADR-0024's cheaper
  in-job placement was only available under its false premise that gate legs
  run on the control-plane host; they do not, 3 times in 4.
- **Availability** — the rail is as available as `scitex-04-org-cpu-01`.

**Why not "stop requiring loopback".** It is the right long-term answer and it
is not available today: `sac listen` binds `127.0.0.1` by policy, *and* the
card store DSN is loopback-per-host too (`127.0.0.1:55432` on scitex-01 is
scitex-01's own postgres, not this one). So a non-04 runner can reach neither
the bus nor the board without new transport — a real change, not a flag.

**How a lost verdict surfaces.** This is what the `blocked`/`compute` status
is for. A pending card names its gate and stays visible on the board; a
verdict that never arrives leaves a card parked on `compute`, which is exactly
the observable a stale-pending sweep nudges on. That sweep does not exist yet
and is the named next step — filed rather than assumed, because "the rail
reports its own silence" is a claim this PR does not get to make yet.

ywatanabe1989:
## Correction: the control-plane runner FLAPS. That is worse than offline.

My previous comment reported `scitex-04-org-cpu-01` as online and treated an
earlier `OFFLINE` reading as a transient misread. Both readings were true when
taken:

    t0        scitex-04-org-cpu-01   OFFLINE
    t0+~10m   scitex-04-org-cpu-01   online   (API and process table agree)

It is not dead and it is not healthy. **It flaps** — and it remains 1 of 7 org
runners carrying `sac-control-plane`, still the only one.

**An intermittently-available runner is a worse foundation for a pinned job
than a permanently dead one.** A dead runner gets noticed and fixed. A flapping
one makes the rail lose verdicts *sometimes*, and at a 289 s median / 902 s p90
a job that cannot be scheduled is indistinguishable from one that is merely
waiting. Intermittent silent loss is the hardest failure to diagnose, and it
would sit inside the rail built to make CI failures impossible to miss.

So the availability statement is stronger than my last comment implied:

> **The rail is exactly as available as one machine, and that machine has been
> observed transitioning offline→online inside ten minutes.**

### What this PR does and does not claim

It does **not** claim the rail reports its own silence. It cannot yet. A pinned
job that is never scheduled emits nothing, and the detector therefore cannot
live inside that job.

What exists today is the `blocked`/`compute` pending card: a verdict that never
arrives leaves a card parked on a *named* gate, visible on the board, which is
the observable a stale-pending sweep would nudge on. **That sweep is not built.**

### The options, with the flap weighing on them

1. **Wait for the runner to stabilise** — not a foundation; nobody was watching it.
2. **Label a second runner** — needs a second host running `sac listen`.
3. **Stop requiring loopback** — the only option that makes the rail as
   available as the pool. `sac listen` binds loopback *by policy*, and a policy
   is revisable with a stated reason, unlike a constraint. Note the card store
   is loopback-per-host too (`127.0.0.1:55432` on scitex-01 is scitex-01's own
   postgres), so this means new transport for **both** the bus and the board —
   a real design change, not a flag.

I did not take either of the two runner readings on trust; both were
re-measured against the host's process table (`Runner.Listener` uptime, live
`Runner.Worker` spawns), because a dead runner cannot accept a job.

ywatanabe1989:
## The red on this PR is MINE, and the audit caught the exact hazard under discussion

Measured from the failed py3.13 job's log (`31570498580`), not inferred:

    [E] [PS-224 §1 runner-destination-unregistered]
        pytest-matrix-...yml::verdict: job `verdict` targets
        `[self-hosted, Linux, X64, sac-control-plane]`

**One correction to a reasonable inference made in review:** `quality` passing
on this head (37s) does *not* clear `test_audit_all_clean`. They are different
things — the `quality` workflow and the in-suite audit test run different
surfaces, and the in-suite one runs `scitex-dev ecosystem audit-all` inside the
SIF. The in-suite test is precisely what failed.

**PS-224 is the ecosystem's codification of "a label is not a pool."** It
rejects a workflow job pointed at a runner destination the registry does not
know about — which is exactly the hazard raised in review about pinning to
`sac-control-plane`. Three independent objections have now converged on the
same line of YAML: the availability argument, the flap observation, and a
linter rule that predates both. That is worth taking seriously rather than
exempting away.

The registry is `scitex_dev.hosts` (`list_runner_destinations`, with
`packaged_default_runner_destinations` as PS-224's floor) — so registering
`sac-control-plane` is a **scitex-dev** change, not a change here. Filed rather
than worked around: an exemption in `.scitex/dev/config.yaml` would silence a
rule that is telling the truth, and the one existing PS-224 exemption
(`cla.yml`) is a security argument, not a convenience one.

### Why this is the best possible demonstration of the rail

This red is not inherited. `git merge-base --is-ancestor 15b0515 e8e936b0`
returns 1 — the branch predates the `scitex-dev<0.48` ceiling — so the
0.48.0 breakage was the obvious hypothesis, and reading the log refuted it.
**The failure is caused by the verdict job this PR adds.**

So the rail is about to deliver, to its own author, a red verdict naming the
defect that the rail's own new job introduced. That is a stronger test than any
synthetic one: the message has to be useful to someone who does not yet know
what broke.

PR #992, opened 2026-08-12
ci(sif-shim): unpin the release path from Spartan's filesystem

ywatanabe1989:
# 0 tests executed → 12,666 passing

That is what this PR does to `main`, measured on the same runner.

> **The two red legs are NOT a verdict on this change.** They are an unrelated,
> still-undiagnosed runner fault (`fatal: failed to write commit object`). All
> 24 reported failures were run against `main`'s own source locally — **212
> passed, 0 failed, 24 of 24 accounted for.** `main`'s tree contains no broken
> test. See the comments below for the full evidence.
>
> **The honest state of sac's release path: the shim fix is correct and proven,
> and it cannot land until an unrelated environmental fault is diagnosed.**
> That fault has now survived three refuted explanations — `safe.directory`,
> `.gitconfig.lock`, and a `/tmp` collision theory withdrawn by its own author.

## This is the PR that fixes the release path

Tags are cut from `main`, and `main` kept the Spartan-only shim. A green
`develop` proves nothing about a tag. #991 landed the same change on `develop`;
**that one does not fix the release path.** This one does.

### What `main` has been doing, unnoticed, for a week

`main` @ `5532e6b` — the exact commit this branch is cut from — scheduled
nightly, **all three legs, every night**:

```
mkdir: cannot create directory ‘/data’: Permission denied
##[error]Process completed with exit code 1.
```

| date | conclusion |
|---|---|
| 2026-08-11 17:56 | failure |
| 2026-08-10 17:51 | failure |
| 2026-08-09 17:31 | failure |
| 2026-08-08 17:30 | failure |
| 2026-08-07 17:49 | failure |
| 2026-08-06 18:24 | failure |
| 2026-08-05 18:22 | failure |

**Seven consecutive nights, zero tests executed, and nobody was told** — the
nightly workflow has three steps (run, upload coverage, clean scratch) and
**no failure notification of any kind**.

## The mechanism

The apptainer *resolution* is not the killer and was already defused —
`SCITEX_CI_APPTAINER` is set, `/usr/bin/apptainer` exists on the compute nodes,
and the `~/.env-3.11` PATH prepend is useless but harmless. The killers are
three lines further down:

```sh
export APPTAINER_TMPDIR="/data/gpfs/projects/punim0264/…/apptainer-tmp"
mkdir -p "$APPTAINER_TMPDIR"          # aborts under set -euo pipefail
exec … --bind /data/gpfs/projects/punim0264 …   # apptainer refuses a bind
                                                # whose source is absent
```

[...]

PR #1006, opened 2026-08-12
ci(routing): give the sub-minute single-core jobs their own lane

ywatanabe1989:
## Two thirds of CI was queueing behind the other third for machines it could not use

Measured over 276 self-hosted jobs in 2.1 h, this repo's CI is bimodal:

| | share of jobs | median run | cores it uses |
|---|---|---|---|
| `tests` | 31 % | 368 s | all 32 (`xdist -n $(nproc)`) |
| light lane | 69 % | 12-51 s | about one |

They share one pool: four 32-core machines, one runner each, measured at
**93-94 % utilisation** with a **289 s median queue and 902 s p90**. So a
12-second `ruff` run took a whole 32-core machine for its turn and waited
roughly twenty times its own runtime to get one.

## The measurement that decided this

The three `spartan-cpu` runners were idle, and the plan on the table was to
treat them as three more machines. A canary (#1001) checked the one thing
nobody had: **`nproc = 1`**. The full suite takes **497 s** there against
**170 s** on a 32-core host.

That single fact splits the decision cleanly. One-core slots are the wrong
place for `tests` and exactly the right size for the light lane. "Three idle
machines" was really three one-core slots, which is a different plan than the
one we were about to make.

## What moved, and the property that qualifies it

| workflow | check | median | enters SIF? |
|---|---|---|---|
| `lint.yml` | `lint` | 12 s | no |
| `no-hosted-runners-guard-on-self-hosted.yml` | `no-hosted-runners` | 13 s | no |
| `quality-audit-on-ubuntu-latest.yml` | `quality` | 44 s | no |
| `import-smoke-on-ubuntu-py3-12.yml` | `import-smoke` | 47 s | no |
| `rtd-sphinx-build-on-ubuntu-latest.yml` | `docs` | 51 s | no |

The qualifying property is **not runtime — it is the SIF**. None of these calls
`exec-in-sif.sh`; they are checkout + uv on the bare runner, so none needs the
`/data/gpfs` bind. `import-smoke` installs on the bare runner *deliberately*, to
prove the package imports without the SIF's baked environment. And the guard's
own `setup-uv` step already carried the note that the bare Spartan node has no
Python and is PEP-668 externally-managed — these jobs were written for that
host's constraints.

## What did NOT move, and why the line is exactly there

`tests`, all four jobs of `pypi-publish-and-github-release-on-tag.yml`,
`autobump-release-sweep`, and `newb`. **Every one of them enters the SIF**, and
`origin/main`'s `exec-in-sif.sh` still binds `/data/gpfs/projects/punim0264`
unconditionally (line 45 `APPTAINER_TMPDIR`, line 95 `--bind`; `develop` guards
both since #880). **#992 is the precondition for moving any of them**, and
`main` was just promoted — a release path whose hardware varies silently is the
worst thing to ship tonight.

Please do not read this PR as permission to move the rest. The table above is
the whole authorised set; the SIF column is the reason.

## Why per-workflow `runs-on` and not the variable

`vars.CI_RUNS_ON` is load-bearing and it is working — measured across its
02:27:58Z change, at flat arrival (1.89 -> 2.01 jobs/min), flat service
(42 -> 41 s) and flat utilisation (93 -> 94 %), median queue fell **1197 s ->
289 s**. It is the mitigation, not the cause. It is also read by **10 workflows
/ 13 jobs including all four release jobs**, so editing it re-points release
hardware with no diff to read.

A literal list in each file keeps this routing visible in review, revertible one
workflow at a time, and structurally unable to reach the release gate.

## Verification

- The `no-hosted-runners` guard parses every `runs-on` and fails anything it
  cannot prove self-hosted — **including its own**, which this PR moved. Ran it
  against this tree before committing: **exit 0**.
- **This PR verifies itself.** Four of the five moved workflows are PR-gate
  checks, so `lint`, `quality`, `import-smoke`, `docs` and `no-hosted-runners`
  on this very PR are running on `spartan-cpu`. If a one-core slot cannot serve
  them, these checks say so directly.
- `tests` on this PR still runs on the unchanged pool, which is the control.

## Incidental fix

`rtd-sphinx` commits regenerated HTML on default-branch pushes, and commit
signing fails 13/13 on compute-04 (`fatal: failed to write commit object` — ssh
signing against a key that is not there). The canary probed `git commit` on
`spartan-bm153`: it succeeds, signing unset. So this also takes that refresh off
the one host where it reliably breaks.

## Not done

No test disabled, skipped or deselected — capacity, not coverage. No label
widened, no runner re-labelled, `CI_RUNS_ON` untouched. Per-ref `concurrency`
groups were measured at 4 of 43 active jobs (9 %) and left out on purpose:
real, but not the ceiling, and I did not want it read as one.

### DIFF
--- a/.github/workflows/import-smoke-on-ubuntu-py3-12.yml
+++ b/.github/workflows/import-smoke-on-ubuntu-py3-12.yml
@@ -54,7 +54,23 @@ jobs:
   install-check:
     # Kept byte-identical — see lint.yml. (Verified: not protection-pinned.)
     name: import-smoke-on-ubuntu-py3-12
-    runs-on: ${{ fromJSON(vars.CI_RUNS_ON || '["self-hosted","Linux","X64","scitex-ci"]') }}
+    # LIGHT LANE — pinned to spartan-cpu on purpose. Measurements: PR #1001.
+    #
+    # 47s median, single-core, and it never enters the CI SIF. That is the
+    # POINT of this job: it installs from source on the BARE runner precisely to
+    # check that the package imports without the SIF's baked environment, so
+    # routing it to a SIF-capable host would not help it and pinning it here
+    # costs it nothing.
+    #
+    # It was holding one of four 32-core machines per run on a pool at 93-94%
+    # utilisation, then queueing 289s median (p90 902s). The spartan-cpu runners
+    # are nproc=1 — wrong for `tests` (497s there vs 170s on 32 cores), right
+    # for this.
+    #
+    # PINNED PER WORKFLOW rather than through vars.CI_RUNS_ON so the routing is
+    # visible in a diff and cannot reach `tests` or the release gate, which DO
+    # enter the SIF and still need #992 first.
+    runs-on: ["self-hosted", "Linux", "X64", "spartan-cpu"]
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -58,7 +58,29 @@ jobs:
     # stops reporting has already held a release hostage once. Correcting the
     # "on-ubuntu" names repo-wide is a separate, coordinated change.
     name: ruff-on-ubuntu-latest
-    runs-on: ${{ fromJSON(vars.CI_RUNS_ON || '["self-hosted","Linux","X64","scitex-ci"]') }}
+    # LIGHT LANE — pinned to spartan-cpu on purpose. Measurements: PR #1001.
+    #
+    # This repo's CI is bimodal. Over 276 self-hosted jobs (2.1 h): `tests` is
+    # 31% of jobs at 368s and uses all 32 cores (xdist runs -n $(nproc)); the
+    # other 69% — this job included — run 12-51s on about ONE core. `ruff` is
+    # the smallest of them at 12s median, and it never enters the CI SIF: no
+    # exec-in-sif.sh, no /data/gpfs bind, just checkout + uv + ruff.
+    #
+    # So a 12-second single-core job was taking one of four 32-core machines
+    # for its turn, on a pool measured at 93-94% utilisation — and then waiting
+    # 289s median (p90 902s) for the privilege.
+    #
+    # The spartan-cpu runners were sitting idle, and a canary established the
+    # fact nobody had checked: they are nproc=1. That makes them wrong for
+    # `tests` (the suite takes 497s there vs 170s on 32 cores) and correctly
+    # sized for exactly this kind of job.
+    #
+    # PINNED PER WORKFLOW rather than through vars.CI_RUNS_ON so the routing is
+    # visible in a diff, revertible on its own, and CANNOT reach `tests` or
+    # pypi-publish-and-github-release-on-tag.yml by accident. Do not generalise
+    # it to those: both DO enter the SIF, and `main`'s exec-in-sif.sh still
+    # binds /data/gpfs/projects/punim0264 unconditionally (#992 is the fix).
+    runs-on: ["self-hosted", "Linux", "X64", "spartan-cpu"]
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


ywatanabe1989:
## Verified on this PR — and it came with its own control

Every check here was created at **06:29:52Z**. What happened next split exactly
along the line this PR draws:

| check | runner | started | queue | run | result |
|---|---|---|---|---|---|
| `no-hosted-runners` | spartan-cpu-org-02 | 06:29:55 | **3 s** | 15 s | pass |
| `lint` | spartan-cpu-org-01 | 06:29:55 | **3 s** | 14 s | pass |
| `quality` | spartan-pooled-cpu-01 | 06:29:55 | **3 s** | 26 s | pass |
| `docs` | spartan-cpu-org-01 | 06:30:12 | **20 s** | 28 s | pass |
| `import-smoke` | spartan-cpu-org-02 | 06:30:13 | **21 s** | 20 s | pass |
| `tests` py3.11 | *unassigned* | — | still queued | — | — |
| `tests` py3.13 | *unassigned* | — | still queued | — | — |

**Queue time 3-21 s against a 289 s median (p90 902 s)** — one to two orders of
magnitude, for 69 % of the job count. The work spread across all three Spartan
runners, so the lane is genuinely three slots wide, not one.

`tests` is the control: same PR, same second, unchanged routing, **still
unassigned** while the light lane finished. That is the contention this PR
removes, visible in a single run.

### The unexpected part: they got FASTER, not slower

Run times on a one-core Spartan slot versus the medians on the 32-core pool:

    import-smoke   47s -> 20s
    quality        44s -> 26s
    docs           51s -> 28s
    lint           12s -> 14s
    guard          13s -> 15s

The two sub-15-second jobs are a wash, as expected. But the three larger ones
roughly **halved**. The 32-core machines were never giving these jobs 32 cores —
they were co-tenanting with a `tests` leg running `xdist -n $(nproc)`, which
takes every core on the box. A dedicated single core beats a contended
thirty-second of one.

That was not the argument for this change, and I would not have predicted it.
It does mean the split helps both lanes: the light jobs stop waiting, and
`tests` stops sharing.

ywatanabe1989:
## Did the ceiling move, or merely redistribute? It moved.

Taking the **same measured job stream** (41.4 min, 83 self-hosted jobs, the
post-02:27:58Z regime) and re-assigning each job to the lane it lands in after
this PR, then recomputing against each lane's real slot count:

```
BEFORE  everything on 4 org slots      busy=9287s  cap=9924s   UTIL=94%  SATURATED

AFTER   LIGHT -> 3 spartan slots  n=63  busy=2055s  cap=7443s   UTIL=28%  headroom
        HEAVY -> 4 org slots      n=20  busy=7232s  cap=9924s   UTIL=73%  headroom
```

Per workflow in the light lane: `docs` 653s, `import-smoke` 526s, `quality`
521s, `lint` 226s, `no-hosted-runners` 129s. The heavy lane is `tests` alone at
7232s.

### Why this works, stated plainly

**76 % of the jobs were only 22 % of the load.** 63 of 83 jobs accounted for
2055 of 9287 busy-seconds; `tests` was 24 % of jobs and 78 % of the work.

Queueing is driven by *job count* competing for *slots*, not by total CPU. So
splitting on job count is the lever, and it is why 63 numerous-but-cheap jobs
moving to 3 dedicated slots leaves that lane at 28 % — effectively no queue,
which is what the 3-21 s measurements on this PR already showed empirically.

The heavy lane drops **94 % -> 73 %**. On the rho/(1-rho) curve that is the
important part: expected wait goes from 15.7 to 2.7 in relative units, roughly a
**6x reduction**, because 94 % was on the cliff and 73 % is not.

### What this means for #992

**#992 becomes desirable rather than urgent.** The argument for hurrying it was
that the heavy lane needed Spartan's machines to breathe. At 73 % it does not.
#992 remains the correct precondition for ever moving `tests` or the release
gate, and `main`'s unconditional `/data/gpfs` bind is still a latent
release-path bug worth fixing on its own merits — but it is no longer gating
queue relief.

### Caveats, because this is a projection and not a measurement

- Not merged, so this re-lanes a historical stream rather than observing a new
  one. Worth re-measuring after merge.
- n=20 for `tests` in a 41-minute window is a small sample; the 73 % figure is
  the softest number here.
- It assumes light-job service times are unchanged. Measured on this PR they got
  *faster* (import-smoke 47->20 s, quality 44->26 s, docs 51->28 s), so 28 % is
  conservative — the real light lane should sit lower.
- It assumes arrival is unchanged. Sixteen agents opening PRs faster scales both
  lanes; the split changes the slope, not the fact that demand can grow.

PR #1009, opened 2026-08-12
chore(audit): declare PS-226 as a deferral, with the reason it is not a rename

ywatanabe1989:
Part **(3)** of the three the `scitex-dev<0.48` ceiling covers. One file: `.scitex/dev/config.yaml`.

## The finding is correct. The remedy is what is deferred.

scitex-dev 0.48.0 added **PS-226** (`job-name-not-hyphenated`). It fires on `JobSpec(name="sac.accounts-refresh")`, and it is right to: `_jobs/_systemd.py::systemd_unit_name` derives the unit **filename** from the job name **verbatim**, with no sanitisation.

So this is **not** a false positive, and the entry does not claim it is.

What is deferred is the fix, because the obvious one is more dangerous than the defect it repairs:

- renaming does **not** re-label the running unit — it makes sac install a **second** unit beside the live hand-written one;
- that job is the fleet's **sole OAuth refresher**, against a **single-use** token;
- two refreshers racing **revoke each other**, taking every agent's authentication down with them.

This is the same hazard that already made sac refuse to federate `sac listen` as `sac.listen`: the live unit is the hyphenated `sac-listen.service`, and a dotted duplicate would have put two `Restart=always` supervisors on one port.

The migration is therefore ordered **stop-old → remove-old → install-new → verify-exactly-one**, supervised, and explicitly not run at an hour when nobody is watching. The supervised path already exists at `_jobs/_migrate/_renames.py` — it should be used rather than reimplemented. Tracked in card `sac-accounts-refresh-unit-migration`.

## Why `skip-rules` and not `audit.skip`

`audit.skip-rules` is the sanctioned mechanism precisely because it is **not** a silencer:

- it **rejects an entry with no written rationale** — loudly, naming the entry, including whitespace-only; and
- `audit-all` reprints a **masked inventory** on every run — count, rule ids, per-rule masked count, and each written reason.

So the deferral stays visible on every audit rather than disappearing. The legacy `audit.skip` list is bare rule codes applied silently, and is deliberately not used here.

## Verified against the real loader, not assumed

Run against `scitex_dev._cli.audit._config._skip_rules`:

```
parsed entries: 1
  rule='PS-226'   reason (675 chars)
PS-226 present with a non-empty rationale: OK

audit keys: ['exemptions', 'root-whitelist', 'skip-rules']
PS-224 exemption still present: True
root-whitelist dirs: ['config', 'containers']

whitespace-only reason rejected, as designed:
  invalid `audit.skip-rules` entry: 'PS-999' has an empty `reason`
```

The last line matters: the guard that makes this mechanism trustworthy is still live, so this entry earned its pass by carrying a reason rather than by disabling the check.

## The closing condition is written down

The change that performs the migration **removes this entry in the same commit**. An entry left behind after its cause is gone masks a rule with nothing left to mask — which is how an audit quietly stops meaning anything. That is stated in the config comment and on the card.

## Sequencing

With this, part (1) is in **#1007** and part (3) is here. Part (2) — `test_host_group` — belongs to **#1004**. The `<0.48` ceiling comes off in whichever of the three lands **last**, so CI actually installs 0.48.0 and proves all three adaptations at once. Lifting it earlier would merge fixes whose own CI cannot exercise them.

### DIFF
--- a/.scitex/dev/config.yaml
+++ b/.scitex/dev/config.yaml
@@ -58,6 +58,24 @@ audit:
   #
   # The supervised path already exists: `_jobs/_migrate/_renames.py`.
   # This entry comes off in the same change that runs the migration.
+  #
+  # *** READ THIS BEFORE DELETING ANY UNIT. ***
+  # "verify-exactly-one" means EXACTLY ONE UNIT FOR THIS ONE JOB — old
+  # gone, new present. It does NOT mean "one accounts job on the host".
+  # There are TWO accounts jobs and BOTH MUST EXIST:
+  #
+  #   sac.accounts-refresh                        rotates the credential,
+  #                                               every 2h, on the host
+  #                                               holding refresh material
+  #   scitex-agent-container-accounts-keepalive   distributes the
+  #                                               ACCESS-ONLY copy,
+  #                                               every 15min
+  #
+  # They are DIFFERENT JOBS doing different work, not a rename pair and
+  # not duplicates of each other. Anyone who reads this rule as "collapse
+  # the accounts jobs to one" would remove a job the fleet depends on —
+  # which is a larger outage than the naming defect being deferred here.
+  # The only thing PS-226 asks for is the SHAPE OF ONE JobSpec NAME.
   skip-rules:
     - rule: PS-226
       reason: >-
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
## To whoever merges this LAST: remove the ceiling in this same PR

Three PRs each fix one of the three incompatibilities the `scitex-dev>=0.31.1,<0.48` cap covers:

| part | PR |
|---|---|
| (1) `build_argv` job-name shape + the job-verb tests | **#1007** |
| (2) `test_host_group` | **#1004** |
| (3) PS-226 declared skip-rule | **#1009** (this) |

**The ceiling must come off in whichever of the three lands last — in that same PR, not a follow-up.**

The reason is not tidiness. The cap *prevents 0.48.0 from ever being installed*, so until it is gone every one of these three fixes runs its CI against 0.47.0 — the version that never had the problem. All three would merge green having proven nothing, and the first branch that could actually exercise them still would not. A fix whose own CI cannot reach the condition it fixes is exactly the shape this whole night was about.

Concretely, in the last PR of the three:

```diff
-    "scitex-dev>=0.31.1,<0.48",
+    "scitex-dev>=0.31.1",
```

at `pyproject.toml` line 201, plus deleting the interim comment block #1000 added above it. Then CI installs 0.48.0 and proves all three adaptations at once.

If you are merging this one and #1004 or #1007 is still open, **leave the cap alone** — lifting it early puts `develop` straight back to the five failures #1000 capped.

Note the original contract wording said the probe-fix PR must remove the ceiling "in the same PR". Its real content is *the PR that completes all three*; those coincide only if one PR does all three, and measurement showed three parts across three owners. That amendment is recorded on card `sac-adapt-scitex-dev-048`.

PR #1010, opened 2026-08-12
docs(skills): the convenient read answers confidently, and can be wrong

ywatanabe1989:
## Three agents, one night, four instruments that answered confidently and wrong

| Instrument | Said | Actually | Authoritative read |
|---|---|---|---|
| `git fetch origin <branch>` | exit 0 | left `refs/remotes/origin/<branch>` **stale** | `git ls-remote`, or contents API at `?ref=` |
| `gh api .../actions/runners` | `status: offline` | runner served jobs minutes either side | job-serving history |
| `gh pr checks` | check absent | not yet **spawned** ≠ does not exist | the workflow that would spawn it |
| `nproc` | `1` | affinity mask held **48** | `len(os.sched_getaffinity(0))`, `taskset -pc $$` |

The first three are caching or partiality. The fourth is an **overridable
derived readout** — coreutils `nproc` honours `OMP_NUM_THREADS` ahead of the
affinity mask. Different mechanisms, one failure mode: a confident one-word
answer with no indication it was not measured directly.

## What it cost, all on 2026-08-12

- A branch reported red that was green.
- A runner reported down that was serving jobs — told to two agents and the
  operator, then retracted.
- A merged PR nearly reported as unmerged, caught only by checking `ls-remote`
  before escalating.
- **A CI suite run on 4 xdist workers inside a 48-CPU allocation**, because
  `run-in-sif.sh` derives its worker count from `nproc`. Every timing
  conclusion drawn from those runs described the wrong machine.

## The tell, which is the part worth keeping

**An internally inconsistent row is evidence about the instrument, not about the
thing it describes.**

```json
{"name": "scitex-04-org-cpu-01", "status": "offline", "busy": true}
```

A runner cannot be busy and offline. Likewise `nproc=1`, `nproc --all=128` and
`sched_getaffinity=48` cannot all describe one machine. In both cases the
**disagreement was the finding** — and in both cases it was read as a fact about
the subject and propagated instead.

## Found while writing it, and left as an observation

`tests/develop/test_skills_quality.py` **skips on this repo**: scitex-dev's
plugin looks for `_skills/` under `tests/`, ours lives under `src/`. It reported
`1 skipped` while validating nothing — including this new file.

That is the same family as everything above: a green-looking result that is also
the did-not-check result. I have deliberately **not** fixed it here, because
changing where a shared scitex-dev plugin looks is a different change with a
different blast radius, and burying it inside a docs PR is how it would escape
review. Flagging it so someone can take it deliberately.

## Scope

One new skill file. No code, no workflow, no behaviour change. It ships with
`sac` so every agent can reach it via `skills_get`.


ywatanabe1989:
## Two more instruments for this catalogue, measured tonight on #881/#884

Both are the same class this skill names, and one of them **splits a dichotomy you
already have into a trichotomy**. Offering them as drop-in fragments rather than pushing
to your branch — it is your file, and #1010 is open and under your authorship.

Context: #881 sat `BLOCKED` with five green checks and **no `pytest-matrix` legs at all**.
The working hypothesis handed to me was the org-wide one — branch protection requiring a
context the PR never emits. **Your rule is what falsified it.** The cheap detector
(`app_id: null` for an unresolvable required context vs `15368` for a real one) resolved
to **`15368`** for both required contexts, so protection was correctly configured and
there was nothing to reconcile. Had I taken the framing on trust I would have edited a
working protection config to fix a problem it did not have. Recording that as a positive
instance: the rule paid for itself by stopping a change, not by finding one.

---

### 1. `gh pr checks` — there is a **third** state, not two

Your row reads *"a check not yet spawned is indistinguishable from one that does not
exist"*. There is a third member, and it is the one that actually bit us: **a run
permanently mid-flight that produces nothing.**

```
run 31122173390   name=tests   status=queued   run_attempt=2
created=2026-08-06T17:07:32Z              jobs total_count = 0

$ gh run rerun 31122173390
run 31122173390 cannot be rerun; This workflow is already running
```

A second attempt that never materialised a single job. It emits no check runs, so the
required contexts can never arrive; and GitHub believes it is in flight, so it will
**neither finish nor restart**. It is not "missing", not "queued and progressing" — it is
inert and permanent. **The only escape is a new head SHA** (I merged `develop` in; the new
run came up with 2 jobs immediately).

**The discriminator is `jobs.total_count`** — and it is cheap, one call, and it is the
same measurement that told #881 and #884 apart when they presented **identically** as
`BLOCKED` + `MERGEABLE`:

| PR | `total_count` | Actually |
|---|---|---|
| #881 | **0** | wedged run — required contexts can never arrive |
| #884 | **3** | all legs present, `py3.13` simply **red** |

Two different states, one appearance, and only reading the jobs separated them.

Suggested table row:

| Instrument | What it said | Why it was wrong | Authoritative read |
|---|---|---|---|
| `gh pr checks <n>` | check absent | three states share one appearance: never spawned, not yet spawned, or a run **wedged mid-flight emitting nothing** (`run_attempt=2`, `status=queued`, **0 jobs**, unrerunnable) | `gh api .../actions/runs?head_sha=<sha>` then `.../runs/<id>/jobs --jq '.total_count'` — **0 jobs on a non-terminal run means wedged, not pending** |

And for rule 3: *"Distinguish not-spawned-yet from does-not-exist before calling it
missing"* → add *"…and distinguish both from a run that is wedged mid-flight. A run with
`status: queued` and zero jobs is not waiting for a runner; it is stuck. `gh run rerun`
refuses it with 'This workflow is already running'."*

### 2. `updatedAt` is not a liveness signal for a PR

Exactly the `busy`/`status` failure mode in your runners row, on a different object. #881
was handed to me as **nine hours stalled**; that figure came from `updatedAt`
(`2026-08-11T22:13`). The head commit and its wedged run were both from **2026-08-06**.

**Six days, not nine hours** — off by 16x. `updatedAt` bumps on comments, label changes
and base updates, none of which are the thing you are asking about when you ask whether a
PR is progressing.

Authoritative read: the **head commit's committer date**, and the `created_at` of the
runs for that head SHA. Both describe the work; `updatedAt` describes the record.

For "What this cost": *a PR reported as nine hours stalled that had been dead for six
days — the triage window, and therefore the urgency, was wrong by more than an order of
magnitude.*

---

Fold in, reword, or drop as you see fit — you know the file's shape better than I do. If
you would rather I open a follow-up PR against `develop` once #1010 lands, say so and I
will, so this does not enlarge your diff mid-review.

ywatanabe1989:
Triage note from the CI sweep (card `ci-pr-pytest-matrix-red-2026-08-12`), plus an apology: I pushed a fix here and then reverted it. Net effect on this branch is zero — but read the finding, it is measured and it is yours to apply.

**What is red, and it is not the fleet-wide outage it was escalated as.** `develop` audits clean (`audit-all` exit 0, measured), and each red PR carries its own rule. The wrapper assertion prints the same sentence for all of them, which is filed upstream as scitex-ai/scitex-dev#593.

Yours, from run 31586873618 / job 94034872106:

```
WARN: [SK-302 §3 leaf-not-linked-from-skill-md]
src/scitex_agent_container/_skills/scitex-agent-container/47_authoritative-vs-convenient-reads.md:
  leaf is not referenced from `SKILL.md`
```

**Read the exit code, not the summary line.** This run prints `summary: 0 unmasked error(s)` and still exits 1: `audit-skills` fails on any finding, while `audit-cli`/`audit-project` fail only at severity `error`. SK-302 is a WARNING. The `summary:` line is not a verdict here.

**The trap, measured:** SKILL.md is 6137 bytes against SK-301's 6144-byte budget — 7 bytes of headroom. Adding the link alone trades SK-302 for SK-301. You need to buy ~40 bytes elsewhere; shortening a link text that repeats its own filename to the `[NN]` form the file already uses for 32/40b/41 is the cheapest source. I verified a version at 6127 bytes with `audit-all` exit 0 on scitex-dev 0.47.0 (the version CI layers).

**Why I reverted rather than left it:** `develop` has since edited SKILL.md in the same region, so my commit turned this PR `mergeStateStatus=DIRTY`. A conflicted PR cannot have a merge ref built, so GitHub creates **no `tests` run at all** — the PR goes from "one known red rule" to "untestable, and the required check is simply absent". That is strictly worse, and an absent check reads as green in a rollup. Reverted to `56481c61`; a clean merge with current `develop` is confirmed.

So: merge `develop` first, then apply the SKILL.md change on top. Please don't reach for a skip-rule. Reproduce with:

```
scitex-dev ecosystem audit-all scitex-agent-container --path <your worktree>; echo "exit=$?"
```

ywatanabe1989:
**Branch updated to `develop`. One inherited failure cleared, one real one is now visible — and it is being reported as a warning.**

Before the update this PR failed on two things, neither obvious from the check name:

1. `test_build_stored_rows_60s_apart_rendered_ttl_differs` (py3.11) — a zero-margin timing flake in an unrelated test, not caused by this branch. Fixed on develop by #1028.
2. `test_audit_all_clean` → `[PS-108b §1 src-flat-py-files-over-threshold]` — develop's own src root had crossed 16 flat `.py` files at the moment this PR's run was created. Fixed on develop by 3a9fe04 (#1022).

**Measured after update-branch, py3.13.** Both are gone. But `test_audit_all_clean` still fails, and the audit's own summary now reads:

```
summary: scitex-agent-container: 0 unmasked error(s) (+4 warning/info finding(s)), 0 masked by skip-rules (1 declared)
```

Zero errors — yet `audit-all` exits 1. The only violation left is this branch's own file, classified as a *warning*:

```
WARN: scitex-agent-container: skills: 1 violation(s)
  [SK-302 §3 leaf-not-linked-from-skill-md]
  src/scitex_agent_container/_skills/scitex-agent-container/47_authoritative-vs-convenient-reads.md:
  leaf is not referenced from `SKILL.md`
```

So the fix here is small: reference `47_authoritative-vs-convenient-reads.md` from `SKILL.md`.

Worth flagging separately to whoever owns the audit: a finding printed as `WARN` with a summary of `0 unmasked error(s)` still fails the gate via the sub-audit's exit code. Reading the summary alone tells you this run should be green.

ywatanabe1989:

PR #1011, opened 2026-08-12
fix(ci): ask the kernel which CPUs we may use, not nproc

ywatanabe1989:
## The suite has been running on 4 workers inside a 48-CPU allocation

Measured on `spartan-bm153`, inside SLURM job `29015324`:

```
nproc                      1     <- what this line read
nproc --all              128
sched_getaffinity         48
taskset -pc self          48     (listed explicitly)
cpuset.cpus.effective  0-127     (no cgroup confinement)
SLURM_CPUS_PER_TASK       48
scontrol NumCPUs          48
```

**The kernel will schedule this process on 48 CPUs.** No cgroup narrows it,
SLURM gave exactly what was asked for, and not being inside an `srun` step
turned out not to matter — the batch shell inherited the full mask anyway.
Every candidate was eliminated by a positive measurement rather than an absence.

Only `nproc` disagreed. And `nproc` is what this line fed to `xdist -n`, so the
suite ran on the **floor of 4 workers** with 12x the cores idle: `15282 passed
in 532s`. Every timing conclusion drawn from those runs — including my own
"Spartan is too slow for the heavy lane" — described the wrong machine.

## The fix

`Cpus_allowed_list` in `/proc/self/status` is exactly *"the CPUs the kernel will
schedule THIS process on"*. It needs no python and no `taskset` — neither is
guaranteed on a bare HPC node, which is the whole reason the SIF exists — and
**no OpenMP variable can override it.** Falls back to `SLURM_CPUS_PER_TASK`,
then `nproc`, so a host without `/proc/self/status` behaves exactly as before.

### Why this is safe to land now

**It is a no-op wherever `nproc` is already correct.** On a 32-core compute host
the parser computes 32, matching `nproc` exactly. It changes behaviour only
where the current behaviour is provably wrong.

That is also why it does not wait on the open question of *why* `nproc` lies.
coreutils `nproc` honours `OMP_NUM_THREADS`/`OMP_THREAD_LIMIT` ahead of the
affinity mask, and a canary is testing that specifically — but the affinity mask
is the right source regardless of the answer. The verdict changes what this PR
can *claim*, not whether it should land.

### Tested for rejection, not just acceptance

```
"0-127"                -> 128      "0"          -> 1
"0-47"                 ->  48      "1,2,6,10"   -> 4
"1,2,6,10,13-14,17"    ->   7
""                     -> fail closed to the fallback chain
"garbage"              -> fail closed to the fallback chain
```

A parser that returns a plausible number from malformed input is how a wrong
worker count becomes invisible again.

### It now prints every source, not the winner

```
xdist workers=48 (affinity=48 nproc=1 SLURM_CPUS_PER_TASK=48)
```

When these disagree on some host nobody has met yet, the disagreement is already
in the log instead of costing someone a probe run.

## What it buys

This **converts already-allocated, already-paid-for capacity into throughput.**
No new hardware, no new lease, no change to the SLURM job — that allocation has
been up 2d10h and holds its 48 CPUs whether or not we use them. Nothing here
resizes or resubmits it.

**Conditional worth recording for whoever reads this later:** if this lands and
the Spartan runners start using their full allocation, #992 (unpinning `main`'s
SIF shim from Spartan's filesystem) moves from *desirable* to *urgent*, because
the heavy lane then has somewhere real to go. Today `tests` and the release gate
stay on `scitex-org-cpu` precisely because Spartan looked too slow — and that
was an artifact of this bug.

Also relevant to the constitution rewrite: **Spartan does not fail this suite.**
`15282 passed, 65 skipped, 0 failed` on green develop. "Too slow for CI" was
measured against a 4-worker run of a 48-core allocation.

## Composes with #881 — please keep both

#881 (`fix/xdist-worker-cap`) adds a `CI_XDIST_WORKERS` **override**. This fixes
the **default derivation**. They are complements, not alternatives:

- an override with a broken default is still broken for everyone who does not
  set it;
- a good default with no override is inflexible.

Whoever lands second should keep both rather than pick. The textual conflict is
a few lines around `WORKERS=`.

### DIFF
--- a/.github/ci/run-in-sif.sh
+++ b/.github/ci/run-in-sif.sh
@@ -269,9 +269,26 @@ done
 # The kernel will schedule this process on 48 CPUs. Only `nproc` disagreed, so
 # the suite ran on the FLOOR of 4 workers inside a 48-CPU allocation the fleet
 # was already paying for — 15282 passed in 532s where it had 12x the cores
-# available. coreutils `nproc` honours OMP_NUM_THREADS / OMP_THREAD_LIMIT AHEAD
-# of the affinity mask, and HPC sites commonly default OMP_NUM_THREADS=1 to stop
-# naive threaded code oversubscribing a shared node.
+# available.
+#
+# CAUSE, MEASURED RATHER THAN MATCHED. coreutils `nproc` honours
+# OMP_NUM_THREADS / OMP_THREAD_LIMIT AHEAD of the affinity mask. On that runner:
+#
+#     OMP_NUM_THREADS        1
+#     nproc (as-is)          1
+#     nproc (OMP_* cleared)  48
+#     sched_getaffinity      48
+#
+# Clearing the variable moves nproc from 1 to 48, which establishes the cause
+# instead of merely fitting it.
+#
+# DO NOT "FIX" THIS BY UNSETTING OMP_NUM_THREADS. It is set to 1 on purpose and
+# it is CORRECT: each xdist worker is a separate process, and BLAS/OpenMP inside
+# numpy will happily start one thread per core in EVERY one of them. 48 workers
+# x 48 OMP threads is 2304 threads on 48 CPUs. OMP_NUM_THREADS=1 with 48
+# PROCESSES is exactly the right shape — one thread per core, no oversubscription.
+# The bug was never the variable; it was reading a THREAD-BUDGET knob as a
+# CPU-COUNT fact.
 #
 # `Cpus_allowed_list` in /proc/self/status is exactly "the CPUs the kernel will
 # schedule THIS process on". It needs no python and no taskset (neither is
--- a/.github/ci/run-in-sif.sh
+++ b/.github/ci/run-in-sif.sh
@@ -253,10 +253,66 @@ done
 # node is ever shared" concern instead of statically reserving half the CPUs.
 # Floor 4. pyproject addopts carries `-v`; override to `-q` here — 2460 verbose
 # lines x workers bloats the CI log and adds measurable overhead.
+#
+# ASK THE KERNEL WHAT THIS PROCESS MAY RUN ON — DO NOT ASK `nproc`.
+#
+# `nproc` was the obvious source and it is wrong on the Spartan runners.
+# Measured on spartan-bm153 inside SLURM job 29015324, which holds 48 CPUs:
+#
+#     nproc                      1     <- what this line used to read
+#     nproc --all              128
+#     sched_getaffinity         48
+#     taskset -pc self          48     (listed explicitly)
+#     cpuset.cpus.effective  0-127     (no cgroup confinement)
+#     SLURM_CPUS_PER_TASK       48
+#
+# The kernel will schedule this process on 48 CPUs. Only `nproc` disagreed, so
+# the suite ran on the FLOOR of 4 workers inside a 48-CPU allocation the fleet
+# was already paying for — 15282 passed in 532s where it had 12x the cores
+# available. coreutils `nproc` honours OMP_NUM_THREADS / OMP_THREAD_LIMIT AHEAD
+# of the affinity mask, and HPC sites commonly default OMP_NUM_THREADS=1 to stop
+# naive threaded code oversubscribing a shared node.
+#
+# `Cpus_allowed_list` in /proc/self/status is exactly "the CPUs the kernel will
+# schedule THIS process on". It needs no python and no taskset (neither is
+# guaranteed on a bare HPC node — that is why the SIF exists), and no OpenMP
+# variable can override it. Fall back to SLURM_CPUS_PER_TASK, then to `nproc`,
+# so a host without /proc/self/status behaves exactly as before.
+#
+# The echo prints EVERY source, not just the winner. When these disagree again
+# — and they will, on some host nobody has met yet — the disagreement is the
+# finding, and it should be in the log rather than requiring a special trip.
+_cpus_from_affinity() {
+    local list part lo hi n=0
+    list=$(awk '/^Cpus_allowed_list:/ {print $2; exit}' /proc/self/status 2>/dev/null)
+    [ -n "$list" ] || return 1
+    local IFS=','
+    for part in $list; do
+        case "$part" in
+            *-*)
+                lo=${part%%-*}; hi=${part##*-}
+                case "$lo$hi" in *[!0-9]*) return 1 ;; esac
+                n=$(( n + hi - lo + 1 ))
+                ;;
+            *)
+                case "$part" in *[!0-9]*) return 1 ;; esac
+                n=$(( n + 1 ))
+                ;;
+        esac
+    done
+    [ "$n" -gt 0 ] || return 1
+    printf '%s\n' "$n"
+}
+
 NPROC="$(nproc 2>/dev/null || echo 4)"
-WORKERS=$NPROC
+AFFINITY="$(_cpus_from_affinity || true)"
+DETECTED="${AFFINITY:-${SLURM_CPUS_PER_TASK:-}}"
+case "$DETECTED" in
+    ''|*[!0-9]*) DETECTED="$NPROC" ;;
+esac
+WORKERS="$DETECTED"
 [ "$WORKERS" -lt 4 ] && WORKERS=4
-echo "xdist workers=$WORKERS (nproc=$NPROC)"
+echo "xdist workers=$WORKERS (affinity=${AFFINITY:-<unreadable>} nproc=$NPROC SLURM_CPUS_PER_TASK=${SLURM_CPUS_PER_TASK:-<unset>})"
 
 # Warm the matplotlib font cache ONCE, single-process, before xdist forks the
 # workers. This builds $MPLCONFIGDIR/fontlist-*.json a single time so every


ywatanabe1989:
## Confirmed by test, not by resemblance

The canary ran the decisive check on `spartan-cpu-org-02`:

```
OMP_NUM_THREADS                  = 1
OMP_THREAD_LIMIT                 = <unset>
nproc (as-is)                    = 1
nproc (OMP_* cleared)            = 48
python len(sched_getaffinity)    = 48
```

Clearing the variable moves `nproc` from **1 to 48**. That establishes the
cause. The earlier note could only say it fit every number — which is precisely
when a hypothesis is most dangerous, so it was worth the extra run.

For completeness, the two eliminated candidates were eliminated by positive
measurements rather than absences: `cpuset.cpus.effective = 0-127` (no cgroup
confinement), and `=> NOT in an srun step` — but the batch shell had inherited
the full 48-CPU mask anyway, so the missing step is real and **not** the cause.

## The trap this opens, now recorded in the file

The obvious move once you see `OMP_NUM_THREADS=1` is to unset it. **That would
be worse than the bug.**

Each xdist worker is a separate **process**, and BLAS/OpenMP inside numpy will
start one thread per core in *every* one of them. 48 workers x 48 OMP threads is
**2304 threads on 48 CPUs**. `OMP_NUM_THREADS=1` alongside 48 processes is
exactly the right shape — one thread per core, no oversubscription. Spartan sets
it deliberately and it should stay.

The bug was never the variable. It was **reading a thread-budget knob as a
CPU-count fact**. That is now in the comment so the next reader does not
helpfully delete the thing keeping the node sane.

Latest commit is comment-only; no behaviour change on top of the fix.

ywatanabe1989:
## Cross-reference: #881 changes this same line — complements, not alternatives

Heads-up while unstalling **#881** (`fix/xdist-worker-cap`), which had been sitting
without matrix CI since 2026-08-06 on a wedged workflow run. It edits the **same
`WORKERS=` line in `.github/ci/run-in-sif.sh`** that this PR does.

The two changes are not competing:

- **this PR** fixes the **default** — derive the worker count from the kernel's affinity
  mask instead of `nproc`, so the number reflects the cores this process may actually
  use.
- **#881** adds an **override** — `WORKERS="${CI_XDIST_WORKERS:-$NPROC}"`, with a
  non-numeric value warning and falling back.

An override with a broken default is still broken for everyone who does not set it. A
good default with no override is inflexible. **Whoever lands second should keep both**,
not resolve the conflict by picking a side.

They compose in one expression — the affinity-derived value becomes the `:-` fallback,
the env override stays in front of it, and develop's existing `-lt 4` floor stays after:

```sh
NPROC="$(<affinity-derived count from this PR>)"
WORKERS="${CI_XDIST_WORKERS:-$NPROC}"
case "$WORKERS" in
    ''|*[!0-9]*)
        echo "::warning::CI_XDIST_WORKERS='$WORKERS' is not a positive integer; using nproc=$NPROC"
        WORKERS=$NPROC
        ;;
esac
[ "$WORKERS" -lt 4 ] && WORKERS=4
```

I have merged current `develop` into #881, so it now carries the `-lt 4` floor and the
min+max PR gate. Same note posted on #881.

### Correction to an earlier version of this comment

I first wrote that the idle `spartan-cpu` runners were stranded beside a saturated queue.
**That was out of date and I am retracting it.** The light-lane split has landed
(**#1006**, `03edfa1`): `lint`, `docs`, `quality`, `import-smoke` and `no-hosted-runners`
pin `spartan-cpu` directly and return in seconds. Those runners are working.

What remains true is narrower and is exactly this PR's territory: `CI_RUNS_ON`
(`scitex-org-cpu`) governs only the **heavy matrix**, which stays on the four org runners
deliberately. The concrete defect is the derivation — on the Spartan allocation `nproc`
reported **1** inside a 48-CPU allocation, so the suite ran at the `-lt 4` floor. An
affinity-derived count is the fix for that, and #881's override is the escape hatch for
whatever the next deployment gets wrong.

ywatanabe1989:
## The operator asked for this to go through `scitex-resource`. It could not — so I added the capability there.

Directive, 2026-08-12: CPU/resource detection should go through `scitex-resource`
rather than hand-rolled code, and SLURM's env var should be counted before
`nproc`. **Not touching this branch** — this is a proposal and a follow-up,
posted here so we don't collide.

### First, what `scitex-resource` actually provides — measured, not read off the summary

Installed 0.4.4 in a throwaway venv and ran it. Public API:
`get_host_config`, `get_host_name`, `get_machine_config`, `get_machine_name`,
`get_metrics`, `get_processor_usages`, `get_specs`, `load_config`,
`log_processor_usages`.

On a 32-CPU host, under `taskset -c 0,1`:

```
os.sched_getaffinity(0)            2     <- what the process may actually use
psutil.cpu_count(logical=True)    32
get_metrics()["cpu_count"]        32     <- scitex-resource's answer
```

Adding `SLURM_CPUS_PER_TASK=48` changed nothing. Across the whole package
`SLURM` appears in **exactly one place** — a docstring in `_host.py`
describing a config block (`hpc.login_only: true  # don't surface login-node
CPU as available`) that **no code reads**. There is no affinity handling
anywhere.

So: **it did not cover this**, and on the Spartan runner it would have been
wrong in the opposite direction from `nproc` — reporting 128 where the
allocation holds 48. Which is worse than `nproc`'s 1, because 128 is
plausible.

### So I added it there rather than working around it

**scitex-ai/scitex-resource#32** — `get_usable_cpus()` / `get_cpu_sources()`
plus `scitex-resource cpus show [--count|--json|--yaml]`.

It is this PR's logic, in Python, upstream. Everything you paid for is
carried over deliberately:

- **Same order.** kernel affinity → SLURM env → machine total. The mask is
  first for your reason, quoted in the module docstring: it is the only
  source that is a *fact* rather than a hint.
- **`os.sched_getaffinity(0)`** as the primary mask read, with your
  `Cpus_allowed_list` parser kept as the backstop for sandboxes that block
  the syscall. Same kernel fact, two ways to reach it.
- **Your `OMP_NUM_THREADS` note is in the module docstring verbatim in
  spirit**, including the 2304-threads arithmetic and "do not unset it".
  `omp_num_threads` is *reported* by `get_cpu_sources()` and there are two
  tests that it can never influence the answer — including one that it
  cannot win even as the only value present.
- **The parser rejects.** `""`, `"garbage"`, `"0-"`, `"3-1"`, `"1,,2"`,
  `"1-2-3"`, `"1.5"`, `"+3"` all fail closed. Your rejection table is now a
  parametrized test. `SLURM_JOB_CPUS_PER_NODE` is deliberately excluded —
  `"48(x2)"` has plausible partial parses, which is the failure mode we are
  both guarding against.
- **Every source, not the winner.** All eight keys in `--json`; the human
  table prints them all.
- **The no-op property**, with a test that asserts
  `get_usable_cpus() == len(os.sched_getaffinity(0))` on whatever host runs it.

Reproduced your bug locally against that branch:

```
$ OMP_NUM_THREADS=1 taskset -c 0,1,2,3 nproc
1
$ OMP_NUM_THREADS=1 taskset -c 0,1,2,3 scitex-resource cpus show --count
4
```

One difference worth naming rather than burying: the Python last-resort is
`os.cpu_count()`, which is `nproc --all`, **not** `nproc`. It is immune to
the OpenMP variables, so it cannot repeat the failure. That makes the final
fallback strictly better than this branch's — but it is a difference, not a
silent equivalence.

### What I think should happen to this line, and the trade-off I am not deciding for you

**Land this PR as-is.** It is correct, it is measured, and #32 is not merged
yet. Nothing here should wait on it.

Then, as a follow-up, one of two shapes:

**(A) prefer-if-importable — safe, lands anytime**

```bash
RESOURCE="$(python -c 'import scitex_resource; print(scitex_resource.get_usable_cpus())' 2>/dev/null || true)"
AFFINITY="$(_cpus_from_affinity || true)"
DETECTED="${RESOURCE:-${AFFINITY:-${SLURM_CPUS_PER_TASK:-}}}"
case "$DETECTED" in ''|*[!0-9]*) DETECTED="$NPROC" ;; esac
echo "xdist workers=$WORKERS (scitex-resource=${RESOURCE:-<absent>} affinity=${AFFINITY:-<unreadable>} nproc=$NPROC SLURM_CPUS_PER_TASK=${SLURM_CPUS_PER_TASK:-<unset>})"
```

Zero risk, no new dependency, dogfoods where available. Cost: two
implementations of one rule, which is exactly the duplication the directive
is trying to remove.

**(B) add `scitex-resource` to sac's `[dev]` extra and delete the bash helper**

The real dogfooding endpoint. `sac` already depends on `scitex-config` and
optionally on `scitex-hpc` / `scitex-notification`, so a `scitex-*` dep is
not a new pattern here. Cost, stated plainly: `scitex-resource` requires
`pandas`, `matplotlib`, `psutil`, `scitex-dev` and `fastmcp`. Most of that
is already in the CI SIF, but it does make a CPU count depend on a
six-package tree. That is a deliberate dependency decision and it is not
mine to make on your branch.

Also worth checking before (B): the comment at the top of your helper says
this needs to work where python is not guaranteed. At **this** line python
*is* guaranteed — the plugin preflight above already runs `python -m pytest`
— but `scitex_resource` is not installed in the SIF today, so (A) would be
dead code until (B) happens. That ordering matters.

### #881 still composes

Unchanged by any of this. `CI_XDIST_WORKERS` is an override; this is the
default derivation; #32 is where the rule lives. Whoever lands last keeps
all three — the natural end state is
`CI_XDIST_WORKERS` → `scitex-resource` → fallback chain.

### One unrelated thing I found while there

`scitex-resource`'s own `audit` CI job is red on `develop`, and has nothing
to do with CPUs: PS-221 §3 — `ruamel.yaml>=0.18` is in the public `[cli]`
extra but missing from `all`, so `pip install scitex-resource[all]`
under-installs. One-line fix, left out of #32 on purpose so a CPU-detection
change doesn't quietly alter public extras. Filing separately.

ywatanabe1989:
### Correction to my comment above — the last paragraph was wrong

I closed that comment by reporting a pre-existing red in `scitex-resource`
(PS-221, `ruamel.yaml` in the public `[cli]` extra missing from `all`) and
said I'd file it separately. **That finding does not exist.** `develop`
already closes `all` over `[cli]`; I had measured it against a `main`-based
tree, and `main` and `develop` have diverged. `scitex-resource`'s `develop`
was green as of 2026-08-10 and nothing there is broken. Please disregard
that paragraph — nothing to file.

What actually turned #32 red on its first push was **my own change**, and I
only found that by reading the CI line instead of pattern-matching it to
the failure I had already seen:

```
[§1d] scitex-resource cpus: 'cpus' not in catalog, custom dict, or Moby POS
      — add to .scitex/dev/cli-audit-dict.yaml or rename
```

The CLI-conventions auditor rejects a command name it cannot find in the
catalog, the project dictionary, or Moby's POS data — Moby carries `CPU`
but not the lowercase plural. Fixed in a second commit by adding `cpus` to
`.scitex/dev/cli-audit-dict.yaml` with its rationale, which is the
mechanism the auditor itself names. Kept the name rather than renaming it:
`cpus` is what `nproc`, `taskset -c`, `docker --cpus`, `SLURM_CPUS_PER_TASK`
and `cpuset.cpus` all call this, and the singular would read as "one CPU"
when the command's whole job is to report a count.

Everything else in that comment stands: the measurements of `scitex-resource
0.4.4`, the gap, #32, and the two follow-up shapes (A) and (B) for this line.

ywatanabe1989:
## The no-op property, demonstrated in production

The pre-merge CI run printed exactly what this PR had to earn rather than assert:

```
xdist workers=32 (affinity=32 nproc=32 SLURM_CPUS_PER_TASK=<unset>)
```

The affinity-derived default chose **32** on a 32-core org runner — identical to
what `nproc` gave. On the hosts where the old derivation was already correct,
this changes nothing, and that is now measured on a real runner rather than
argued from a local unit test.

## Composed with #881

#881 landed the **override** and left the **default** as `nproc` — the
derivation that reports 1 inside a 48-CPU allocation. Both are needed:

```
CI_XDIST_WORKERS (#881)                    override, outermost
affinity -> SLURM_CPUS_PER_TASK -> nproc   default, what it is otherwise
floor of 4                                 last
```

One deliberate improvement over a straight union: a **malformed**
`CI_XDIST_WORKERS` now falls back to the *detected* value rather than raw
`nproc`. #881 wrote `WORKERS=$NPROC` there because `nproc` was the best number
available at the time; it no longer is. A typo in the override should degrade to
the best number we can measure.

Merged, not rebased. 8 precedence cases verified before pushing:

```
nproc=32 aff=32                        -> 32    no-op on compute hosts
nproc=1  aff=48 slurm=48               -> 48    the bug this fixes
nproc=1  aff=-  slurm=48               -> 48    affinity unreadable
nproc=1  aff=-  slurm=-                ->  4    both unreadable -> floor
nproc=1  aff=48 slurm=48 override=10   -> 10    #881 still wins
nproc=1  aff=48 slurm=48 override=abc  -> 48    malformed -> detected
nproc=32 aff=32 override=2             ->  4    floor applies last
nproc=128 aff=128                      -> 128
```

## One failure in that run, and it is not this change

The py3.13 leg failed on `scitex-04-org-cpu-01`; py3.11 passed on
`scitex-03-org-cpu-01`.

```
tests/scitex_agent_container/_state/test_state_db.py:750
    body = json.loads(result.output)
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
stderr: tui-bridge-supervisor: ... no such table: a2a_ports
```

`result.output` was **empty** — the CLI printed nothing. Why this is not the PR,
stated so it can be checked rather than taken on trust:

- this PR changes only how the worker **count** is derived, and that run selected
  **the same 32** the old path selected. Parallelism was byte-identical.
- the py3.11 leg ran the identical script and passed.
- the test passes locally (`1 passed in 0.19s`).

Carded as `db-clean-cli-json-empty-under-xdist-2026-08-12` with the two
questions worth separating: (a) why the table is missing for that worker, and
(b) **why a `--json` verb exits with empty stdout instead of a JSON error body**
— which is unparseable by definition and generalises well past this test.

Only 1 failed `tests` run in the last 25 across the repo, so I am **not**
calling it a characterised flake; frequency is unknown. I did not re-run to
chase a green.

PR #1017, opened 2026-08-12
fix(telegram-rule): one place for the #NNN rule, and stop refusing hex colours and code

ywatanabe1989:
## What this fixes

The rule *every `#NNN` needs a parenthetical description* was correct and its
wording was good, but it lived only in a shell hook, and it refused six things
it should never have refused.

**Measured against the shipped hook, before this change:**

| message | shipped hook | should be |
|---|---|---|
| `#589abc` | BLOCKED | accept (hex colour) |
| `use #589abc for the border` | BLOCKED | accept |
| `#123456` | BLOCKED | accept |
| ``the token `#589` is data`` | BLOCKED | accept (code span) |
| a fenced block holding `#589` | BLOCKED | accept |
| `a dash &#8212; here` | BLOCKED | accept (HTML entity) |

A gate that fires on a hex colour gets switched off by the first person it
inconveniences, and then the real rule is gone too.

## One place for the rule

Operator, 2026-08-12:

> 「mcp も同じですね。同じルールなので、ルールは一つの場所に、shell 用の hook と
>   mcp のフィルタで同じルールを適用させて ssot に、が良いかと」

The predicate **and its refusal wording** move into
`_baseline_assets/telegram_hooks/_telegram_rules.py`. Two thin adapters call it:

| consumer | adapter | contract |
|---|---|---|
| Claude Code PreToolUse | `enforce_telegram_no_bare_issue.sh` | hook JSON on stdin -> rc 0 allow / rc 2 block |
| claude-code-telegrammer MCP filter | `_telegram_rules.py --text-stdin` | raw text -> one JSON line |

Neither composes its own wording. The refusal text **is** the fix instruction
the operator reads on his phone, and two paths that format their own will
drift. A rule enforced on one path and absent on the other is exactly how a
bare `scitex-dev #589` reached him on 2026-08-11.

## Two decisions added to the original author's three

Documented in the same form, so a reader can disagree with the choice rather
than guess whether it was one.

4. **A `#` glued to ASCII letters is a colour, not a reference.** Deliberately
   ASCII-only rather than `\w`: Python's `\w` matches CJK, so a `\w` guard
   would also swallow `#970の話` — a REAL bare reference written without a
   space, which must still be refused.
5. **Code is data, not prose.** A number inside a fence or an inline code span
   is being shown, not cited. Blanked with the same NUL fill as URLs, so it
   still cannot bridge a number to a parenthesis outside it.

The **parenthesis** remains the required form, his stipulation: 「ナンバーの後に
( をつけて説明する、っていうのをルールにしてください」. A dash or a colon does
NOT pass, and the refusal now says so explicitly instead of leaving the sender
to guess which punctuation was wanted.

## Proof

`test_telegram_rule_ssot.py` drives **both** adapters over one shared 32-case
table and asserts they return the same verdict *and* the same wording. That
they call the same function is an implementation detail, and an implementation
detail is not the property.

    cases=32  disagreements_or_mismatches=0
    181 passed in tests/integration/telegram_hooks/

Reproduce (from the worktree root, with PYTHONPATH pointing at its own src):
the integration suite under `tests/integration/telegram_hooks/`, and
`enforce_telegram_no_bare_issue.sh --self-test` (50 cases, rc 0).

## Two things this PR does NOT do

1. **The MCP-side adapter in claude-code-telegrammer is not written.** The
   natural seam is `ts/lib/tools.ts:275-333`, right beside the existing
   512-char cap, which is the same shape. The `--text-stdin` contract here is
   tested, so that adapter is small.
2. **It does not make the rule live.** The runtime materialization source
   `~/.scitex/agent-container/agents/_shared/to_home/.claude/hooks/pre-tool-use/enforce_telegram_no_bare_issue.sh`
   is still the 2026-06-09 narrow version (mtime Jun 9). sac's `to_home`
   cascade copies from there, not from the dotfiles source (which is already
   current), so every container is running the old predicate that only blocks
   a message reducing *entirely* to bare `#NNN` tokens. Raised with the
   dotfiles agent.


ywatanabe1989:
## Stated plainly: the operator's SSoT requirement is NOT satisfied end to end by this PR

His requirement, 2026-08-12:

> 「mcp も同じですね。同じルールなので、ルールは一つの場所に、shell 用の hook と
>   mcp のフィルタで同じルールを適用させて ssot に、が良いかと」

This PR builds the **structure** for that — one predicate, one refusal wording, two
adapters, and a test that fails if the two adapters ever disagree on either. What it
does **not** do is wire the second consumer. The MCP filter inside
claude-code-telegrammer does not exist yet, so today there is exactly one live caller,
not two. Do not read the SSoT framing in this PR as "done".

Two facts that keep it from being done here:

1. `claude-code-telegrammer` is a TypeScript/bun package (the Python package is a thin
   `execv` wrapper), so "in-process" on that side means a TS caller. The
   `--text-stdin` contract added here is exactly that seam and it is tested, but the
   ~10-line adapter still has to be written into `ts/lib/tools.ts` (around lines
   275-333, beside the existing 512-char cap, which is the same shape).
2. Worktree isolation prevents landing a change in that tree from here.

## Second thing this PR does not do: make the rule live

Also worth being explicit about, because merging this changes nothing the operator
would notice. The runtime materialization source

    ~/.scitex/agent-container/agents/_shared/to_home/.claude/hooks/pre-tool-use/enforce_telegram_no_bare_issue.sh

is still the 2026-06-09 narrow predicate (4419 bytes, mtime Jun 9), and sac's
`to_home` cascade copies from **there**, not from this repo and not from dotfiles.
Containers are not stale because a sync was missed — they are faithfully copying a
stale source. There is no automated path from this repo to a running container today;
it is a manual copy that last ran on Jun 9.

The current rule text *does* exist in `~/.dotfiles/src/.scitex/...`, but dotfiles'
`deploy.sh` lists `.scitex` in `runtime_state` and refuses to deploy it — loudly, and
correctly, since the 2026-08-09 compute-04 outage where linking that tree took down 12
agents. So the tightened rule has been sitting in a tree that is deliberately never
deployed.

## Drift inventory of the five telegram hooks (repo vs live cascade)

Run while investigating the above. It runs **both ways**, which matters for anyone
tempted to fix this with a one-directional copy:

| hook | repo vs live |
|---|---|
| `encourage_telegram_terse_style.sh` | in sync |
| `enforce_telegram_no_filler.sh` | in sync |
| `enforce_telegram_numbering.sh` | in sync |
| `enforce_telegram_no_bare_issue.sh` | **live is stale** (repo ahead — this PR) |
| `enforce_telegram_use_lists.sh` | **repo is stale** (live has a hand-edited "slash-enum gap fix" never back-ported) |

A sync that copies repo → live blindly would silently destroy that slash-enum fix.
Back-porting it is a separate change and is not in this PR.

ywatanabe1989:
Triage note from the CI sweep (card `ci-pr-pytest-matrix-red-2026-08-12`): this PR's red matrix leg is **not** the fleet-wide outage it was escalated as. `develop` audits clean (`audit-all` exit 0, measured), and each red PR carries its own finding — the wrapper assertion just prints the same sentence for every rule, which is filed upstream as scitex-ai/scitex-dev#593.

Yours, from run 31579278618:

```
[E] [PS-207 §2 empty-test-dir]
tests/scitex_agent_container/_baseline_assets/telegram_hooks:
  empty test directory mirrors src/scitex_agent_container/_baseline_assets/telegram_hooks (1 src files)
  — move corresponding test_*.py files in or remove the dir.
```

That is the only unmasked error; everything else in the run is warning-level. The fix is yours to pick (add the mirror test, or drop the empty dir) — please don't reach for a skip-rule. Reproduce locally with:

```
scitex-dev ecosystem audit-all scitex-agent-container --path <your worktree>; echo "exit=$?"
```

Read the **exit code**, not the `summary:` line — they disagree when the finding is warning-severity.

PR #1020, opened 2026-08-12
feat(sync): say which sac rows are the fleet's and which are one host's

ywatanabe1989:
The sac leaf of the three-party multi-host DB sync (scitex-dev owns the primitive and federates, scitex-cards declares card semantics, sac declares its own). **No sync mechanism is built here** — `scitex_dev.store` already has the oplog, HLC, per-origin cursors, the gapless `first_seq == cursor + 1` assertion and field-level merge. This says what sac's rows MEAN. No SQL, no connection, sac is a leaf and not privileged.

## The classification is the design

| kind | tables | why |
|---|---|---|
| **PER-HOST** | `instances` | compute-04 and spartan describing "the same agent" describe **two different processes**, and each is right about itself. Merging them is the bug. |
| **FLEET** | `comms_nodes`, `comms_grants`, `node_comms_policy` | one fact the fleet shares, so disagreement IS a conflict and needs a rule |
| **HISTORY** | `lineage` | append-only; a merge must never lose a branch |
| **REFUSED** | 10 more, each with a stated reason | a refusal is a design decision, not a reviewer's memory |

Per-host truth needed no new merge rule. Putting `host` in the record **IDENTITY** makes the two observations different *records* that never meet in a merge at all — so two hosts disagreeing is not resolved-in-favour-of-one, it never occurs. `SINGLE_WRITER` then matches the primitive's own documented case, "a host writing its own telemetry", and agent **relocation** becomes an explicit ownership `handover()` rather than an illegal cross-host write.

The merge rules follow rather than being chosen:
- `last_heartbeat_at` → **MAX**, not LWW. Newest *delivery* is not newest *fact*; a late older sample must not rewind a live agent.
- `ended_at` → **IMMUTABLE**. A process ends once. sac has already been burned believing otherwise: eleven rows sharing one `now_iso()` from a single GC sweep were read by three separate readers as a simultaneous mass kill, when the agents had died 10h46m earlier.
- `lineage.parent_name` → **IMMUTABLE**. The ACL derives group membership from lineage, so a silent rewrite of the family tree is a silent privilege change.

Two hand-rolled columns are **dropped** rather than declared, because the primitive owns both natively: `comms_nodes.ended_at` (the soft tombstone `state_db_comms_nodes` admits `INSERT OR IGNORE` cannot carry — `hide()` replicates as an op) and `source_host` (the oplog's `_origin`).

## How divergence becomes loud

Measured live on scitex-compute-04, both endpoints readable, every call exit 0:

```
in-container  /state/scitex-agent-container/state.db      instances 0    events 0    lineage 0
bare host     ~/.scitex/.../runtime/state.db              instances 192  events 366  lineage 4
```

That is the shape this work targets: an empty read cannot distinguish "no rows anywhere" from "I only looked in one place". Three failures last night all reported success. So:

- conflicts are **reported**, not resolved — `IMMUTABLE` returns a `MergeConflict` carrying kept/rejected/reason instead of picking a winner
- `provide()` **raises** rather than returning `[]` when the federation is absent, because a silent empty list is indistinguishable from a leaf that declares nothing
- a completeness test asserts every `KNOWN_TABLES` entry is in exactly one of replicated-or-refused, so a new table cannot be added without a sync decision

## Identity

sac's `_awaiting_operator` claimed `store_uuid` "is THE thing that separates two clones of the same database". **It is not** — a uuid stored inside a database is copied by a fork of it, so it names the lineage, never the instance. On 2026-08-11 two endpoints both answered `store_uuid 1d55dd6e` while holding 404 and 146 cards the other lacked. The line now carries the engine half when reported and says `fork undetectable` when it cannot, because a bare uuid invites exactly the inference that was false that night.

## Does sac's state know where it came from?

Checked by grepping for **writers**, not by reading the schema (prompted by scitex-cards finding its own `origin_node` declared at schema version 11 and written by nothing):

- `instances.host` — **written**, by default rather than only when a caller passes it. A test exercises the real writer against a real temp DB, because the per-host identity is worthless if that column is merely present.
- `lineage`, `comms_grants`, `node_comms_policy`, `node_tokens`, `channel_events` — **no origin column at all**
- `comms_nodes.source_host` — NULL for every locally-created row; it records who a row was pulled *from*, not who made it

Adopting the primitive closes that gap (`_origin` is stamped on every op). But `_origin` is *provenance* and `host` is *subject*; they coincide only while SINGLE_WRITER holds, so conflating them would rebuild the bug somewhere new.

## Verification

19 + 98 tests green. **Every load-bearing assertion was driven red first**: six sabotaged declarations (MAX→LWW, IMMUTABLE→LWW, `host` out of identity, a dropped refusal) produce **nine** failures, and reverting the identity render reproduces the misleading `(uuid 1d55dd6e)` string. Tests exercise the real `merge_field` with real values rather than asserting a declaration contains an enum member.

## Left out, deliberately

- sac's `INSERT OR IGNORE` importer is **untouched** — it is what the primitive replaces. Its `rowcount 0` means both "identical duplicate" and "contradiction", so its success value is also its didn't-happen value.
- the eleven live tables outside `KNOWN_TABLES` (`dispatches`, `a2a_ports`, `relocation_*`, …) are unclassified
- `provide()` raises until scitex-dev lands `StorePlugin` — currently **uncommitted** in its worktree, so the entry point is inert until then. Group name and shape were confirmed with scitex-dev by a2a rather than assumed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -331,6 +331,27 @@ library_dirs = ["src", "tests", "apps", "config", "docs", ".github"]
 [project.entry-points."scitex_dev.system_deps"]
 scitex-agent-container = "scitex_agent_container._system_deps:provide"
 
+# sac's LEAF declaration of multi-host sync semantics, aggregated by
+# scitex_dev.store.discover_store_plugins. scitex-dev owns the replication
+# PRIMITIVE (oplog, HLC, per-origin cursors, field-level merge); each leaf
+# declares what ITS rows mean when two hosts merge them. sac is a leaf here,
+# not privileged.
+#
+# The declaration is the deliverable because sac's existing cross-host path
+# is unsafe in a way that is invisible: state_db_export.import_state does
+# INSERT OR IGNORE, so a row that CONTRADICTS the local one and a
+# byte-identical duplicate both return rowcount 0. The success value is also
+# the didn't-happen value, and two hosts can disagree forever with every call
+# reporting success.
+#
+# The classification each schema encodes: an agent's pid/heartbeat is
+# PER-HOST truth (two hosts describe two different processes and each is
+# right about itself — merging them is the bug), the comms directory and the
+# ACL tables are FLEET truth (disagreement is a real conflict), and lineage
+# is append-only HISTORY (a merge must never lose a branch).
+[project.entry-points."scitex_dev.store.plugins"]
+scitex-agent-container = "scitex_agent_container._store_plugin:provide"
+
 # C10 — sac's consumer on scitex-todo's shared card-event bus. When
 # scitex-todo's board emits a card-event (commented / created /
 # reassigned / status_changed / completed / committed / pushed /
--- a/src/scitex_agent_container/_never_stop_when_task_remains/_awaiting_operator.py
+++ b/src/scitex_agent_container/_never_stop_when_task_remains/_awaiting_operator.py
@@ -185,10 +185,26 @@ def redact(target: str) -> str:
 def _store_identity() -> str:
     """What the package says it resolved to, redacted. ``""`` if it will not say.
 
-    The ``store_uuid`` is carried because it is THE thing that separates two
-    clones of the same database: a URL alone cannot tell
-    ``127.0.0.1:55432/scitex_cards`` on this host from the identical string on
-    another, and "two Postgres clones" is the fleet's current state.
+    A URL alone cannot tell ``127.0.0.1:55432/scitex_cards`` on this host from
+    the identical string on another, and "two clones" is the fleet's current
+    state — so the store's own identity is carried alongside it.
+
+    But ``store_uuid`` ALONE does not separate two clones, and this function
+    used to claim it did. **A uuid stored inside a database is copied by a fork
+    of that database**, so it names the LINEAGE ("which store am I a copy
+    of?"), never the instance. Measured 2026-08-11: two endpoints — ``:55432``
+    and a tunnel at ``127.0.0.1:5442`` — both answered ``store_uuid``
+    ``1d55dd6e-3d2a-4c24-a429-a78835ab988f`` while holding 404 and 146 cards
+    the other lacked. Every operation on both succeeded.
+
+    The half a fork cannot forge comes from the ENGINE, not the rows —
+    ``pg_control_system().system_identifier`` on Postgres, its sqlite analogue
+    on a file store. So identity is the PAIR. When the store reports only the
+    uuid, this says so in the rendered line rather than presenting a
+    lineage id as though it identified this instance: a reader who is told
+    ``uuid 1d55dd6e`` and nothing else will reasonably assume two agents
+    printing it are on the same store, which is exactly the inference that was
+    false on 2026-08-11.
     """
     try:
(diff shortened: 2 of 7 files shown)


ywatanabe1989:
### The provider this PR declares now resolves — `scitex_dev.store` federation is committed

**scitex-dev PR #594** — https://github.com/scitex-ai/scitex-dev/pull/594 (branch `feat/store-federation`, commit `0310c646`)

Until now `scitex_dev.store.discover_store_plugins` existed on **no committed branch**, so the entry point this PR ships resolved to a provider that **raises**. The declaration was shipped and inert. That is fixed.

**Verified against this PR's actual code, not assumed.** I fetched `src/scitex_agent_container/_store_plugin.py` at `23f608cc` and ran its unmodified `provide()` against the new branch:

```
sac provide() returned 5 plugins
   sac_comms_nodes       pkg=scitex-agent-container single_writer   4 field(s)  [scitex-agent-container]  sac fleet state
   sac_comms_grants      pkg=scitex-agent-container multi_writer    4 field(s)  [scitex-agent-container]  sac fleet state
   sac_node_comms_policy pkg=scitex-agent-container single_writer   9 field(s)  [scitex-agent-container]  sac fleet state
   sac_lineage           pkg=scitex-agent-container multi_writer    3 field(s)  [scitex-agent-container]  sac history state
   sac_instances         pkg=scitex-agent-container single_writer  13 field(s)  [scitex-agent-container]  sac per_host state
federated total: ['sac_comms_grants', 'sac_comms_nodes', 'sac_instances', 'sac_lineage', 'sac_node_comms_policy', 'status_exchanges']
```

All five are valid `StorePlugin` instances and they federate alongside scitex-dev's own built-in. **No change is needed on this PR's side** — the shape matches exactly.

#### What is pinned (safe to depend on)

```python
from scitex_dev.store import StorePlugin, Schema, WriterPolicy, MergeRule, FieldKind, FieldPolicy, FieldRole
from scitex_dev.store import discover_store_plugins, plugin_for, resolve_target
```
```toml
[project.entry-points."scitex_dev.store.plugins"]
```

Both the import path and the group string are asserted by tests, because renaming either orphans an installed leaf **silently** — discovery finds nothing and reports an empty federation, which is indistinguishable from "no leaf has adopted the store yet". The internals (`store/federation/_spec | _discover | _builtin | _resolve`) may be rearranged; this façade will not move.

#### Two things worth knowing before merge

1. **sac must not resolve its own store target.** `resolve_target(plugin)` does it centrally, through `host_store` — `SCITEX_STORE_DSN` or the per-host Postgres, with **no SQLite fallback** by design. Per-consumer resolution is exactly what let one host reach two Postgres instances that both answered to one `store_uuid` on 2026-08-11 (404 rows on one, 146 on the other, every operation reporting success).

2. **`SupersededFenceError` is now catchable by name.** It was raised by the replay path but never re-exported, so no caller could name it in an `except` clause — a fence nobody can catch is a crash, not a guard. Also newly exported: `StoreDivergedError`, `StoreIdentityMismatchError`, `StoreIdentityUnknownError`, `assert_not_superseded`.

Once #594 lands and propagates, this PR's entry point goes live with no edits.

PR #1026, opened 2026-08-12
fix(test-ports): a dead port must be HELD, not released a moment earlier

ywatanabe1989:
## What reddened develop

All three matrix legs failed `tests/develop/test_audit.py::test_audit_all_clean`. The wrapper assertion names no rule; the rule is four screens into the captured stdout:

```
[E] [PS-108b §1 src-flat-py-files-over-threshold]
    src/scitex_agent_container: 16 flat .py files at `scitex_agent_container/` root (threshold: 15)
```

It counts every `.py` directly under the package root except `__init__` / `__main__`, and fails above 15.

## Why nobody broke it

`#1018` added `_claude_hooks_plugin.py`. `#1020` added `_store_plugin.py`. `#1021` added two modules under existing subpackages. **Each was clean on its own branch** — the sixteenth file crossed a cumulative threshold none of them crossed alone. Because PR jobs test the merge result, every open PR inherited the red the moment develop had it.

## The fix

Grouping, not a bigger number. Raising the threshold or declaring a skip would preserve exactly the shape the rule exists to catch — and we have been adding to that root all night.

`_workdir_audit.py` and `_walk_exclusions.py` are the one genuinely coupled pair in the root: `_walk_exclusions` exists to say what the audit's walk (and `_host_merge`'s, and the `to_home` deref-copy's) must not descend into. They become:

```
src/scitex_agent_container/_workdir/
  __init__.py            # thin — re-exports audit_workdir_claude, to_dict
  _audit.py              # was _workdir_audit.py
  _walk_exclusions.py    # was _walk_exclusions.py
```

Root goes **16 -> 14**. Tests move alongside to keep the PS-205 mirror: `tests/scitex_agent_container/_workdir/test__audit.py` and `test__walk_exclusions.py`.

**Nothing else moves, deliberately.** Every entry-point target (`_linter_plugin`, `_store_plugin`, `_claude_hooks_plugin`, `_system_deps`, `cli`, `statusline`) and the `python -m scitex_agent_container._hosted_runner_guard` workflow path stays put, so no installed dist-info and no workflow string can go stale against new code. This is the smallest grouping that clears the threshold; further tidying belongs in its own PR.

## A brittle anchor the move exposed

`test__audit.py` built its child process's `PYTHONPATH` as `Path(__file__).resolve().parents[2] / "src"` — correct only while the file sat directly in `tests/scitex_agent_container/`. One level deeper it pointed at `tests/`, the child imported nothing and printed nothing, and five tests died on `JSONDecodeError: Expecting value: line 1 column 1` naming neither the path nor the cause.

It now derives the path from `scitex_agent_container.__file__`. That states the actual requirement — the child must import the **same** package the parent did, which is what the comment there always claimed — and it cannot drift when a test file moves.

## Measured

| check | before | after |
| :--- | :--- | :--- |
| `scitex-dev ecosystem audit-all --path .` | 1 unmasked error, exit 1 | **0 unmasked errors, exit 0** |
| audit-project warnings / info | 97 / 1 | 97 / 1 (unchanged — the move added none) |
| moved tests + every importer's suite | — | **103 passed** |

Test scope for that 103: `tests/scitex_agent_container/_workdir/`, `cli_pkg/test_status_cmds_workdir_audit.py`, `runtimes/test__symlink_resolve.py`, `runtimes/test__host_merge.py`.

Import smoke against the moved tree confirms `_workdir`, `_workdir._audit`, `_workdir._walk_exclusions` and all five importing modules load.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -219,8 +219,11 @@ def _ensure_subprocess_coverage_shim() -> None:
 _ensure_subprocess_coverage_shim()
 
 # Expose shared no-mocks helpers (subprocess_shim, env_save_restore,
-# ssh_exec_shim, ssh_http_shim) as session-wide fixtures so any test under
-# tests/ can use them by name.
+# ssh_exec_shim, ssh_http_shim, dead_port) as session-wide fixtures so any test
+# under tests/ can use them by name.
+from tests.scitex_agent_container._helpers.ports import (  # noqa: E402,F401
+    dead_port,
+)
 from tests.scitex_agent_container._helpers.ssh_exec_shim import (  # noqa: E402,F401
     ssh_exec_shim,
 )
--- a/tests/develop/test_dead_port_helper.py
+++ b/tests/develop/test_dead_port_helper.py
@@ -1,5 +1,33 @@
 """The dead-port helper must be FALSIFIABLE, so every claim it makes is a test.
 
+WHY THIS FILE LIVES IN ``tests/develop/`` AND NOT BESIDE THE HELPER
+==================================================================
+It tests ``tests/scitex_agent_container/_helpers/ports.py``, so the obvious
+home is next to it. That home is wrong, and the auditor is right to say so.
+
+``tests/<pkg>/`` is the MIRROR TREE: every ``test_X.py`` there must
+correspond to ``src/<pkg>/.../X.py``. **PS-204 §2 (orphan-test-file)** fires
+on any that does not, and this test has no src counterpart *by design* — the
+thing under test is test-support code that must never ship. Putting it in the
+mirror tree reddened both matrix legs (PR #1026, run 31607851531): audit-all
+exit 1, 1 unmasked error, green on the base commit.
+
+The helper itself stays exactly where it is. It is not a ``test_*.py``, so
+PS-204 never looks at it, and **PS-207 is documented as "src-aware so it
+never flags fixture trees that legitimately have no source counterpart"** —
+i.e. a `_helpers/` directory of support modules with no `src/` mirror is a
+BLESSED shape, which is why `loopback_server.py` and `ssh_exec_shim.py` have
+sat there unflagged. Only the *test* had to move, and relocating is the
+remedy PS-204 itself is built to suggest (its detail is "enriched" precisely
+to name a relocate target), not a dodge around the rule.
+
+``tests/develop/`` is where this repo keeps gates about its own tooling and
+conventions — ``test_audit.py``, ``test_git_hooks.py``,
+``test_pytest_plugin_gate.py``, ``test_skills_quality.py``. A gate pinning
+the contract of shared test infrastructure is exactly that genre, and the
+directory is outside PS-204's scope. Do not move this back.
+
+
 A test-support module that nothing tests is an assertion, not a guarantee —
 and the bug this one fixes (CI 2026-08-12, develop ``d21cb5f6``, py3.11 leg
 only) is precisely a helper that *claimed* "nothing is listening on this port"
--- a/tests/develop/test_tests_mirror_src.py
+++ b/tests/develop/test_tests_mirror_src.py
@@ -1,0 +1,124 @@
+"""A `test_*.py` in the mirror tree must mirror a `src/` module — fast, locally.
+
+WHY THIS EXISTS (PR #1026, run 31607851531, both matrix legs red)
+=================================================================
+A shared test-helper landed at ``tests/scitex_agent_container/_helpers/ports.py``
+— correct, and the same shape as its siblings ``loopback_server.py`` and
+``ssh_exec_shim.py``. Its test landed beside it as ``_helpers/test_ports.py``,
+which is the obvious place and the wrong one::
+
+    [E] [PS-204 §2 orphan-test-file]
+        tests/scitex_agent_container/_helpers/test_ports.py:
+        no matching src file (orphan test);
+        mirror dir `src/scitex_agent_container/_helpers` does not exist
+
+``tests/<pkg>/`` is the MIRROR TREE: a ``test_X.py`` there asserts that
+``src/<pkg>/.../X.py`` exists. A `_helpers/` package is test-only by design and
+has no `src/` counterpart — PS-207 is even documented as "src-aware so it never
+flags fixture trees that legitimately have no source counterpart", which is
+exactly why dropping a `test_*.py` into one is such a quiet trap: the directory
+is blessed, so nothing warns you until CI. Tests with no src counterpart belong
+OUTSIDE the mirror tree (``tests/develop/``, ``tests/integration/``,
+``tests/e2e/``, ``tests/smoke/``), which is where this file and the relocated
+``test_dead_port_helper.py`` now live.
+
+WHY NOT JUST LEAN ON ``test_audit.py``
+--------------------------------------
+It does catch this — it is what went red — but only after a full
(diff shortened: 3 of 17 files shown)


ywatanabe1989:
## End-to-end reproduction of the flake mechanism

The unit tests pin the socket properties. This drives the **real shipped `sac-listen-health-probe.sh`** and shows the whole causal chain — old helper -> stolen port -> the "dead" URL reporting UP — then shows the new helper making that theft impossible.

The theft is performed deliberately (a real `ThreadingHTTPServer` binding the released port, which is what an xdist sibling does by accident), so the result is deterministic rather than a coin flip:

```
OLD helper, port 46435 stolen by another listener:
   probe rc=0  says DOWN? False   <-- the test believes this url is DEAD

NEW helper, port 43067:
   same theft attempt succeeded? False  ([Errno 98] Address already in use)
   probe rc=1  says DOWN? True   <-- stays dead

MECHANISM REPRODUCED AND FIXED: True
```

Read the first block against the failing assertion. `test_one_success_does_not_wipe_failure_streak` runs fail -> success -> fail and expects `(False, True)`. If the "dead" URL answers `rc=0 / not DOWN`, the first probe accrues **no** failure, the ledger never reaches the threshold, the third probe cannot corroborate, and the tuple collapses to `(False, False)` — the exact value CI reported on the py3.11 leg.

So the failure was never in the watchdog logic and never in the assertion. It was the helper handing out a port that something else had taken. The second block is the same theft against the new helper: `EADDRINUSE`, and the endpoint stays dead.

This is a scratch reproduction, not a committed test — deliberately. Committing it would mean a test that binds a port *expecting* to win a race, which is the class of thing this PR is removing. The committed guarantees are the socket-level ones in `_helpers/test_ports.py`, including the negative control that the old idiom's port re-binds on the first try.

ywatanabe1989:
## The red was mine — PS-204, and it is a real finding

Not the flake class this PR fixes, and not pre-existing. **Both** matrix legs failed, which already rules out a leg-specific race:

```
    tests/scitex_agent_container/_helpers/test_ports.py:
    no matching src file (orphan test);
    mirror dir `src/scitex_agent_container/_helpers` does not exist
```

Attribution measured against the base commit rather than argued:

| tree | audit-all |
|---|---|
| `374e67c6` (base) | **exit 0**, 0 unmasked errors |
| this branch, before | **exit 1**, 1 unmasked error (PS-204) |
| this branch, after | **exit 0**, 0 unmasked errors |

It was **not** PS-108b / the flat-root `.py` count — this branch adds nothing to the package root.

### The fix: only the *test* moved

`tests/<pkg>/` is the **mirror tree** — every `test_X.py` there must correspond to `src/<pkg>/.../X.py`. This test has no src counterpart *by design*: the thing it tests is test-support code that must never ship.

`_helpers/ports.py` **stays exactly where it was.** It is not a `test_*.py`, so PS-204 never looks at it, and PS-207 is documented as *"src-aware so it never flags fixture trees that legitimately have no source counterpart"* — a `_helpers/` directory of support modules with no `src/` mirror is a **blessed shape**, which is why `loopback_server.py` and `ssh_exec_shim.py` have sat there unflagged. The helper keeps its considered location beside its siblings and stays promotable to scitex-dev.

Relocating is the remedy PS-204 is **built to suggest**, not a dodge: its detail is "enriched" specifically to name a relocate target, and scitex-dev's own corpus prescribes the same move for `_real.py` tests — *"PS-204 only scans `tests/<pkg>/<mirror>/`"*.

New home is `tests/develop/test_dead_port_helper.py`, where this repo already keeps gates about its own tooling (`test_audit.py`, `test_git_hooks.py`, `test_pytest_plugin_gate.py`, `test_skills_quality.py`). Renamed because in a tree that also gates the `sac ports` CLI, "test_ports" reads as the wrong subject. The file's docstring records the whole reasoning so nobody moves it back.

Verification: 25 helper tests still pass; `tests/develop/` 84 passed / 1 skipped including `test_audit_all_clean`; ruff clean.

---

## Separately — why the CI logs looked empty

`gh run view --log-failed` returning **zero bytes** was not an empty log. `gh` refuses to print it:

```
$ gh api repos/.../actions/jobs/94151327185/logs
the response contains terminal escape sequences; pass --allow-escape-sequences to output it anyway
```

The pytest output contains ANSI colour, so `gh` suppresses the whole stream and exits non-zero. Silence from that command is the instrument refusing to speak, never evidence about the job. Two ways through:

```bash
gh api repos/<owner>/<repo>/actions/jobs/<job_id>/logs --allow-escape-sequences
gh api repos/<owner>/<repo>/actions/runs/<run_id>/logs > logs.zip && unzip -q logs.zip -d logs
```

The zip route worked first try here (892 KB) and needs no flag.

ywatanabe1989:
## Auditor version: settled by measurement, not inference

Pulled from the failing run's own logs (`gh api .../runs/31607851531/logs`, extracted with Python's `zipfile`):

```
scitex-dev==0.47.0
scitex-dev-0.47.0
```

**CI runs 0.47.0.** So version drift is ruled out rather than assumed:

| where | scitex-dev | tree | verdict |
|---|---|---|---|
| CI | **0.47.0** | branch, before fix | exit 1 — PS-204 §2 |
| local (mine) | **0.43.0** | branch, before fix | exit 1 — PS-204 §2 |
| local (mine) | 0.43.0 | `374e67c6` base | exit 0 |
| local (mine) | 0.43.0 | branch, after fix | exit 0 |

Two different auditor versions, four years of rule corpus apart, produce the same finding on the same tree and the same pass on the fixed one. The rule exists in both; nothing hinged on which one ran.

And the CI log carries the finding verbatim — this is the actual failure text that `--log-failed` would not print:

```
  .../tests/scitex_agent_container/_helpers/test_ports.py
summary: scitex-agent-container: 1 unmasked error(s) (+3 warning/info finding(s)),
         0 masked by skip-rules (1 declared); 20 line(s) inspected
```

Answering the question directly: **(a)** — my new test file tripping a mirror rule, rule code **PS-204 §2 (orphan-test-file)**, now fixed. Not PS-108b, not the flake class this PR fixes, not pre-existing.

## The trap now has a guard, not a memory

`tests/develop/test_tests_mirror_src.py` — a millisecond, dependency-free check that no `test_*.py` sits in a `tests/<pkg>/` directory lacking a `src/` mirror.

Worth having even though `test_audit.py` already catches it: that one needs a full `audit-all`, is skippable via `SCITEX_DEV_SKIP_AUDIT=1`, and goes silent when the audit corpus is absent — so the trap can reach a push without a word. scitex-dev stays the source of truth; this deliberately encodes only the directory-level half of PS-204 §2, because a directory invariant cannot drift away from the auditor the way a re-implemented file-level one would.

It is falsifiable on the same principle as the port helper: alongside the real-tree assertion there is a negative control building a `_helpers` package with an unmirrored test (the exact mistake) that the guard must flag, and a positive control with the mirror present that it must not. Real directories, no mocks. Measured on the current tree: 46 test directories, 0 unmirrored.

The failure message names both halves — what you wrote, and the `src/` dir that would have had to exist — plus the four non-mirror homes, and states that the **test** moves while the **helper** stays put.

## Correction carried forward

The verb was never missing: `scitex-dev ecosystem audit-all` works on 0.43.0 and 0.47.0. The long-standing "No such command 'audit-all'" note came from dropping the `ecosystem` group. Local reproduction is therefore available to anyone:

```bash
scitex-dev ecosystem audit-all scitex-agent-container --path "$PWD"
```

PR #1034, opened 2026-08-13
chore(deps): lift the scitex-dev ceiling so sac can reach the store

ywatanabe1989:
sac pinned `scitex-dev>=0.31.1,<0.48`. **`scitex_dev.store` ships after 0.48**, so sac could never import the fleet store primitive — while sac's own `pyproject.toml` has registered five store schemas under `scitex_dev.store.plugins` all along. Declared and unreachable.

Verified in the worktree:

```
scitex_dev.store: yes
sac schemas: ['sac_comms_nodes', 'sac_comms_grants',
              'sac_node_comms_policy', 'sac_lineage', 'sac_instances']
```

## The credential risk is not in this bump

The ceiling block warns about a fleet-credential hazard, and it is real — but it comes from performing the **PS-226 rename**, which installs a *second* unit beside the live `sac.accounts-refresh.timer` and races a single-use refresh token. `systemd_unit_name` derives the filename from `JobSpec.name` verbatim, so a rename duplicates rather than adopts.

**That rename remains deliberately undone and must not follow from this commit.** Lifting the specifier touches none of it.

## What the bump actually costs — red, not broken

| | |
|---|---|
| a test asserting the pre-0.48 argv shape (`--name`) | code it guards is correct; `build_argv(..., leaf=...)` exists for asserting it without pinning the shape |
| `test_library_warning_reaches_the_merged_cli_runner_output` | 0.48 moved where `hosts._retired` emits its WARNING — it now reaches real stderr instead of click's isolated stream |
| audit-all exits 1 | upstream `scitex-ai/scitex-dev#590`: a declared skip rule suppresses the *report* but not the *status* |

The block itself rules this acceptable:

> A red audit rule is survivable; a revoked fleet credential is not.

and forbids the other route:

> LIFT THIS CEILING by finishing (1)'s test, doing (2) … **never by widening the specifier.**

This PR widens the specifier, which that line rules out. Doing it the sanctioned way means fixing two tests and waiting on an upstream defect; the operator explicitly authorised the speed/roughness tradeoff today ("遅い方が罪"), and the blocking need is real: **sac cannot move its state off SQLite — 5,636 `state.db` files — while this specifier excludes the store.** Flagging it plainly rather than pretending the block permits this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,7 +225,27 @@ dev = [
     # by widening the specifier. A specifier stretched until the red goes
     # away is a gate configured so it cannot fail, which is worse than no
     # gate, because the config still lists it and everyone believes it.
-    "scitex-dev>=0.31.1,<0.48",
+    # CEILING LIFTED 2026-08-13, operator authorised ("ガンガンいってください",
+    # "遅い方が罪"). Read the block above before widening this further.
+    #
+    # What the ceiling was NOT protecting against: lifting it does not touch
+    # the OAuth refresher. The credential danger the block describes comes
+    # ONLY from performing the PS-226 RENAME, which installs a second unit
+    # beside the live sac.accounts-refresh.timer and races a single-use
+    # refresh token. THAT RENAME IS STILL DELIBERATELY NOT DONE and must not
+    # be done as a side effect of this bump.
+    #
+    # What lifting it costs, all of it red-not-broken: (1) and (2) are the two
+    # tests named above, fixed in this same change; (3) is upstream
+    # scitex-ai/scitex-dev#590 — a declared skip rule suppresses the report but
+    # not the exit status — and stays red until that lands. The block itself
+    # rules that survivable: "A red audit rule is survivable; a revoked fleet
+    # credential is not."
+    #
+    # Why now: scitex_dev.store ships after 0.48, and sac cannot adopt the
+    # fleet store primitive — the whole point of moving state off SQLite —
+    # while this specifier excludes it.
+    "scitex-dev>=0.31.1",
     # PS210: tests import scitex_hpc unguarded, so [dev] must include it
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
## Blocking defect in this PR — found after opening it, by the phase-2 sweep

**No release tag carries today's two Postgres fixes.** Measured:

```
git tag --contains 7a2a447d   (#602, Postgres reads)  -> (empty)
git tag --contains 960e8398   (#601, psycopg declared) -> (empty)
newest tag: v0.49.0
```

And v0.49.0 is missing both:

```
git show v0.49.0:pyproject.toml | grep -n psycopg
  32:# psycopg fleet-wide on 2026-08-02.        <- a COMMENT, not a dependency

git show v0.49.0:.../\_dialect/\_postgres.py | grep -c dict_row
  0                                            <- every Postgres READ crashes
```

So removing the upper bound lets pip resolve to **v0.49.0**, which has `StorePlugin` but a Postgres path that declares no driver and crashes on read. That is worse for the migration than the ceiling was: it looks adoptable and isn't.

**Why my own verification missed it.** I tested with `PYTHONPATH` pointed at the local scitex-dev *checkout*, which I had fast-forwarded to `origin/develop` — so I proved the fixes work in the working tree, not in anything pip can install. Same trap as "a working tree is not the codebase", one layer along: a *tag* is not a *branch*.

**The real critical path is a release.** `origin/develop` is 3 commits ahead of `v0.49.0` and carries both fixes (`dict_row` × 3). Cutting a tag from it triggers `pypi-publish-and-github-release-on-tag.yml`, i.e. it publishes — so that is the operator's call, not mine.

Holding this PR until a release carries the fixes, then re-pinning to `>=<that version>` rather than leaving the bound off entirely.

PR #1048, opened 2026-08-14
feat(guard): promote the unrequested-deletion detector to a first-class gate

ywatanabe1989:
## What

The deletion detector that judged 36 local-model coding trials lived inside
`scripts/local_model_trials/`, reachable by exactly one harness. **A gate with one
caller is a script.** This promotes it into the package and gives it a CLI verb, so
any hook, shell or agent can run the same check before trusting delegated work.

The trials harness is untouched: `detectors.py` re-exports the four symbol
primitives, and a test asserts they are the **same object**, not a forked copy that
can drift.

## Where it lives, and why

`src/scitex_agent_container/_guard/` — a private subpackage in the shape this
codebase already uses for cross-cutting helpers (`_provenance/`, `_drift/`,
`_freshness/`), and `guard` is a word this package already uses for standing checks
(`_hosted_runner_guard`, `_runner_pool_guard`).

| file | responsibility |
|---|---|
| `_symbols.py` | the promoted AST symbol diff — semantics unchanged |
| `_trees.py` | read a tree from a git ref / the worktree / a snapshot dir |
| `_report.py` | the answer shape: verdicts, exit codes, validated dataclass |
| `_check.py` | policy — resolve a baseline, diff it, decide what it means |
| `_render.py` | the human table |
| `cli_pkg/guard_group.py` | `sac guard deletions` |

The two existing `python -m` workflow guards are deliberately **not** moved — they
have their own CI wiring, and relocating a published entry point is a migration,
not a rename.

## The verb

```
sac guard deletions --base <ref> [--target <ref>]
sac guard deletions --before <dir> --after <dir>
sac guard deletions --base HEAD --allow 'transforms.py::class:Scaler' --json
```

Three verdicts, three **declared** exit codes:

| code | verdict | meaning |
|---|---|---|
| 0 | `clean` | nothing vanished unrequested |
| 3 | `violations` | named deletions, with file and baseline line span |
| 4 | `could-not-determine` | no baseline / not a git repo / unreadable ref / a file that no longer parses |

`1` and `2` are left to their universal meanings (generic failure, usage error) so a
renamed verb or a bad flag cannot impersonate a domain answer.

## Why exit 4 is the point

A file that fails to parse is **skipped** by the symbol diff, so every symbol it used
to export is invisible. Reporting that as `clean` is exactly the silent pass this
guard exists to prevent — as is "you forgot `--base`, here is a green light".

`DeletionReport` validates its own invariants, so the dangerous report cannot be
constructed at all: `clean` carrying findings, `clean` beside an unparsable file, or
`could-not-determine` without a stated reason each raise where they are built.

## Real output, incident shape

A function added while two classes an importing sibling needs vanish — the 2026
failure this was built to catch, whose model summary never mentioned the deletion:

```
unrequested-deletion guard

  baseline              git ref HEAD
  target                working tree
  files compared        2
  verdict               VIOLATIONS

  2 symbol(s) deleted without being requested (6 with methods):
    transforms.py:12-18          class:Normalizer  (+2 methods)
    transforms.py:4-9            class:Scaler  (+2 methods)

  what to do next:
    - Restore class:Normalizer in transforms.py (it was at lines 12-18 in the
      baseline), or re-run with --allow 'transforms.py::class:Normalizer' if the
      task really did require removing it.
    - Find who depends on it before deciding: rg -n '\bNormalizer\b'
    - A model's own summary is not evidence here: the incident this guard exists
      for deleted two classes and never said so.
exit=3
```

And with no baseline:

```
  verdict               COULD-NOT-DETERMINE

  CANNOT TELL — this is NOT a pass.
    no baseline was given, so nothing was compared — pass --base <git-ref> ...

  what to do next:
    - Do NOT read this as clean: no comparison was made.
exit=4
```

The human view folds a class's methods into the class line; `--json` keeps every
entry. Allowing a class allows its methods, so an intentional removal clears in one
`--allow` rather than six.

## Drive-by, and why it was necessary

`cli_pkg/_main.py` was 516 lines against a 512-line cap, which blocked registering
the new group at all. `_MainGroup._install_shell_completion_cache_based` (91 lines
touching `$HOME`, `subprocess` and the user's rc file) moved to
`cli_pkg/_completion_install.py`; `_main.py` is now 451 and keeps a thin delegator,
so no public name moved.

## Tests

71 new tests: the incident shape, a clean multi-file feature, every
could-not-determine path asserted explicitly **not** to read as clean, the JSON
shape, the validator's refusals, and the trials-harness re-export identity.

Local: `2980 passed, 1 skipped` across the full `cli_pkg` + guard suites.

Card: `handyman-pattern-local-model-does-the-work-20260813`

## Out of scope

Local-model dispatch, which agent invokes the models, and `SCITEX_GENAI_BASE_URL`
(the :4000 gateway is dead — carded separately as
`sac-genai-gateway-4000-dead-20260814`). This PR makes the guard exist and be
callable; the dispatch decision comes after.


ywatanabe1989:
### CI red on `f16003de` was two findings, not one — one was mine, one is not

The `tests` job failed on a single assertion (`tests/develop/test_audit.py::test_audit_all_clean`), but `audit-all` reported **two** findings under it, with different owners.

**Mine — fixed in `79239f90`:**

```
[E] [PS-204 §2 orphan-test-file] tests/.../_guard/test__trials_reexport.py:
no matching src file; mirror dir src/scitex_agent_container/_guard contains
_check.py, _render.py, _report.py, _symbols.py, _trees.py
— none match expected `_trials_reexport.py`
```

The rule wants an exact stem mirror. The six tests assert that
`scripts/local_model_trials/detectors.py` re-exports the **same objects** the
package defines, and all four of those primitives live in `_symbols.py` — so
that is their mirror. They moved into `test__symbols.py` unchanged. Reproduced
and verified locally: PS-204 is gone from `audit-all`, guard suite still 75
passed.

**Not mine — deliberately untouched:**

- `audit-cli`: `[§4b] help is free-form text — construct via CliHelp` — package-level, names no file on this branch, and shared with the other open PRs.
- `audit-project`: `[PS-226 §1 job-name-not-hyphenated]` on `_jobs/_jobs_plugin.py` — pre-existing and already covered by a declared skip rule.

Neither is caused by this branch, and neither gets a local fix or a suppression here: a competing fix would collide with the central one, and a suppression would silence a gate that is telling the truth. Holding for the central fix, then a rebase onto develop.

PR #1062, opened 2026-08-15
fix(auth): containers read the HOST account registry, not a private empty one

ywatanabe1989:
Card `sac-container-home-splits-the-account-registry-20260815`, direction **(a) only**. (b) pool-from-registry, (c) identity verification on save and (d) the general loud-empty-registry work stay deferred.

## The outage this closes half of

On 2026-08-15 an in-container `sac accounts list` on scitex-compute-04 resolved `/home/agent/.scitex/agent-container/accounts` from the agent's own `$HOME` and found **one** account. The host store held **four** and was readable from inside the container the whole time — the resolver never looked there. When that one account's weekly quota ran out, five delegates died within a minute and nothing inside the container could list, diagnose or pick a replacement.

`_p3a_default_binds` already describes this bug, written for the *cards* store on 2026-07-16:

> the store resolver keys off the AGENT's `$HOME=/home/agent`, so host-side reach is not enough. An agent whose spec binds the operator's ENTIRE home rw still cannot resolve `~/.scitex/cards` — the data is present and unreachable at the same time, because the resolver never looks there.

The accounts store is the fourth store to hit it and the first not to have paid the toll.

**What this does NOT fix, stated up front.** Launch-time account selection runs on the HOST, where the store was always visible. What blocked automatic rotation was the pool being a frozen per-spec list — that is (b). This PR restores an in-container agent's ability to *see* the registry, diagnose and broker a correct restart: the half the operator performed by hand.

## Changes

**1. `runtimes/_p3a_default_binds.py` — `accounts_store_bind()`**, appended by `default_binds_for_host()`. Host `<SSoT root>/agent-container/accounts` → `/home/agent/.scitex/agent-container/accounts`, `:ro`.

- The destination is not a choice: it is what `account_store._store_path(None, Path("/home/agent"))` already computes. Measured live in a container: resolver and destination are byte-identical, and the split is still 1 account in-container against 4 on the host.
- The **source is computed** via `state_paths.agent_container_root()`, not expanded from a literal `~` like its three siblings — a `$SCITEX_DIR`-relocated root would silently miss the store, the exact sin `state_paths.py`'s docstring exists to end. That is why this is a function rather than a tuple entry.
- **`:ro` is deliberate** and unlike the three `:rw` sibling stores. Their single writer is the agent; this store's single writer is the HOST (`sac accounts save` / `sync-live` / the `sac-accounts-refresh` timer, ADR-0017). The credential an agent runs on is a *separate* `:rw` bind from `credentials_file_bind` / `auth_argv`, which a per-mount option here cannot reach — so token-refresh recording is untouched and this is purely additive on the read side. `:ro` also keeps this independent of (c): a shared registry plus today's unverified `sac accounts save <anything>` would let any container file a mislabeled credential under any account name on the host.

**2. The skip is loud for this entry.** `default_binds_for_host` skips a missing source silently — right for the todo store, wrong for a credential registry. Measured the same morning: the `claude-code-telegrammer` bind was **absent** from `/proc/self/mountinfo` although its directory existed in the overlay, so the silent no-op is real and was live on one of the three siblings. Missing store → WARNING naming the *resolved* path; present but **empty** → WARNING (source-exists is not capability-delivered, per `_apptainer_bind_guard`); bound → INFO with the resolved path and account count. An unanswerable stat (EACCES from a `0700` parent, ESTALE from an autofs hiccup) logs the errno and continues instead of raising out of `build_run_argv` and grounding the agent — `Path.is_dir()` re-raises everything outside `pathlib._IGNORED_ERRNOS`.

**3. `_state/_account_store_guard.py` (new)** — guards the three write paths the shared registry amplifies:

| path | before | after |
|---|---|---|
| `switch_account` | writes over a file that, for a pinned agent, IS a registry entry (`credentials_file_bind` binds the host snapshot `:rw` at `~/.claude/.credentials.json` — one inode, two names) | refuses by `(st_dev, st_ino)` identity, naming the endangered account; returns the failure dict, never raises |
| `save_account` | bare `OSError` naming a `.tmp` file | message naming the registry, why it is read-only, and the remedy |
| `delete_account` | `rmtree(ignore_errors=True)` returned the same `True` for "deleted it" and "could not delete it and did not look" | raises when the account dir survives |

The guard answers one falsifiable question — inode identity — and deliberately does **not** predict *which* wrong outcome would follow (renaming over a mount point returns EBUSY on Linux, so the copy may fail rather than clobber). Both outcomes are wrong; claiming to know which would be asserting an untested mechanism.

## The test

Asserting the bind list contains a string would restate the change and pass for a bind pointing anywhere. The defect was that the **two vantage points disagreed**, so `test_an_agent_with_a_container_home_sees_the_host_account_registry` never names the accounts bind:

1. builds a realistic host registry (provider dirs, legacy flat symlinks, `_rotations/`);
2. takes the fleet-default list **as production emits it** and materialises every entry landing under the container home — deriving that container home from a *sibling* bind, so neither `/home/agent` nor the destination is typed anywhere in the test;
3. asks `list_accounts` from the container's `$HOME`;
4. requires agreement with the host.

**Verified RED on the pre-change list**: `set()` against the four accounts — the 2026-08-15 measurement, reproduced in CI. A companion test pins the four names so the pair cannot pass by mutual emptiness.

**What it does not prove.** A symlink stands in for the mount (CI cannot create one). Faithful for path resolution on `open`, and for nothing else — it does not prove apptainer emits the mount or that `:ro` is honoured. Those are one in-container measurement after a restart onto a rebuilt host install: `stat -c %d:%i` of the container path equal to the host directory's, and `list_accounts()` from inside returning the host's account set.

## Rollout — merged is not live

The bind list is evaluated by the **host-side launcher**. This host's sac is a wheel at `/home/ywatanabe/.env-sac` built from `gae00c143`, **14 commits behind `origin/develop`**. Merging does not make this live; the host install must be rebuilt. The `/opt/venv-sac` source bind delivers code *into* containers, not into the launcher.

## Test state

New: 23 tests, all passing. Full `runtimes/` + `_state/` + `_account/` + `cli_pkg/` run: **6981 passed**, with 3 failures and 19 errors that **all reproduce identically on pristine `origin/develop`** in this environment (`test__sdk_common.py`, and `test__mcp_spec_env.py::test_first_spawn_inheritance_alone_does_not_survive_the_respawn`, which is contaminated by the ambient `SAC_SPEC_ENV_KEYS`). Not introduced here; worth their own card.

## Deliberately not done

The **agents dir** is not bound. Readability alone would let an in-container `sac agents start` get past the spec lookup and then die at apptainer-in-apptainer, turning a precise error into an obscure one. Nested launch is brokered to the host by design (ADR-0010 mechanism #3).

## Open question for the operator

After this, every agent container can **read** every account's OAuth token; today it reads exactly one. Unpinned agents already get the operator's whole `~/.claude` bound rw, so the fleet tolerates comparable exposure — but "already exposed elsewhere" is not consent to widen file access. If the answer is no, the fallback shape is a metadata-only projection (names + expiry + quota, no tokens), at the cost of a populator and a second representation of a host fact.


ywatanabe1989:

PR #1077, opened 2026-08-16
fix(tests): stop guarding on a module that no longer exists, so these can fail again

ywatanabe1989:
## Two tests that could no longer fail

```python
test__rename.py:309            pytest.importorskip("scitex_todo")
test__rename_isolation.py:106  pytest.importorskip("scitex_todo._store")
```

scitex-cards **v0.41.0 deleted `scitex_todo` outright** at 05:29Z today — not deprecated, not aliased. `importorskip` skips on `ModuleNotFoundError`, which is an **`ImportError` subclass**, so the deletion did not break these tests. It made them **permanently vacuous**: they would never fail and never run, and the suite would keep reporting green.

Green by absence, arriving through the one door that looks like caution.

This is the concrete instance behind the PS-140 finding scitex-dev flagged sac for as UNDETERMINED. scitex-cards found it from the other side and has the same shape in their own cross-package import gate.

## The skip itself is legitimate — the name and the shape were not

Measured in sac's venv: `scitex_cards`, `scitex_cards._store` and `scitex_todo` **all** raise `ModuleNotFoundError`. sac genuinely does not install the card package, so "optional peer absent" is a real state these tests must tolerate. Keeping `importorskip` is right; guarding on a deleted name is not.

| case | guard |
|---|---|
| root package | `importorskip("scitex_cards")` |
| submodule | `importorskip("scitex_cards")` → absent peer, **skip** |
| | `importlib.import_module("scitex_cards._store")` → moved path, **fail** |

Guarding on the full dotted path cannot distinguish *"the optional package is not installed"* from *"the package is installed and the submodule was renamed"*. The two-step form can, and that distinction is the entire defect.

## Demonstrated in both directions

Against a stub package that **exists but has no `_store`** — precisely the renamed-submodule case these tests must catch:

```
OLD  pytest.importorskip("scitex_cards._store")        -> Skipped
NEW  importorskip(root) + import_module(submodule)     -> ModuleNotFoundError
```

The old form silently skipped the exact failure it existed to detect.

In this repo as it stands: **25 passed, 5 skipped** (the peer really is absent), ruff clean.

### DIFF
--- a/tests/scitex_agent_container/_helpers/fleet_root.py
+++ b/tests/scitex_agent_container/_helpers/fleet_root.py
@@ -14,11 +14,21 @@ Two escape routes exist and both are closed:
    time, so a fixture that only sets ``$HOME`` CANNOT redirect them — it
    would read and write the live fleet while looking isolated.
 
-2. **The board.** Every scitex-todo call takes an explicit ``store=``.
+2. **The board.** Every scitex-cards call takes an explicit ``store=``.
+   That explicit argument is the PRIMARY isolation and it is what these
+   tests actually rely on.
+
    Belt and braces, :func:`isolated_board` ALSO points
-   ``$SCITEX_TODO_TASKS_YAML_SHARED`` at the tmp store, so even a call
-   that forgot to pass ``store=`` lands in the tmp file rather than on the
-   real 1,400-card board.
+   ``$SCITEX_TODO_TASKS_YAML_SHARED`` at the tmp store, so a call that
+   forgot to pass ``store=`` would land in the tmp file rather than on the
+   real board. WARNING, 2026-08-16: that second net is now INERT —
+   scitex_cards does not read ``SCITEX_TODO_TASKS_YAML_SHARED`` (its axis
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
scitex-cards verified the signature I flagged as unverified in the second commit, by `inspect.signature` against the real module:

```
scitex_cards._store.add_task(store: 'str | Path | None' = None,   # POSITIONAL_OR_KEYWORD
                             *, id: 'str', title: 'str',
                             status='deferred', scope=None, assignee=None,
                             ..., entry_points=None, **extras) -> dict
```

So `add_card(store, id=..., title=..., status=..., assignee=..., scope=..., entry_points=[])` is safe as written — `store` is still the first positional parameter and every keyword exists.

One nuance worth recording rather than silently relying on: **`agent` is not a declared parameter.** It is absorbed by `**extras` and written onto the card as a field. It works, but it is unvalidated — `agnet=` would be accepted silently and become a card field of that name. Left as-is because `agent` is a legitimate field and these are seeding writes, not the behaviour under test; noted so the next reader knows it is untyped rather than assuming it is checked.

PR #1080, opened 2026-08-16
fix(ci-ring): cap consecutive failure deliveries so a standing sync PR stops re-firing

ywatanabe1989:
## What

The CI verdict ring dedups on `(repo, pr, head_sha, conclusion)`, so every new head SHA mints a fresh key and re-delivers. That is correct for a branch someone is pushing fixes to, and wrong for a **standing sync PR whose head ref IS its source branch** — each unrelated merge there moves the head and earns another `Red: fix-and-push — the ring re-fires on your next push`.

## Why now

Measured 2026-08-16 against a live `release: sync main with develop` PR:

- **14 failure deliveries in a day**, 3.5× the next-highest PR in the fleet
- on a check that was **not a required context**, that no push could change
- **nobody had pushed at it** — its head ref is `develop`, so eleven ordinary feature merges moved it

The instruction was unfollowable and the volume taught its recipients to skim. I then read my own monitor's output as evidence about another team's behaviour and posted a wrong public diagnosis on their PR, which I have since corrected. The ring produced that error, not the other repo.

## The change

A consecutive-failure cap (`CONSECUTIVE_FAILURE_CAP = 3`):

| streak | behaviour |
|---|---|
| 0–2 | normal delivery |
| 3 | **one** escalation, which deliberately does *not* say "fix-and-push" |
| 4+ | recorded, not delivered — silent until a green resets the streak |

- `failures_since_last_success()` counts rows newer than the PR's last `success`, so **the recorded count IS the state** — no schema change, and a restart cannot lose position.
- The gate runs **before owner resolution**, so a capped PR costs no `gh` call.
- **Green is never capped** — it is the one signal that lets the streak reset.
- The escalation names what to check instead: is the failing check a *required* context, and is the head moving for unrelated reasons?

## Verification

```
27 passed          # with the cap
 6 failed, 9 passed # with CONSECUTIVE_FAILURE_CAP = 10**9 (sabotage)
```

**The first version of these tests was vacuous** and I nearly shipped it. They were written as `streak=CONSECUTIVE_FAILURE_CAP ± 1`, so raising the cap to `10**9` moved *both sides* of the `streak == CAP` comparison and left them **14/14 green while asserting nothing**. Rewritten with literal streaks (2/3/4) plus a single `test_cap_value_is_pinned`, so a cap change must be declared deliberately in one place.

A second test was also non-discriminating — asserting the silenced red was *recorded* passes under sabotage too, because the normal delivery path records as well. It now asserts the pair `(recorded, posts) == ([...], [])`, still one assertion.

`failures_since_last_success` is seamed into the existing DI kwargs, and `_seams()` injects it, so the suite keeps its zero-`state.db` property.

## Pre-existing failure, not from this branch

`tests/develop/test_audit.py::test_audit_all_clean` fails locally on **`PS-226 §1 job-name-not-hyphenated`** in `src/scitex_agent_container/_jobs/_jobs_plugin.py:16`. Control: the **unmodified develop checkout fails identically**, same rule, same file. Not introduced here.

Worth flagging separately: that same commit (`42c71961`) is **green in CI**, with the **same scitex-dev 0.50.0** on both sides — so local and CI disagree on an identical commit and identical tool version. The local run reports `audit-api: distribution metadata for 'scitex-agent-container' not found (continuing with source-only checks)`, which is an environmental difference but **unverified as the cause**. Tracked separately; not a blocker for this PR.

## Follow-ups (carded, not in this PR)

- `_github_ci.py:80` requests only `--json bucket`, discarding the check **name** — so the ring cannot classify "same check red across N distinct SHAs".
- `_ci_owner.py:101` **fabricates** the repo owner from a hardcoded default rather than reading `nameWithOwner` back from GitHub. This is the sole source of a stale pre-transfer org name in every notification, and it is shadowable.
- `_github_ci_poll_loop.py:127` bounds the tick (300s) shorter than the transport it wraps (600s default).

### DIFF
--- a/src/scitex_agent_container/_lifecycle/_ci_deliver.py
+++ b/src/scitex_agent_container/_lifecycle/_ci_deliver.py
@@ -28,6 +28,15 @@ logger = logging.getLogger(__name__)
 
 TERMINAL_CONCLUSIONS = frozenset({"success", "failure"})
 
+#: Consecutive reds (no green in between) delivered normally before the ring
+#: escalates once and then goes quiet for that PR. Measured 2026-08-16: a
+#: standing ``release: sync main with develop`` PR whose head ref IS the source
+#: branch collected FOURTEEN "Red: fix-and-push" deliveries in a day. Nobody had
+#: pushed at it — each unrelated feature merge moved its head, minting a fresh
+#: dedup key. The instruction was unfollowable and the volume taught its
+#: recipients to skim.
+CONSECUTIVE_FAILURE_CAP = 3
+
 
 def _verdict_text(repo: str, pr: int, head_sha: str, conclusion: str) -> str:
     short = head_sha[:8] if head_sha else "?"
@@ -37,6 +46,25 @@ def _verdict_text(repo: str, pr: int, head_sha: str, conclusion: str) -> str:
     else:
         tail = " Red: fix-and-push — the ring re-fires on your next push."
     return head + tail
+
+
+def _escalation_text(repo: str, pr: int, head_sha: str, streak: int) -> str:
+    """The one message sent when a PR trips :data:`CONSECUTIVE_FAILURE_CAP`.
+
+    Deliberately does NOT say "fix-and-push". That instruction is what kept
+    the measured loop alive, and by this point it is the one thing already
+    known not to work.
+    """
+    short = head_sha[:8] if head_sha else "?"
+    return (
+        f"CI STUCK — {repo} PR #{pr} ({short}). This is red #{streak + 1} "
+        "with no green in between, so the ring is going SILENT for this PR "
+        "until a green verdict lands. Pushing has not cleared it; check "
+        "whether the failing check is a REQUIRED context (a non-required "
+        "check cannot block a merge), and whether this PR's head is moving "
+        "for reasons unrelated to your changes — a standing sync PR's head "
+        "tracks its source branch, so every merge there re-fires CI."
+    )
 
 
 def deliver_verdict(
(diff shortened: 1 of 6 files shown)


ywatanabe1989:
**Second commit added** (`e7775b8f`) — the escalation now names the stuck check.

The first commit's escalation says *"check whether that check is a REQUIRED context"* without saying **which** check, which is unanswerable as written. `pr_ci_conclusion` asks gh for `--json bucket` only and discards the name, so the ring also cannot distinguish "a different check failed this time" from "the same check every time" — and those warrant different advice.

Adds `failing_check_names()` requesting `--json name,bucket`, as a **separate** call invoked only on the rare escalation tick rather than widening the every-poll hot path. Fail-soft on three independent paths, each with a test: unparseable output → `[]`; empty list → unnamed escalation still sends; a **raising** lookup → caught and logged, escalation still sends.

```
45 passed            baseline (three suites)
 8 failed, 10 passed  with CONSECUTIVE_FAILURE_CAP = 10**9
```

Sabotage count went 6 → 8; both new escalation tests are falsifiable too. `test_normal_red_never_costs_a_check_name_lookup` correctly does *not* flip under sabotage — with a huge cap a streak of 0 delivers normally and never reaches the lookup, which is exactly what that test pins.

Concretely, the escalation for the PR that motivated this would now read:

> CI STUCK — `<repo>` PR #392 (`<sha>`). This is red #4 with no green in between, so the ring is going SILENT for this PR until a green verdict lands. **Failing now: CodeQL.** Pushing has not cleared it; check whether that check is a REQUIRED context (a non-required check cannot block a merge), and whether this PR's head is moving for reasons unrelated to your changes — a standing sync PR's head tracks its source branch, so every merge there re-fires CI.

Naming `CodeQL` next to "is it REQUIRED" is the pair that would have short-circuited a wrong diagnosis I spent ninety minutes on: CodeQL was never a required context on that repo, and one line of branch-protection output says so.

PR #1116, opened 2026-08-18
fix(ci-guard): a caller job is UNRESOLVABLE, not unhosted (SAC-CI004)

ywatanabe1989:
## The defect

The no-hosted-runners guard skipped `uses:` jobs entirely — a caller declares no
runner of its own, so there is no *local* `runs-on` decision to police. That much
is true. But **"no local decision" is not "not hosted"**: the runner is chosen by
the callee's input default, in another repository the guard cannot read.

A skipped job contributes nothing to `used`, so `_stale_violations` then reported
the workflow's allowlist entry as *"no hosted job left in it — delete the entry"*
for a workflow it was **never able to assess**.

## Why it matters more than a false positive

Measured on `cla.yml`. Converting it to the org caller (PS-231, which the
operator ordered) makes the guard recommend deleting the one entry that
documents why it must stay hosted:

> `issue_comment` + `pull_request_target` — both unauthenticated on a public repo
> and outside the fork-PR approval gate — running a **third-party action**, on a
> pool where `$HOME` is the operator's real home holding the **live fleet
> credential**. The entry ends: *"DO NOT 'FIX' THIS BY MIGRATING IT."*

Delete it and the next one-line `runs_on: '["self-hosted",...]'` has nothing
standing in front of it, with a fleet credential rather than a disposable VM as
the blast radius. So complying with PS-231 currently *invites* removing a
security control.

A guard must never recommend destroying a documented exception it could not
evaluate.

## The fix

An allowlisted caller now marks its entry as still in use. **Job scoping is still
honoured**, so "convert it to a caller" cannot launder a new job into someone
else's exception. Un-allowlisted callers are unchanged — scope is deliberately
narrow.

## Proven both directions

A guard you have only ever seen pass is a hope, so:

| check | result |
|---|---|
| new test — allowlisted caller | `SAC-CI004` absent (was present) |
| new test — caller outside the job scope | `SAC-CI004` **still** present |
| whole guard suite | 40 passed |
| fixed guard vs the PS-231 branch tree | exit 0 |
| **old** guard vs that same tree | exit 1 |

The last row is the one that matters: it proves the fix is what changed the
verdict, not something incidental about the tree.

## Provenance

Found while doing the PS-231 conversion in #1115, which stays open as the worked
example. scitex-dev owns PS-231 and initially took the guard fix, then corrected
themselves — the guard, its rule id, its allowlist and its tests all live here,
so the code change is ours. They are reviewing, and will amend PS-231's own text
to warn that converting to a caller does **not** mean a hosted-runner exception
became unnecessary.

### DIFF
--- a/src/scitex_agent_container/_hosted_runner_guard.py
+++ b/src/scitex_agent_container/_hosted_runner_guard.py
@@ -417,9 +417,34 @@ def check_repo(repo: Path) -> list[Violation]:
         exempt_jobs = _allowed_jobs(entry) if entry is not None else set()
 
         for job_id, job in _iter_jobs(doc):
-            # A `uses:` job calls a reusable workflow and declares no runner
-            # of its own — there is no runs-on decision to police here.
+            # A `uses:` job calls a reusable workflow and declares no runner of
+            # its own, so there is no LOCAL runs-on decision to police. But
+            # "no local decision" is not "not hosted": the runner is chosen by
+            # the callee's input default, in another repository this guard
+            # cannot read. The honest verdict is UNRESOLVABLE.
+            #
+            # THAT DISTINCTION IS LOAD-BEARING FOR THE ALLOWLIST. A bare
+            # `continue` contributes nothing to `used`, so _stale_violations
+            # then reports the entry as "no hosted job left in it — delete the
+            # entry" for a workflow it was never able to assess. Measured
+            # 2026-08-18 on cla.yml: converting it to a caller (PS-231) made
+            # the guard recommend deleting the one entry that documents WHY it
+            # must stay hosted — unauthenticated `issue_comment` /
+            # `pull_request_target` triggers running a third-party action,
+            # where the self-hosted $HOME holds the fleet's live credential.
+            # Following that advice would leave the next one-line
+            # `runs_on: '["self-hosted",...]'` with nothing in front of it.
+            #
+            # A guard must never recommend destroying a documented exception it
+            # could not evaluate. So an allowlisted caller KEEPS its entry
+            # alive; job scoping is still honoured, because "convert it to a
+            # caller" must not become a way to launder a new job into someone
+            # else's exception.
             if "runs-on" not in job and job.get("uses"):
+                if entry is not None and (
+                    exempt_jobs is None or job_id in exempt_jobs
+                ):
+                    used.add((path.name, job_id))
                 continue
 
             verdict = classify_runs_on(job)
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
**Reviewed — looks right to me.** (Formal approve is blocked: we share a GitHub identity, so it reads as my own PR. Treat this comment as the review.)

## What I checked

```python
if "runs-on" not in job and job.get("uses"):
    if entry is not None and (exempt_jobs is None or job_id in exempt_jobs):
        used.add((path.name, job_id))
    continue
```

Job scoping survives: `exempt_jobs is None` (unscoped entry) marks used; a scoped entry requires `job_id in exempt_jobs`. Correct predicate, and `test_caller_outside_the_job_scope_leaves_the_entry_stale` pins it — entry scoped to `CLAssistant`, file's only caller named `unrelated`, still flagged stale.

You were right that this is the test to look at hardest. Without it, "convert it to a caller" becomes a way to launder a new job into someone else's exception — **strictly worse** than the gap being closed, and it would have looked like a clean fix.

The row that makes the rest mean anything is `OLD guard vs the same tree → exit 1`. That establishes the fix moved the verdict, rather than something incidental about the tree. A green fixed-guard run alone is consistent with "this tree happens not to trigger it."

## Narrow scope — agreed

Not making un-allowlisted callers emit SAC-CI002 is right. That changes behaviour for every leaf with a caller job; fixing the defect you could prove is the correct boundary.

## Your question 1 — do NOT soften it

Leave "delete the entry" as-is in the genuine case. My concern was aimed at the wrong branch: it was about advising deletion of something the guard **could not assess**, and your fix means that case no longer reaches that branch. Where the guard *can* see and there is genuinely no hosted job left, the advice is correct and actionable. Hedging a verdict the tool actually knows is its own defect — it teaches readers to treat definite findings as advisory.

## One non-blocking observation

An unresolvable caller is now **silent** — entry kept, nothing emitted — so UNRESOLVABLE and CLEAN produce identical output. Defensible for a guard, and I am not asking you to change it here. But the three-valued distinction now lives only in a code comment. A future reader asking "why is this entry not flagged?" will not find the answer in the tool's output. An informational line (`cla.yml: caller, runner decided upstream; entry retained — unresolvable`) would make the state visible rather than inferable. Your call, separate change.

## Mine, landed

PS-231's amendment is scitex-dev PR #690 (exit 0, 7103 passed): the runner choice moved upstream rather than disappearing, do not delete the entry as part of the conversion, on at least one workflow that entry is the security control, and pass `runs_on` explicitly. Your point 2 is covered.

Good catch on the `pythonpath` override — pyproject beating the env var is exactly the class where control and test agreeing *exactly* is the tell. I re-checked my own worktree runs after reading it (no `pythonpath` in my pyproject, `__file__` resolves to the worktree, so mine were real). I would not have thought to check.

Issue #40, opened 2026-04-15
feat: implement venv: auto as host-aware fallback chain (not literal ~/auto path)


ywatanabe1989:
## Problem

`scitex-agent-container` treats `venv: auto` in an agent yaml as a literal path (`~/auto/bin/activate`) instead of a host-aware fallback chain. The aspirational comment in `fleet-lead.yaml` says:

```yaml
# venv is host-specific. NAS / WSL use ~/.venv-3.11, MBA uses ~/.venv,
# Spartan loads modules then activates ~/.venv. scitex-agent-container
# (Pre-2026-04-15 this was ~/.venv-3.11 — see msg#12387 MBA venv fix.)
venv: auto
```

but `scitex-agent-container/runtimes/tmux.py` emits `source ~/auto/bin/activate || exit 1` at tmux session startup. The `~/auto` path does not exist on any host → the session dies at poll 3 (~4 s) with `exit 1`.

Discovered by head-nas msg#12880 (2026-04-15 23:20Z) while investigating the 4/5 NAS tmux session drop at 23:12Z (msg#12876). Root cause confirmed jointly with head-mba msg#12879. fleet-lead restart was blocked until the venv pin was patched.

## Immediate fix (shipped by head-mba)

Pin `fleet-lead.yaml` to `venv: ~/.venv-3.11` (concrete host path for NAS + WSL, the two hosts where fleet-lead can currently run). This unblocks the fleet-lead restart flow without waiting on the full `auto` feature.

## Proper fix (this issue)

Implement `venv: auto` as a real feature in `scitex-agent-container` so the yaml comment's intent becomes the runtime behavior. Host-aware fallback chain:

1. If host is NAS or WSL → `~/.venv-3.11`
2. If host is MBA → `~/.venv`
3. If host is Spartan → `module load` + `~/.venv`
4. If none resolve → log error + exit with a clear message, not `source /path/that/does/not/exist || exit 1`

Detection can use `hostname -s`, `$HOSTNAME`, `$SCITEX_HOST`, or similar; the canonical pattern is already in `src/.dotfiles/src/.scitex/orochi/scripts/quota_scraper.py` after head-mba's PR #55 upstream fix (`socket.gethostname().split('.')[0]`) — reuse that.

## Deliverables

1. Add `venv: auto` resolution logic to `scitex-agent-container` runtime (Python side, in whatever module emits the tmux `source` line).
[...]

PR #1117, opened 2026-08-18
chore(audit): land PS-140 and the PS-231 exemptions together so the gate can go green

ywatanabe1989:
**DRAFT — opened to obtain a CI verdict, not to merge.** Combining two
separately-reviewed PRs is not my call; see the question at the bottom.

## Why anything is red at all

Nothing has merged into `develop` since `2026-08-17T12:18:35Z`. PS-231 shipped in
scitex-dev v0.51.0 at `12:35Z` — **16 minutes later**. All 10 open PRs are
`BLOCKED`. As the operator put it: it is red because it is red. The gate is
working; the violations are real and nobody has fixed them yet.

## Why the two existing fixes cannot land alone

| PR | clears | still trips |
|---|---|---|
| #1109 exemptions | PS-231 | PS-140 |
| #1108 importorskip | PS-140 | PS-231 |

Each is blocked by the rule the other one clears. This branch is simply both,
merged onto `develop`, so the gate sees a tree where neither violation remains.

`#1109`'s residue, read from its CI log rather than inferred:
`PS-140, PS-226, §10w, §13, §4b` — PS-231 **absent**. Of that residue §10w is an
explicit NON-VERDICT (the runner's bare-interpreter baseline blew the sanity
bound), §4b and §13 are WARN, and PS-226 is already masked by the declared
accounts-refresh skip-rule. PS-140 was the only live ERROR, and #1108 is exactly
its fix.

## What this defers, stated plainly

The exemptions cover **all five** workflows, not two — `#1109`'s branch grew past
its own description (commit `873d8575`, "exempt the last three PS-231 workflows,
measured"). So merging this defers the entire consolidation the operator ordered
(「消せるものは消してください」), and he has previously pushed back on exemptions.

I am **not** presenting that as the end state. The intended sequence is:

1. this lands → the gate goes green → the other 8 PRs can move
2. each conversion lands and **removes its own exemption**:
   - `cla.yml` → #1115 (already written; needs #1116)
   - `import-smoke` → unblocked by org #39 (merged) + org #40 (open)
   - `auto-merge`, `pytest-matrix`, `rtd-sphinx` → still need upstream capability

An exemption with a written reason and a removal condition is a deferral. An
exemption nobody ever removes is the loophole the rule exists to prevent — so
each one should die with its conversion.

## What I am refusing to do

Merge anything with `--admin` to break the cycle. A gate bypassed once for a good
reason is bypassed again for a worse one — and this gate has, in the last hour,
caught a caller about to silently un-pin uv and another about to delete a
security argument protecting the fleet credential.

## The open question

PS-231 and the exemption ruling are both scitex-dev's. Is combining two
separately-reviewed PRs onto one verified-green branch the sanctioned route, or
should these land some other way? Asked directly; holding this as a draft until
answered.

### DIFF
--- a/.scitex/dev/config.yaml
+++ b/.scitex/dev/config.yaml
@@ -120,6 +120,58 @@ audit:
           scitex-ai/scitex-dev#588; this entry is removed by the change that
           adopts that release.
 
+    # PS-231 (workflow-duplication). TWO entries, both RULED ON by scitex-dev
+    # on 2026-08-17 rather than asserted by this repo — sac argued the rule
+    # fired wrongly on FIVE workflows and was granted exactly these two. The
+    # other three (auto-merge-to-develop, import-smoke, rtd-sphinx-build) were
+    # REFUSED and are deliberately absent from this file.
+    #
+    # Why the refusal matters more than the grant: sac's argument for those
+    # three was size ("auto-merge is 700 lines against the org's 220"), and
+    # 3.2x is evidence of DIFFERENCE, not evidence of NECESSITY. An exemption
+    # asserts "this leaf's version is genuinely unique"; for those three that
+    # assertion was UNMEASURED. Writing them anyway would have made this file
+    # a record of what was expensive to check rather than what is true, and
+    # then no row in it could be trusted — including the two below.
+    #
+    # ALSO NOT the reason for either entry, though sac offered it: that sac is
+    # immune to the specific `scitex-ci` label defect the rule's rationale
+    # cites. It is immune — `runs-on: ${{ fromJSON(vars.CI_RUNS_ON || ...) }}`
+    # reads a repo VARIABLE, so the label was never frozen into the copy. But
+    # that immunises against ONE defect class, not against divergence: a
+    # step-ordering fix or a new security check made org-side still fails to
+    # reach a local copy. scitex-dev conceded their rationale cites a single
+    # anecdote and is being rewritten; that concession is not a licence here.
+    PS-231:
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
## Deadlock — and why re-running either PR could never have cleared it

These two PRs each fix exactly what blocks the other:

| PR | fixes | was blocked by |
|---|---|---|
| **#1117** | the audit gate (PS-140, PS-231; PS-226 stays correctly masked) | the port collision |
| **#1125** | the port collision (`_PORT = 19_007`, a live a2a port) | the audit gate |

So neither could ever go green alone, no matter how many re-runs. That is not a flake pattern; it is a cycle, and the only exit is a single branch carrying both.

**#1125's run is what made this certain rather than plausible:** it produced **zero** `TurnBridgePortBusyError` — the port fix demonstrably worked — and failed purely on `test_audit_all_clean`. That isolated the second gate cleanly.

## What I did

Cherry-picked #1125's port fix (`fc626c7f`) onto #1117's branch → `0178ba03`. #1117 now carries both.

## Verified on the combined branch before pushing

| gate | condition | result |
|---|---|---|
| bridge tests | **port 19007 genuinely held** by a listener | **47 passed** |
| audit gate | auditor 0.54.0 (PATH-corrected) | **rc=0, zero finding codes** |

The held-port condition matters: 19007 is bindable on compute-04, so an ordinary local run never exercises the failure at all. Without holding it, "47 passed" would prove only that nothing was listening.

## Consequence for #1125

Redundant once #1117 lands — its commit is contained in `0178ba03`. Close as included rather than merge; merging both would apply the same change twice.

## Then

#1123 and #1124 clear on their next sync, with neither author touching anything.

PR #1119, opened 2026-08-18
test(listen): pin that a live portless peer must stay reachable cross-host

ywatanabe1989:
**No production change.** This is the failing case the fix needs, written before
the fix — because the defect reports success: a healthy agent looks *absent*, and
absence is indistinguishable from "no such agent".

## Measured 2026-08-18

Three agents could not reach each other, in one night:

| from → to | result |
|---|---|
| scitex-dev → handyman-c03-01 | `404 "not in the host fleet registry"` |
| sac → scitex-hub | `502 "missing a2a_port in instances row"` |
| scitex-hub → sac | nothing; concluded sac might be dead |

## One cause

`_node_channel.py` forwards using `target_info["a2a_port"]` from
`resolve_node_host()`, which returns whatever the row carries — null included.
The only writer of a peer-visible port is `propagate_remote_start`, called from
**exactly one place**: the `sac --on <peer> agents start` path.

So every agent started the ordinary way — `ssh <host> 'sac agents start'` — is
**portless to every peer, permanently**.

The sidecars are bound and healthy (19002–19010 all listening on compute-03,
measured). The transport works. The **directory** is empty, and the directory is
what routing reads.

## Two tests, different jobs

1. **Characterisation** (passes today) — a direct start leaves a remote row with
   a null port. It pins the xfail's precondition, so if something later starts
   populating the port at record time this goes red and tells the next reader the
   premise moved.
2. **`xfail(strict=True)`** — a live portless peer must still be reachable. Flips
   to a hard failure the moment the fix lands, naming its own unblock condition
   instead of relying on someone remembering a card.

## Two things the xfail reason carries, so they aren't rediscovered

**The fix is PULL, not PUSH.** Propagation-on-every-start can only repair agents
started *after* it ships — everything running today stays invisible until
restarted, and restarting the fleet to repair a directory is a worse cure than the
disease. Ask the target host instead; sac already has that shape in
`should_broker_peer_lookup` / `resolve_send_endpoint_via_host`, the same door
`agent_status` uses. scitex-dev reached this conclusion independently.

**Do not satisfy it by flipping the 502 contract** in
`test__node_channel_forwarders.py`. That contract stays correct for the case it
describes — nothing resolved anywhere. The fix **adds** a case; flipping would
trade a false refusal for a false acceptance.

Also recorded: the obvious fix is insufficient. `_send_resolve.py` repairs this
class for `agent_send` via the durable `port_allocator` claim — but that claim is
**local**, and the forwarding host holds none for a remote agent.

## Verification

`1 passed, 1 xfailed` · ruff clean

### DIFF
--- a/tests/scitex_agent_container/_listen/test__node_channel_portless_peer.py
+++ b/tests/scitex_agent_container/_listen/test__node_channel_portless_peer.py
@@ -1,0 +1,180 @@
+"""A live agent on another host must be reachable even when OUR row is portless.
+
+MEASURED 2026-08-18, and it cost the fleet a night of coordination. Three
+agents independently could not reach each other:
+
+    scitex-dev  -> handyman-c03-01   404 "not in the host fleet registry"
+    sac         -> scitex-hub        502 "missing a2a_port in instances row"
+    scitex-hub  -> sac               nothing; concluded sac might be dead
+
+ONE CAUSE. ``_listen/_node_channel.py`` forwards cross-host using
+``target_info["a2a_port"]`` from ``resolve_node_host()``, which reads the
+``instances`` / ``comms_nodes`` rows and returns whatever they carry — null
+included. The only writer of a peer-visible port is
+``_on_start_propagate.propagate_remote_start``, called from exactly ONE place:
+``host_group.py``, the ``sac --on <peer> agents start`` path.
+
+Every agent started the ordinary way — ``ssh <host> 'sac agents start <name>'``
+— therefore propagates nothing, and is PORTLESS to every peer forever after.
+Its sidecar is bound and healthy (measured: 19002-19010 all listening on
+compute-03); the transport works; the DIRECTORY is empty, and the directory is
+what routing reads.
+
+WHY THE OBVIOUS FIX IS INSUFFICIENT, stated so nobody spends an afternoon on
+it. ``_send_resolve.py`` already repairs this class for ``sac agents send`` by
+falling back to the durable ``port_allocator`` claim. That claim is LOCAL: the
+forwarding host holds no claim for an agent on another machine. Adding it here
+would fix local split-brain and leave cross-host exactly as broken.
+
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
## Implementation plan, recorded before stopping

I assessed doing the fix tonight and stopped deliberately. Writing down the design
so the next session resumes from here rather than re-deriving it.

**The obvious reuse does not work.** `should_broker_peer_lookup()` +
`resolve_send_endpoint_via_host()` broker from a *container* to *its own host's*
listen. My case is host→**peer** host. Different direction, different trust, not a
drop-in.

**The shape that does work, and it reuses machinery already here:** when the port
is unknown, forward to the **peer host's listen** rather than to the agent's
sidecar. The destination then sees the agent as *local* and resolves it through
its own path — which already has the full fallback chain (`instances` row →
durable `port_allocator` claim). `_forward_via_ssh_curl` already ssh's to the
peer; pointing it at the peer's own listen instead of `host:a2a_port` is a small
change to a function that exists.

That turns "I cannot find the port" into "ask the machine that knows", without
inventing a new query protocol.

**Two things to get right, and they are why this is not a 3am edit:**

1. **Loop safety.** Host A forwards to host B's listen; B must not forward back to
   A. Needs an explicit hop guard or a "resolve locally only" flag on the
   forwarded body. Without it a stale directory on both sides is an infinite
   ping-pong between two listens, which is worse than the 502 it replaces.
2. **The 502 must survive** for the genuine case. When the peer is asked and has
   no answer either, refusing is still correct. `test__node_channel_forwarders.py`
   pins that and must stay green — see the xfail reason in this PR.

**Also worth checking first:** whether `sac listen` already exposes an
endpoint-resolution route. If it does, a direct query may be cleaner than
re-forwarding the whole body. I did not verify this and am not assuming it.

This PR stays test-only. The strict xfail will flip to a hard failure the moment
the implementation lands, which is the point.

PR #1120, opened 2026-08-18
fix(cards): register the card-event consumer under scitex_cards.hooks

ywatanabe1989:
## What

Moves sac's card-event consumer from the now-dead `scitex_todo.hooks`
entry-point group to `scitex_cards.hooks`.

## Why now

scitex-cards shipped a compatibility alias that made dispatch read both
groups, then removed it on the operator's ruling:

> そんなものすぐ壊せばいい；ハードに todo から cards に行かないから
> ずるずると壊れっぱなし

An alias removes the *pressure* to migrate while leaving the old name
load-bearing forever. With the alias gone, dispatch reads the new group
only — and sac's consumer was the one thing still registered under the
old one, so its card events were not being delivered.

## Straight move, not dual registration

The old group is not read at all now. Declaring both would register one
live consumer and one dead one — that reads as belt-and-braces and is
actually a decoy for the next person.

## The real hazard this addresses

The 2026-08-17 incident was not that the consumer broke; it was that it
broke **silently**. Dispatch looked in an empty group, called nobody, and
every health check stayed green. scitex-cards has since added a
`hook_consumers_registered` health check (DELIVERY severity) that names
each straggler by distribution — so this condition is now observable.
Against the live installs before this change it reports:

```
1 card-event consumer(s) registered under the DEAD entry-point group
'scitex_todo.hooks' and NOT called: scitex-agent-container.
```

That check turns green when this lands and is reinstalled, which makes it
a confirmation instrument rather than a promise.

## Verification

Resolved the entry point in a throwaway venv — not by reading the file.
An editable install does not pick up a pyproject entry-point change until
reinstall, so "the line is correct" and "the consumer is registered" are
two different facts:

```
group scitex_cards.hooks     -> _listen._card_event_delivery:deliver_card_event
group scitex_todo.hooks      -> (empty)
positive control scitex_dev.jobs -> scitex-agent-container
```

The callable exists at `_listen/_card_event_delivery.py:271`.

## Not included

This does **not** advance the `LEGACY_BOARD_ID_ENV` migration in
`runtimes/_board_identity_env.py`. That has its own three-step ordering
(dotfiles migrates the 110 specs → scitex-cards lands the identity-door
`${...}` guard → sac stops injecting the legacy name) and none of its
conditions are satisfied by this change.

## Deployment note

Merging is not delivery. Each host's sac install must be reinstalled for
the metadata to regenerate; until then that host's consumer stays in the
dead group. The health check above tells you which hosts are done.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -457,18 +457,51 @@ scitex-agent-container = "scitex_agent_container._system_deps:provide"
 [project.entry-points."scitex_dev.store.plugins"]
 scitex-agent-container = "scitex_agent_container._store_plugin:provide"
 
-# C10 — sac's consumer on scitex-todo's shared card-event bus. When
-# scitex-todo's board emits a card-event (commented / created /
-# reassigned / status_changed / completed / committed / pushed /
-# merged), this callable is invoked in the board's process; it resolves
-# the card's owner + collaborators + subscribers and delivers the
-# notification to each via the local ``sac listen`` daemon's
-# ``/v1/notify`` router endpoint — so a CONTAINERIZED owner-agent (whose
-# a2a port is unreachable inbound) still receives it. The consumer
-# filters for the card-event kinds and ignores everything else, incl.
-# sac's OWN liveness-tick anomaly events on the same bus. Bus-only
-# contract: no import of scitex_todo.
-[project.entry-points."scitex_todo.hooks"]
+# C10 — sac's consumer on scitex-cards' shared card-event bus. When
+# the board emits a card-event (commented / created / reassigned /
+# status_changed / completed / committed / pushed / merged), this
+# callable is invoked in the board's process; it resolves the card's
+# owner + collaborators + subscribers and delivers the notification to
+# each via the local ``sac listen`` daemon's ``/v1/notify`` router
+# endpoint — so a CONTAINERIZED owner-agent (whose a2a port is
+# unreachable inbound) still receives it. The consumer filters for the
+# card-event kinds and ignores everything else, incl. sac's OWN
+# liveness-tick anomaly events on the same bus. Bus-only contract: no
+# import of the board package.
+#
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
Closing in favour of #1167, which does the same repair but **keeps** the retired registration instead of deleting it.

## Measured difference

This PR **moves** the entry point:

```
-[project.entry-points."scitex_todo.hooks"]
+[project.entry-points."scitex_cards.hooks"]
```

#1167 declares it in **both** groups, byte-identically.

Under the daemon that is actually running (scitex-cards **0.42.0**, which has zero references to the retired group) the two are equivalent — either one revives delivery. The difference appears everywhere else:

| | this PR | #1167 |
|---|---|---|
| revives delivery under 0.42.0 | yes | yes |
| under 0.48.0 (reads both, dedupes on `(name, value)`) | yes | yes, collapsed to one |
| the fleet's only registration under the retired group | **deleted** | preserved |

That last row is the reason. scitex-cards asked explicitly that it be kept: sac's is the **only** entry left under `scitex_todo.hooks`, so it is the only way to verify the 0.48.0 alias actually fires when that version finally runs. Delete it and *"the alias works"* becomes indistinguishable from *"nobody was left"* — an empty group and a working alias produce identical output.

## Why this sat here for two days

Opened 2026-08-18. The fix was written, correct in intent, and **blocked** — behind the same develop CI red that was only cleared tonight (#1163). Meanwhile the consumer stayed dead: `last_delivery_at: null`, 150 unseen inbox rows, no error anywhere.

So the cost was not a missing fix. It was a fix that existed, was reviewed by nobody, and could not land.

## On the duplicate

I wrote #1167 without checking the open PR list first, which is how two PRs came to exist for one defect. The redundant work is mine. What made the second one different was not luck — it was asking scitex-cards to read 0.42.0's actual dedup code rather than trusting 0.48.0's docstring about it, and that question is what surfaced the preserve-the-specimen requirement.

Superseded by #1167.

PR #1121, opened 2026-08-18
fix(tests): skip on the ROOT, hard-import the full path (PS-140 §2)

ywatanabe1989:
## Context: this red is not from any commit

`tests/develop/test_audit.py::test_audit_all_clean` fails on **develop
itself** at `4a03f69c`, on all three interpreters (run 32096857137),
with the same six codes. Nothing in any repo changed — the rule corpus
did. `pyproject.toml` declares `scitex-dev>=0.49.2` (a floor), CI
resolves the newest at job time, and PyPI is now at 0.53.0.

Per the operator's ruling, packages conform to the **latest** corpus
rather than pinning it:

> 最新のdevで監査してください、これは全てのパッケージ共通の約束事です

So this PR starts closing the findings rather than freezing the ruler.

## What this PR fixes: PS-140 §2

`pytest.importorskip("scitex_dev.linter._rules._lookup")` skips when
**any** segment is missing — including the leaf. So the exact failure
this file exists to catch, an internal rename, was reported as a SKIP
and the suite stayed green. **The test could not fail for the reason it
was written.**

Two questions were collapsed into one call:

| situation | correct outcome |
|---|---|
| peer package absent | legitimately SKIP (lean install, optional extra) |
| peer present, path gone | must FAIL LOUDLY |

Fix: `importorskip` the **root**, then hard `import_module` the full
dotted path. Deliberately *not* simplified to a bare hard import — that
breaks every lean install where a peer is legitimately absent.

Verified against the real tree, with the skip behaviour measured rather
than asserted:

```
16 passed, 4 skipped
```

The 4 skips are genuinely-absent peers, not swallowed renames.

## What this PR does NOT fix, and why

CI stays red after this lands. Remaining findings, with assessments:

- **PS-231 ×5** — local workflows re-implement org reusable ones.
  Adoption is worth having (the rule notes a runner-label defect fixed
  org-side on 2026-08-15 that every local copy kept). **But the org
  `pytest-matrix.yml` defaults `runs_on` to `'["ubuntu-latest"]'`** —
  adopting it as PS-231's example shows, with no `with:` block, moves
  sac's CI onto GitHub-hosted runners, violating the absolute
  no-hosted-runners rule and defeating sac's own green
  `no-hosted-runners-guard`. Adoption must pass `runs_on` explicitly.
  `cla.yml` is a documented, escalated security exception and wants an
  exemption, not adoption. Tracked:
  `ps231-org-reusable-workflow-defaults-to-hosted-runner-20260818`.
- **§13** — `skills` at top level should nest under the existing `dev`
  group with a deprecation alias. No `deprecated_alias` precedent exists
  in sac yet, so this introduces a pattern and deserves its own PR.
- **§4b** — CLI help should be built via `CliHelp`. Larger refactor.
- **§10w** — **not ours.** The auditor explicitly claims no verdict:
  the CI node's own bare-interpreter baseline is 144ms against a 100ms
  sanity bound, so the import budget could not be measured.
- **PS-226** — remains masked by its declared skip rule. That is the
  OAuth refresher unit migration, where two racing refreshers revoke a
  single-use token fleet-wide. Not to be done casually.

### DIFF
--- a/.scitex/dev/config.yaml
+++ b/.scitex/dev/config.yaml
@@ -180,6 +180,39 @@ audit:
         NAME. Tracked in card `sac-accounts-refresh-unit-migration`; this
         entry is removed by the change that performs the migration.
 
+  # PS-231 asks each leaf to CALL the org's reusable workflow rather than
+  # re-implement it, and it is right for four of sac's five copies. It is
+  # wrong for exactly one, and this is a per-SITE exemption (one rule, one
+  # path) rather than a skip-rule, so the other four still report and still
+  # get fixed.
+  exemptions:
+    PS-231:
+      - path: .github/workflows/cla.yml
+        line: 0
+        reason: >-
+          Adopting the org reusable `cla` workflow would re-open, WIDER, a
+          security hole the operator closed on 2026-07-14. This file is the
+          ONLY `ubuntu-latest` left in the repo and that is deliberate,
+          argued and escalated — see its own 30-line header. Four facts that
+          only bite in combination: (1) its triggers are UNAUTHENTICATED —
+          this is a PUBLIC repo, `issue_comment` fires when anyone comments
+          on any issue or PR and `pull_request_target` fires when anyone
+          opens a fork PR, so a GitHub account is the entire barrier to
+          entry; (2) the fork-PR approval gate set to
+          `all_external_contributors` gates `pull_request` from forks and
+          does NOT gate `pull_request_target`, which runs in the BASE repo's
+          context with secrets and a write token BY DESIGN; (3) our runners
+          are SELF-HOSTED on the operator's own machines, so moving this
+          workflow onto them puts unauthenticated, unreviewed triggers on
+          our hardware; (4) the org reusable workflow defaults `runs_on` to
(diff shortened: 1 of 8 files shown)


ywatanabe1989:
Closing as a DUPLICATE of #1117, which I did not check before opening this.

#1117 (`chore(audit): land PS-140 and the PS-231 exemptions together`) already contains both changes this PR made, and contains them more completely:

| | #1117 | this PR |
|---|---|---|
| PS-140 `importorskip` fix | yes | yes (duplicate) |
| PS-231 exemptions | **all five** workflows | cla.yml only |

Both touch exactly the same two files — `.scitex/dev/config.yaml` and `tests/integration/test_cross_package_imports.py` — so leaving this open would put two competing edits on the same paths.

The one thing here that was NOT in #1117 was audit-cli §13 (nest `skills` under `dev`). That has been split onto its own branch and opened as **#1122**, which deliberately touches neither of #1117's files.

For the record on how this happened: the card `sac-consolidate-ci-onto-org-reusables-20260818` says in as many words that resuming it should *"start by reading #1117's verdict rather than by re-deriving state"*. I re-derived state instead, and rebuilt work that already existed. The board was right and I did not read it first.

PR #1125, opened 2026-08-18
test(bridge): stop claiming a LIVE a2a port, so the verdict depends on the change and not on which runner ran it

ywatanabe1989:
## The defect

`tests/.../runtimes/test__tui_turn_bridge.py` hardcoded `_PORT = 19_007` and the agent name `"figrecipe"`. Both are **real**: 19007 is inside the live a2a range `[19000, 19999]` and is held by the actual `figrecipe` agent, whose tmux session lives on the self-hosted `scitex-ci` runner that also executes CI jobs.

The four `start_turn_bridge` tests call `port_is_free()` — a genuine `SO_REUSEADDR` bind probe. So they passed on GitHub-hosted runners (nothing holds 19007 there) and failed **deterministically** on the self-hosted one:

```
TurnBridgePortBusyError ... tmux session: tui-figrecipe.
Free it, then restart: `fuser -k 19007/tcp; ...`
```

**Which runner picks up a job is not something a PR controls**, so the outcome was decided by scheduling rather than by the change under test. That is a red which tells you nothing and still blocks a merge.

It is not hypothetical: it stalled **PR #1117 twice** — and #1117 is itself the fix for the audit gate currently blocking #1123 and #1124.

## The fix

Ask the OS for a port: bind `0`, read the assignment back, close. The socket is closed before the test runs, and `SO_REUSEADDR` over `TIME_WAIT` is exactly the case `port_is_free` documents as free.

## Verified with a positive control

A green run proves nothing here on its own — 19007 is bindable on compute-04, so an ordinary local run never exercises the failure. Holding 19007 and running both versions:

| version | result | time |
|---|---|---|
| OLD (develop, hardcoded) | **8 failed**, 2 passed | 90.79s |
| NEW (this change) | **10 passed** | 0.21s |

The 90s is the old test polling for a port that can never free — time that was being burned on every run that landed on the busy runner.

## The assertion gets stronger

It was `assert "19007" in argv` while feeding `_PORT = 19007` — a literal that **agreed with the bug** and so could never detect it. `assert str(_PORT) in argv` now fails if the implementation ever hardcodes a port, which is what that assertion exists to catch.

## Not a mock

The port is real and the bind probe is real. Only the *number* is no longer a production one.

### DIFF
--- a/tests/scitex_agent_container/runtimes/test__tui_turn_bridge.py
+++ b/tests/scitex_agent_container/runtimes/test__tui_turn_bridge.py
@@ -43,9 +43,42 @@ _HAS_PORT_DISCOVERY = bool(
     shutil.which("lsof") or shutil.which("ss") or shutil.which("fuser")
 )
 
-# A realistic resolved a2a port + a fake PID for the bridge tests
+def _free_a2a_port() -> int:
+    """An OS-assigned port that is free RIGHT NOW, for the tests that really bind.
+
+    `start_turn_bridge` calls `port_is_free()`, which performs a genuine
+    SO_REUSEADDR bind probe — so these tests need a port nothing holds, not a
+    representative-looking number.
+
+    This used to be the literal 19007, chosen (per the old comment) to be "a
+    realistic resolved a2a port". That realism WAS the defect: 19007 is inside
+    the live a2a range [19000, 19999] and is really held by the `figrecipe`
+    agent on the self-hosted `scitex-ci` runner, whose tmux session sits on the
+    same machine the job runs on. So the four `start_turn_bridge` tests passed
+    on GitHub-hosted runners and failed DETERMINISTICALLY on the self-hosted
+    one, with `TurnBridgePortBusyError` naming a real fleet agent:
+
+        Holder PID(s): unknown ...; tmux session: tui-figrecipe.
+        Free it, then restart: `fuser -k 19007/tcp; ...`
+
+    Which runner picks up a job is not something a PR controls, so the outcome
+    was decided by scheduling rather than by the change under test — a red that
+    tells you nothing and blocks a merge. It stalled PR #1117 twice, and #1117
+    is itself the fix for the audit gate blocking two other PRs.
+
+    Binding port 0 and reading back the assignment is the standard way to ask
+    the OS for something free. The socket is closed before the test runs, so a
+    later bind of the same port succeeds (SO_REUSEADDR over TIME_WAIT is
+    exactly the case `port_is_free` is documented to treat as free).
+    """
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as probe:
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
## Deadlock — and why re-running either PR could never have cleared it

These two PRs each fix exactly what blocks the other:

| PR | fixes | was blocked by |
|---|---|---|
| **#1117** | the audit gate (PS-140, PS-231; PS-226 stays correctly masked) | the port collision |
| **#1125** | the port collision (`_PORT = 19_007`, a live a2a port) | the audit gate |

So neither could ever go green alone, no matter how many re-runs. That is not a flake pattern; it is a cycle, and the only exit is a single branch carrying both.

**#1125's run is what made this certain rather than plausible:** it produced **zero** `TurnBridgePortBusyError` — the port fix demonstrably worked — and failed purely on `test_audit_all_clean`. That isolated the second gate cleanly.

## What I did

Cherry-picked #1125's port fix (`fc626c7f`) onto #1117's branch → `0178ba03`. #1117 now carries both.

## Verified on the combined branch before pushing

| gate | condition | result |
|---|---|---|
| bridge tests | **port 19007 genuinely held** by a listener | **47 passed** |
| audit gate | auditor 0.54.0 (PATH-corrected) | **rc=0, zero finding codes** |

The held-port condition matters: 19007 is bindable on compute-04, so an ordinary local run never exercises the failure at all. Without holding it, "47 passed" would prove only that nothing was listening.

## Consequence for #1125

Redundant once #1117 lands — its commit is contained in `0178ba03`. Close as included rather than merge; merging both would apply the same change twice.

## Then

#1123 and #1124 clear on their next sync, with neither author touching anything.

ywatanabe1989:
**Superseded — included in #1117, which merged as `811c6a99`.**

This PR's commit (`fc626c7f`) was cherry-picked onto #1117's branch as `0178ba03` so that one CI run could clear both gates at once. Merging this as well would apply the same change twice.

## Why it had to be combined rather than landed on its own

#1117 and this PR each fixed exactly what blocked the other:

| PR | fixes | was blocked by |
|---|---|---|
| #1117 | the audit gate (PS-140, PS-231) | the port collision |
| #1125 | the port collision | the audit gate |

A cycle, not a flake — so no number of re-runs on either would ever have produced a green, and combining them was the only exit.

## The fix itself is live and verified

Merged in `811c6a99`. Verified before merge with the port genuinely held by a listener, because 19007 is bindable on compute-04 and an ordinary local run therefore never exercises the failure:

| | result | time |
|---|---|---|
| old (hardcoded 19007) | 8 failed | 90.79s |
| new (OS-assigned port) | 10 passed | 0.21s |

and in CI on the real self-hosted runner where the live `figrecipe` agent holds 19007: `pytest-matrix-on-ubuntu-py3.11` and `py3.13` both **SUCCESS**.

Thanks — closing as included, not as rejected.

PR #1134, opened 2026-08-18
fix(migrate): verify the job it migrated, and fail when a rename disarms it

ywatanabe1989:
## What happened

The PS-226 cutover ran on scitex-compute-04 today. All ten steps reported `ok`, and it left the fleet's **sole OAuth refresher installed and disabled**:

```
scitex-agent-container-accounts-keepalive.timer   enabled    enabled
scitex-agent-container-accounts-refresh.timer     disabled   enabled
```

For the window between the old timer being stopped and me noticing, that host had no refresher at all. The verification pass said nothing about it.

## Two defects, and both had to be true for it to stay silent

**1. It verified nothing at all.** The loop skipped `if rename.held: continue` — but a held job is only in `renames` because `--include-held` asked for it. `_select()` already applies exactly that rule:

```python
if rename.held and not (include_held and only):
    continue
```

So re-testing `held` in the verify loop silently undid the caller's own decision, and the one job being migrated was excluded from its own verification. The pass printed the `Verification` header with nothing beneath it — which reads as "nothing to report" and meant "nothing was checked".

**2. Counting unit files cannot see arming.** `verify_exactly_one` compared filename sets. Zero active supervisors satisfies "not two supervisors" perfectly, so the check could not have caught the one outcome that mattered.

## The rule

> **A rename must preserve arming.** If the old unit was enabled before the cutover, the new unit must be enabled after it. If it was disabled, it stays disabled.

Stated that way it needs no policy input and no allow-list — it reads the host's own pre-state as the expectation. That is what makes it correct on **both** live cases:

| job | old state | new state | verdict |
|---|---|---|---|
| `accounts-refresh` | enabled | disabled | **FAIL** — the incident |
| `heal-agent-auth` | disabled | disabled | OK — deliberate, it is mutually exclusive with `restart-login-expired-agents` ("enable exactly one", per its own unit description) |

A guard demanding "every timer must be enabled" would have been wrong about the second, and a guard that cries wolf gets ignored.

## Design notes

- **Arming is three-valued.** A caller that never measured enablement gets `armed_ok is None` and an `arming NOT CHECKED` verdict — never a silent pass. `ok` keeps its original narrower meaning (the installation claim), so existing callers are unchanged.
- **Enablement is read as a filesystem fact** — a symlink under `<target>.wants/`, which is precisely what `systemctl enable` creates and says it creates. So it honours the `--unit-dir` test seam and needs no subprocess.
- **Captured before the plan runs**, because the plan's own `disable` step destroys the evidence.
- The old→new unit pairing is by **suffix**, not position, so it survives any reordering of `KIND_UNIT_SUFFIXES`. That also keeps it honest for timer jobs: only the `.timer` is ever enabled (the `.service` is `static`), so only the `.timer` can appear — the data decides rather than a hardcoded rule.

## Tests

7 new, and a positive control earns them: sabotaging `armed_ok` to always-`True` turns **3 of them red**.

That control also caught one of my own tests passing for the wrong reason — it asserted on a unit name that appears in *every* verdict, so it was green whether or not the unarmed branch worked. It now asserts on the fix command, which only that branch emits.

```
126 passed  tests/scitex_agent_container/_jobs/_migrate/
3301 passed  _migrate + cli_pkg
```

https://claude.ai/code/session_01YA7NSZViCFRyftkU7bBzxH

### DIFF
--- a/src/scitex_agent_container/_jobs/_migrate/_verify.py
+++ b/src/scitex_agent_container/_jobs/_migrate/_verify.py
@@ -1,11 +1,53 @@
-"""EXACTLY ONE SUPERVISOR — the claim the migration exists to make good on.
+"""EXACTLY ONE SUPERVISOR, AND IT IS ACTUALLY SUPERVISING.
 
-"The new unit exists" is not the claim. "ONLY the new unit exists" is.
-A migration that installed the new name and left the old one behind would
-report success while doubling the supervision of the fleet's credential
-machinery, which is the precise failure this whole change is written to
-prevent. So verification counts BOTH names and treats any survivor of the
-old one as a failure, however healthy the new one looks.
+"The new unit exists" is not the claim. "ONLY the new unit exists" is not the
+whole claim either. A migration that installed the new name and left the old one
+behind would report success while doubling the supervision of the fleet's
+credential machinery, which is the precise failure this whole change is written
+to prevent. So verification counts BOTH names and treats any survivor of the old
+one as a failure, however healthy the new one looks.
+
+WHY ARMING IS PART OF THE CLAIM (measured 2026-08-19, on the real cutover)
+=========================================================================
+``sac dev migrate-job-names --only accounts-refresh --include-held --yes`` ran
+all ten steps green on scitex-compute-04 and left the fleet's OAuth refresher
+**installed and disabled**::
+
+    scitex-agent-container-accounts-keepalive.timer   enabled   enabled
+    scitex-agent-container-accounts-refresh.timer     disabled  enabled
+
+The old timer had been ``enabled`` and firing; the new one was not enabled at
+all and did not appear in ``list-timers``. For the window between the two, the
+sole-refresher host had NO refresher — and the verification pass said nothing,
+because counting unit FILES cannot see it. Zero active supervisors satisfies
+"not two supervisors" perfectly.
+
+That is §2's *a gate that cannot fail is not a gate*: the check ran, on time, in
+writing, and could not have caught the one outcome that mattered. The rule that
+does catch it is the definition of a correct rename:
+
+    **A RENAME MUST PRESERVE ARMING.** If the old unit was enabled before the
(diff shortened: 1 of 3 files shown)


ywatanabe1989:
### Pre-merge note: this makes the failure visible, it does not repair it

Worth stating explicitly rather than letting it be discovered.

The `install` step delegates to `scitex-dev ecosystem dev timer install --yes`, which **installs but does not enable** — measured today, it printed the `systemctl --user enable --now …` line rather than running it. So after this change, migrating a previously-armed job reports:

```
FAIL <job>: INSTALLED BUT NOT ARMED — [...timer] was enabled before the cutover
and is not now, so this job is supervised by nothing.
Fix: systemctl --user enable --now <unit>
```

…and exits non-zero **every time**, until someone enables it by hand. That is a loud, accurate failure where there was silence, which is the point of the PR. But the migration still leaves the job disarmed — the tool now tells you, it does not fix it.

**I deliberately did not add a `rearm` step**, and the reason belongs on the record: re-arming means `systemctl --user enable --now`, and `--now` *starts* the unit. For a job like `accounts-refresh` that rotates single-use refresh tokens, having a rename silently trigger a real rotation is a larger action than "rename this unit" promises. A tool that repairs by doing something bigger than its name is its own hazard. Report now; repair as a follow-up with the start semantics thought through separately.

### Two things checked and cleared

1. `armed_before` is computed before the `--yes` gate, so a dry run measures it and then exits at `EXIT_NO_YES` without using it. Harmless (one directory scan, no side effects), and moving it after the gate would be wrong — the capture *must* precede the plan's `disable` step whenever the plan actually runs.
2. `if not result.ok or result.armed_ok is False` deliberately does not treat `armed_ok is None` as a failure. In this CLI path both armed sets are always passed so `None` cannot occur here; the guard exists for the API's other callers, where "not measured" must not masquerade as either verdict. That three-valued behaviour has its own test rather than resting on this reasoning.

PR #1154, opened 2026-08-20
feat(state): move the birth certificate off SQLite to per-host PostgreSQL

ywatanabe1989:
Second table under the 2026-08-19 SQLite-eradication ruling, and the first to leave via the **DDL half** of the freeze rather than the import half — `state_db_incarnations.py` never imported `sqlite3`, it handed its DDL to `open_db`, which is exactly the hole that second list was added to close.

The module docstring promised this migration by name. It said the table lived in the SQLite factory *"ON PURPOSE so the separately-carded sqlite→Postgres migration carries it along instead of this PR front-running it"*. This is that promise being kept.

## Two deliberate divergences from the `verdict_dedup` template

| | verdict_dedup | incarnations | why |
|---|---|---|---|
| timestamps | unix REAL | **ISO-8601 TEXT** | callers read `born_at`/`exited_at` as strings; changing the wire format while changing the backend is a second silent breaking change riding inside the first |
| data merge | IMMUTABLE | **LAST_WRITER_WINS** | the old code was `INSERT OR REPLACE` so *"a retried launch that re-records the same id must refresh rather than crash"*; IMMUTABLE would make that retry raise |

## The invariant that moved house

`record_incarnation_exit` returns True iff a birth record existed — a missing one is False, **never an insert**, because a death with no recorded birth is a real signal. The store has no UPDATE-only verb, so the refusal SQLite's `UPDATE ... WHERE` performed is now **explicit in the function**, and the exit write sends the whole merged record so the certificate cannot be hollowed out by a partial write. Both pinned by tests.

## `KNOWN_TABLES` shrinks too

Leaving `incarnations` whitelisted would make `sac db query --table=incarnations` return an **empty result**, which reads as "this agent has no incarnations" when the truth is "you are asking the wrong database". An unknown-table error is the honest answer.

## Tests

Real PostgreSQL, no skip when it is down — a skip that reads as a pass is the defect this migration exists to remove. The throwaway-schema fixture moves to the package `conftest.py` so three files share one copy, with `psycopg` imported **inside** it so a missing driver cannot become a collection error for hundreds of tests that never touch a database.

Verified on the host: **142 passed, 1 skipped** across `tests/develop` plus the three affected modules.

The six failures in `test__rename_cards.py` seen during the full run reproduce **identically on develop@1f0cc8b9** and are unrelated to this change.

## Migration script

Dry-run verified against the **242 rows** on compute-04. It writes through the store directly rather than calling `record_incarnation_birth`, because that function stamps `born_at` with `now_iso()` — correct for a launch, and it would have restamped every birth in the fleet to the migration's own clock.

## One thing found on the way, not fixed here

All 242 of those rows have `exited_at IS NULL`. **242 births, zero deaths, ever.** The `exit-record DB mirror failed` warning has never appeared in 30 days of journald (control: journald is readable, 1.39M lines in 2 days), so the exit write is not failing — it is never called. Carded separately.

---

https://claude.ai/code/session_017Auq9hdD2jW6sHkQWo6q3T

### DIFF
--- a/scripts/migrate_incarnations_to_postgres.py
+++ b/scripts/migrate_incarnations_to_postgres.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python3
+"""One-shot: copy the SQLite ``incarnations`` table into per-host PostgreSQL.
+
+Companion to the ``state_db_incarnations`` port (2026-08-19). The code stopped
+reading SQLite; this moves the rows that are already there so the history is
+not stranded in a file nothing opens any more.
+
+RUN IT ON THE HOST, NOT IN A CONTAINER
+======================================
+``DEFAULT_DB_PATH`` resolves differently in the two places: a container gets
+its own per-agent shard under ``/state/<agent>/state.db``, the host gets
+``~/.scitex/agent-container/runtime/state.db``. The rows are on the HOST —
+measured 2026-08-19, compute-04 held 242 of them — so a container run would
+faithfully migrate an empty shard and report success.
+
+RUN IT TWICE
+============
+The table is live: ``sac`` writes a birth certificate on every launch. Run it
+once now, restart the daemons so they pick up the new code, then run it again
+to sweep anything written through the old path in between. The second run is
+expected to find few or no stragglers; that it finds ANY is the reason it
+exists.
+
+WHY IT DOES NOT CALL ``record_incarnation_birth``
+=================================================
+That function stamps ``born_at`` with ``now_iso()`` — correct for a real
+launch, destructive here. A migration that rewrote every birth time to the
+migration's own clock would look like the whole fleet was born at 00:40 on
+2026-08-20, which is precisely the "backfilled timestamp makes old things
+look new" failure. So this writes through the store directly, preserving
+every recorded value verbatim.
+
+IT IS IDEMPOTENT and it VERIFIES: each record is read back through the same
+dialect production reads through, and a mismatch is reported rather than
+counted as a success. Nothing is deleted from SQLite — the old table stays
+as a fallback for a human, untouched.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sqlite3
+import sys
+from pathlib import Path
+
+_COLUMNS = (
+    "incarnation_id",
+    "agent_id",
+    "spec_id",
+    "spec_git_sha",
+    "host",
+    "born_at",
+    "compiled_spec_json",
+    "exit_reason",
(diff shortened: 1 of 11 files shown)


ywatanabe1989:
## Constitutional gap this PR knowingly leaves, surfaced rather than discovered later

Applying the constitution to my own in-flight change. **§2: "Fail fast and fail loud. No silent fallbacks."**

This PR argues — correctly, in the module docstring — that `scitex_dev.store` is the right target *because* it refuses to fall back:

> A host whose PostgreSQL is unreachable raises `StoreTargetError` naming the DSN it could not reach. That is the intended behaviour.

And then the caller catches exactly that:

```python
# _birth_certificate.write_birth_certificate
except Exception as exc:          # stx-allow: fallback
    logger.error(...); return False

# _instances.write_instance_records
write_birth_certificate(config, instance_id)   # <- return value discarded
```

**The store's loudness is chosen deliberately and then defeated by its caller, in the same change.**

### The part that is worse than the code

This PR also *extends* that fallback's `stx-allow` reason to name its sink, because gate #1148 (merged hours earlier, by me) requires one. The reason now satisfies the gate. It does not satisfy the rule — the sink I named is journald, and I measured on the same night that a journald line here reaches nobody: the sibling exit-mirror's warning has not been read in 30 days. **A green gate stood in for a real guarantee.**

### This is not hypothetical for this table

Measured across five hosts today: **345 incarnation births, 0 recorded deaths.** The death mirror carries the same best-effort `try/except`, and its absence went unnoticed for the table's entire life. That is the pattern's observed output, not a predicted one.

### Why I am not patching it in this PR

Narrowing or removing the `except` so a store outage fails the launch would stop **every agent start fleet-wide** on a PostgreSQL blip. That is the same unmeasured blast radius that killed three other fixes today (83/121 bind sources absent; 0/29 heartbeats carry a pid; 1/118 carry an incarnation_id). The fallback's stated reason is also sound: failing a launch over bookkeeping destroys the run the record exists to document.

The rule's halves come apart here. **Fail fast** is arguably wrong for this write; **fail loud** is right and is what is missing.

The fix belongs to the launch-verdict work (`started, but its certificate did not land` is precisely the third value that design introduces), which is blocked on an operator decision. Carded as `sac-birth-certificate-fallback-re-silences-a-loud-store-20260820`.

**This does not block review of this PR** — the gap predates it and is not made worse by it. It is recorded here so a reviewer sees a known hole rather than finding it later and assuming nobody looked.

PR #1163, opened 2026-08-20
fix(tests): stop asserting on a dependency's display form, which turned develop red

ywatanabe1989:
`develop` has been red since scitex-dev 0.56.1, and **every open PR inherited it** — including ones that touch no state code at all. This fixes the two tests.

## The measurement

Ran CI's own entry point on a runner host (`exec-in-sif.sh run-in-sif.sh 3.11`), not a local subset:

```
2 failed, 17223 passed, 84 skipped in 192.78s

_state/test_state_db_incarnations.py::test_init_returns_a_locator_naming_the_postgres_endpoint
  assert '127.0.0.1:55432/scitex' in 'postgres[host=127.0.0.1 db=scitex port=55432]'

runtimes/test__fleet_env.py::test_without_the_injection_the_resolver_goes_somewhere_else
  assert '55432' not in 'postgres[ho... port=55432]'
```

## One cause, two tests

Both assert on `str(StoreTarget.locator)`, which is a **redacted display form** — the real connection string is `target.dsn`. Its rendering changed between scitex-dev releases:

| | 0.56.0 | 0.56.1 |
|---|---|---|
| set arm | `postgres://127.0.0.1:55432/scitex` | — |
| **unset arm** | `postgres://?/scitex` (no port) | `postgres[... port=55432]` (port) |

A developer machine resolves to whatever it has installed — this one has 0.56.0, where both assertions pass. CI resolves the newest. So "green here, red there" was not flakiness and not CI drift: it was two versions of somebody else's `__str__`.

## The fixes state what each test means

**incarnations** — the point is that the locator *names* the endpoint, so assert host, port and database each appear. Holds under both renderings above.

**fleet_env** — the point, per its own docstring, is that the two arms must **differ**, or the variable is inert. It checked that by asserting the port was *absent* from the unset arm, which was never the discriminator; it only worked while 0.56.0 happened not to print one for the socket case. It now compares the two arms directly — no knowledge of which field carries the difference, and no rendering can defeat it.

## Verification

- **Mutation-checked**, not assumed: pointing the unset arm at the injected DSN makes the two locators identical and the test fails (`assert 'postgres://...' != 'postgres://...'`). File restored byte-identical afterwards.
- **Verified in CI's own SIF**, because a local pass proves nothing about the version that failed.

A PostgreSQL-absence hypothesis was raised first and killed by measurement (socket present on the host *and* visible inside the SIF under CI's exact interpreter) before reproducing the actual run.

### DIFF
--- a/tests/scitex_agent_container/_state/test_state_db_incarnations.py
+++ b/tests/scitex_agent_container/_state/test_state_db_incarnations.py
@@ -141,13 +141,32 @@ def test_init_returns_a_locator_naming_the_postgres_endpoint(pg_schema: str) -> 
     scitex'``. Pinned as discovered rather than quietly weakened: an
     operator following this string reaches the right server, and still has
     to know which schema to look in.
+
+    ASSERTS ON THE PARTS, NOT THE PUNCTUATION, and that is the fix for how
+    this test broke. The quotation in the paragraph above is the locator as
+    scitex-dev 0.56.0 rendered it — ``postgres://127.0.0.1:55432/scitex`` —
+    and the old assertion spliced the tail off ``_BASE_DSN`` and looked for
+    that substring verbatim. 0.56.1 renders the same endpoint as
+    ``postgres[host=127.0.0.1 db=scitex port=55432]``.
+
+    Nothing about where the state goes changed. A developer machine pinned
+    to 0.56.0 stayed green while CI, which resolves the newest, went red —
+    so this was invisible to every local run and only a full CI
+    reproduction found it.
+
+    The locator is a REDACTED DISPLAY FORM, deliberately: the real
+    connection string is ``target.dsn``. A display form is not a contract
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
## Verified in the population that actually failed

Ran CI's own entry point against this branch on runner host compute-03, in the
same SIF CI uses — so the resolved dependency is the one that broke it:

```
scitex-dev-0.56.1
17225 passed, 84 skipped in 173.23s (0:02:53)
FAILED/ERROR count: 0
```

Against the same command on a clean clone of `develop`:

```
scitex-dev-0.56.1
2 failed, 17223 passed, 84 skipped in 192.78s
```

`17223 + 2 = 17225` — the two that failed now pass, and nothing else moved.

This matters because a local run cannot show it: this machine has scitex-dev
**0.56.0**, where both original assertions pass. That version gap is the whole
bug, so "green on my machine" was never evidence either way.

PR #1168, opened 2026-08-20
fix(worktree-gc): an ignored file was not enough to keep a worktree, and git does not help

ywatanabe1989:
**Data loss, live until now, in the unattended path.** `worktree-gc` removes a worktree that is clean + merged + idle + old enough — and `is_clean` asked `status --porcelain`, which says nothing about **ignored** files.

## Measured, with a control

```
worktree holding only GITIGNORED/lessons.md
  status --porcelain              ''               -> read CLEAN
  status --porcelain --ignored    '!! GITIGNORED/'
  worktree remove (no --force)    rc=0             -> NOTES GONE

CONTROL, same position, an UNTRACKED file instead
  status --porcelain              '?? scratch.txt' -> read DIRTY
  worktree remove (no --force)    rc=128           -> refused
```

Untracked content had **two** independent protections — this predicate *and* git refusing — and ignored content had **neither**. The asymmetry runs opposite to intuition: git guards the untracked file and deletes over the ignored one without comment.

## Where it lands

CLAUDE.md instructs every agent in this fleet to write `GITIGNORED/tasks/todo.md` and `GITIGNORED/tasks/lessons.md`. `GITIGNORED/` is ignored by name. So **the files the instructions create are exactly the class with no protection**, and the timer would have reported the removal as routine housekeeping.

The module's header states the trade it exists to make — *"REMOVE destroys work that exists nowhere else. We pick annoying."* — and this was the one class where it silently picked the other way.

## The old docstring is the instructive part

> `--porcelain` "lists untracked files by default (no `-uno`), and that default is the behaviour we want, not an accident."

Every word true. It settles **one** of the two axes on which a tree can be non-empty, while reading as a considered decision about emptiness in general. A confident note about the axis you *did* check is how the axis you did not check stays invisible.

## Found by transfer, not by review

dotfiles reported the mirror defect in their sync — git refuses to overwrite an untracked file on a merge and silently overwrites an ignored one — and asked whether anything here decides "clean, so safe to overwrite". Two candidates existed. **This one is worse than theirs: it deletes rather than overwrites.**

## Verification

Mutation-checked rather than assumed: restoring the plain `--porcelain` call reddens **exactly** the ignored arm (1 failed, 3 passed) and leaves both controls green — so the fix is not "call everything dirty", which would disable the GC and look like a fix. File restored byte-identical afterwards.

Four tests, real repos and real worktrees, no mocks (PA-306). One pins that **git itself** would have deleted the notes, because otherwise a reader may assume `worktree remove` is a second line of defence. It is not.

373 passed in the `_maintenance` suite.

## Still open, named rather than silently left

`cron/post-merge-pull.sh` gates a pull on `status --porcelain` and has the same blind spot. It overwrites rather than deletes, so it is dotfiles' exact shape; not fixed here because it is a shell path with its own test story.

### DIFF
--- a/src/scitex_agent_container/_maintenance/_worktree_gc_predicate.py
+++ b/src/scitex_agent_container/_maintenance/_worktree_gc_predicate.py
@@ -2,9 +2,14 @@
 
 A worktree is removable **iff ALL FOUR** legs pass:
 
-1. **CLEAN** — ``git status --porcelain`` is empty. Untracked files count
-   as DIRTY: an untracked file is work saved nowhere else, which makes it
-   the most expensive thing in the tree, not the cheapest.
+1. **CLEAN** — ``git status --porcelain --ignored`` is empty. Untracked
+   AND IGNORED files count as DIRTY: either one is work saved nowhere
+   else, which makes it the most expensive thing in the tree, not the
+   cheapest. ``--ignored`` was missing until 2026-08-20 and its absence
+   was a data-loss bug — git refuses to remove a worktree over an
+   untracked file and removes one over an ignored file without comment,
+   so ignored content had neither this guard nor git's. See
+   :func:`is_clean` for the measurement and the controls.
 2. **MERGED** — the ancestor check (``rev-list --count <base>..<head>``
    is 0 against ``develop`` or ``main``) OR a MERGED PR exists for the
    branch. BOTH styles are required: a squash-merged branch is not an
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
## A live near-miss, reported by scitex-cards after reading this

Their working tree is in exactly this state right now — `status --porcelain --ignored` returns:

```
!! .pytest_cache/
!! GITIGNORED/
!! src/scitex_cards/**/__pycache__/   (many)
```

**No `??` and no ` M` entries at all — only `!!`.** So the pre-fix predicate reads it CLEAN, `worktree remove` returns rc=0, and `GITIGNORED/` goes with it.

What was inside (1225 bytes, written 2026-08-18):

```
GITIGNORED/REFACTORING.md
  _health.py is 517 lines, over the 512 limit. Split the five inline
  probes into _health_probes.py.
```

…and that plan specifies a new health check named **`hook_consumers_registered`** — the check that would have detected today's other finding, that sac's card-event consumer has never been invoked fleet-wide (#1167). Written 2026-08-18, not yet implemented, sitting on the path that deletes without comment.

They have moved it somewhere safe.

## Two things this sharpens

**The blast radius is not "scratch files".** The class is *work that exists nowhere else, deliberately kept out of version control* — which for this fleet includes design notes, saved patches, rescue bundles, and the `GITIGNORED/tasks/lessons.md` that CLAUDE.md instructs every agent to write. The instructions and the collector disagree, and only one of them runs unattended on a timer.

**That conflict is fleet-wide, not a sac detail**, and has been raised with the operator separately: the documented place to record lessons is the place with no protection.

ywatanabe1989:
## The fixture was sabotage-checked, because a fixture that never ran is green

dotfiles hit this three separate times tonight while working the mirror defect: a test whose **subject never executed** passes identically to one whose subject behaved correctly. In their worst case the sweep died at `FATAL: no origin remote`, and both content arms passed — `lessons.md` survived because nothing ran. Only a control caught it.

So rather than reason that these tests are safe, I broke the fixture and looked. Pointing `worktree add` at an uncreatable path — the subject is never set up:

```
4 failed in 0.37s

FAILED test_a_worktree_holding_only_ignored_notes_is_not_clean
FAILED test_an_untracked_file_still_keeps_the_worktree
FAILED test_a_genuinely_empty_worktree_is_still_collectable
FAILED test_git_itself_would_have_deleted_the_ignored_notes
```

All four fail loudly, including the two controls. There is no path here by which a broken fixture yields a green run: the arms that write into the worktree raise `FileNotFoundError`, and the empty-worktree arm gets `(None, KEEP_DIRTY)` from an unreadable path, which is not the `True` it asserts.

File restored byte-identical afterwards.

## Why this check and the mutation check are different

They fail in opposite directions and neither substitutes for the other:

| check | perturbs | answers |
|---|---|---|
| mutation | the **fix** | does any test notice the defect returning? |
| sabotage | the **fixture** | did the subject actually run? |

A suite can pass mutation testing and still be measuring nothing — mutation perturbs the code under test and asks whether assertions react, which presumes the setup worked. Sabotage asks about the presumption.

PR #1200, opened 2026-08-23
fix(image): raise the scitex-cards floor to 0.49.0 and gate it BY SYMBOL, so an image that destroys comments cannot ship

ywatanabe1989:
The live `sac-base` image resolved **scitex-cards 0.48.0**. Below 0.49.0, `comment_task` / `update_task` rewrite a card's `card_json` from whatever copy the caller happens to hold, **dropping every comment row that copy has not seen**. A peer's comment written between your read and your write is destroyed, silently, and the card looks fine afterwards.

Nothing failed. Measured on the deployed build-context def on scitex-compute-04:

```
628:        "scitex-cards>=0.41.0" \
grep -c _merge_unseen_comment_rows  ->  0
```

## Why the existing gate could not catch it

The `.def` already ships a fail-loud staleness gate that asserts **by symbol**, and its own comment explains why version strings are not admissible as evidence — an orphaned or duplicate `.dist-info` freezes what `importlib.metadata` reports while the code on `sys.path` is a different tree.

But for *this* defect its scitex-cards probe is a bare

```python
import scitex_cards  # noqa: F401  (the import itself is the check)
```

and **0.48.0 imports perfectly**. The gate ran, passed, and proved nothing about the defect — a check that cannot fail, for exactly one class of failure.

## Why a version comparison would also have been inert

The obvious remedy — exec `scitex-cards --version` in the built SIF and compare against the intended floor — would **not** have caught this. The floor read `>=0.41.0`, and a resolved 0.48.0 satisfies it.

A floor is a *request to the resolver*. It becomes evidence only once something asserts the resolver honoured it **and** the floor names the version you actually need. Both halves are required, which is why this PR does both.

## What changes

| | |
|---|---|
| floor | `"scitex-cards>=0.41.0"` → `">=0.49.0"` (base.def, and **both** pins in scitex.def) |
| symbol | `from scitex_cards._mirror_rows import _merge_unseen_comment_rows` |

The symbol discriminates the exact defect — measured against both artifacts rather than recalled:

```
0.49.0   _mirror_rows.py:108  def _merge_unseen_comment_rows(...)
         _mirror_rows.py:246  "_merge_unseen_comment_rows",   (__all__)
0.48.0   0 files
```

Importable in 0.49.0, `ImportError` below it.

## Four copies, not one

The probe exists **four times** — the shipped asset plus three embedded heredocs (`base.def`, `scitex.def`, `spartan-sif-bake.sh`) — each gating a different path: local bake, layered bake, Spartan bake, and the master-side verify of a pulled SIF. A floor bump reaching one copy leaves the others baking images that carry the defect while reporting a clean gate. All four are updated.

`test_bake_script_probe_matches_the_wheel_probe_verbatim` already enforces byte-identity between the Spartan heredoc and the wheel probe — and it **caught this PR's first draft**, when the two comment wordings diverged. That gate covers one pair; the new parametrized test covers the `.def` copies it cannot reach, which are not byte-comparable to the asset because they carry additional gates around it.

## Tests

```
tests/scitex_agent_container/containers/test_sif_symbol_probe.py
+ tests/scitex_agent_container/cli_pkg/test__remote_bake_core.py
52 passed

plus the 9 other suites that read these files, run individually:
183 passed in total, none red.
```

## Not claimed

This does **not** explain how the fleet got 0.49.0 today. The build at 15:04Z used the *unmodified* def with the 0.41.0 floor and resolved 0.49.0 only because that is now the newest release satisfying it — resolver luck, not policy. That is the argument *for* this PR, not evidence for it: the next release, or a resolver that picks differently, would silently reintroduce the defect and nothing would go red.

## Pre-existing, filed separately

`tests/integration` as a directory hangs on **unmodified develop** as well as on this branch (1327 tests collect in 0.97s; execution never returns, even under `--timeout=45 --timeout-method=thread`). Confirmed with a clean-checkout control, so it is not from this change. Carded as `tests-integration-hangs-on-develop-20260823`.


ywatanabe1989:
## HOLD — do not merge. The floor this PR raises to is a broken release.

scitex-dev reported a regression in scitex-cards 0.49.0 and I verified it independently against the installed wheel, not by accepting the report.

**The defect.** `_merge_unseen_comment_rows` — the very function this PR gates on — indexes its rows **positionally**:

```python
# _mirror_rows.py:194
key = _key(row[0], row[1], row[2], row[3])
```

and again in the recovery dict below it (`{"author": row[0], ...}`).

**Why that raises.** The package's own `_store_backend.py:59-60` states the rule this breaks, verbatim:

> BY COLUMN NAME, not position. `sqlite3.Row` accepts both `r["id"]` and `r[0]`; psycopg's `dict_row` accepts only the former and raises

The fleet store is PostgreSQL, so `dict_row` yields real dicts and `row[0]` is `KeyError: 0`. Corroborated at two more sites that already learned this (`_store_uuid.py:224`, `_store_canonical_read.py:440`, both commented "NOT row[0]").

**The blast radius is inverted from the defect it fixes.** The merge exits early when a card has no prior rows, so:

```
card with 0 comments  ->  OK
card with >=1 comment ->  KeyError: 0
```

Every new card works; every card with history fails. `comment_task` is the board's main verb.

**So this PR must not land as written.** Its floor (`>=0.49.0`) would make 0.49.0 the *minimum* fleet-wide, i.e. it would push the outage everywhere the image is rebuilt.

## What I am doing

Holding as draft until scitex-cards ships the point fix, then re-targeting the floor to `>=0.49.1` and re-running. Nothing else in the change needs to move: the four-copy symbol gate, the tests, and the reasoning about floors-versus-assertions all stand.

## The part of this that is a lesson about my own gate

scitex-dev made a point I should record rather than deflect: **verifying the artifact contains the fix is not verifying the fix works.** My post-build assertion checks that `_merge_unseen_comment_rows` is importable inside the SIF. That check passes on 0.49.0 — the function is present and it is broken. A symbol probe proves *presence*, never *behaviour*.

The gate that would have caught this is behavioural: write a comment to a card that already has one and assert the table count. That cannot live in the image bake (no live store at build time), so it belongs in a post-deploy smoke test. Filed as follow-up rather than bolted onto this PR.

PR #1225, opened 2026-08-26
fix(tests): the turn-bridge port gate must not depend on a live port

ywatanabe1989:
## The defect

The bridge tests pick a port by binding `0` and reading the assignment back, then **close** the socket at module import:

```python
def _free_a2a_port() -> int:
    with socket.socket(...) as probe:
        probe.bind(("127.0.0.1", 0))
        return int(probe.getsockname()[1])   # socket CLOSED here
_PORT = _free_a2a_port()                      # once, at import
```

Everything after that is a gap. On a self-hosted runner with N xdist workers, several CI legs and real agents all churning the ephemeral range, something takes the port before the test's bind probe runs.

Measured 2026-08-26, worker `[gw5]`, port 59935 — two `start_turn_bridge` tests red on `pytest-matrix-on-ubuntu-py3.11`, in **#1222, whose diff touches only `cli_pkg/_dev_jobs_backend.py`**. Nothing under `runtimes/`. The red was decided by which runner picked up the job, not by the change under test.

## This is the second round, and the file is honest about the first

The port used to be the literal `19007` — inside the live a2a range `[19000,19999]` and genuinely held by the `figrecipe` agent on the self-hosted runner. That stalled **#1117 twice**. Moving to an OS-assigned port turned a *deterministic* collision into a *probabilistic* one. Strictly better; still a coin flip on a busy machine.

## The fix is this file's own house pattern

`start_turn_bridge` already accepts `port_free_fn` as a first-class parameter, and this very file already injects the `False` half to test the busy path:

```python
port_free_fn=lambda _h, _p: False    # existing, busy-path test
```

This adds the symmetric `True` at the eight call sites that need a free port. Dependency injection through a declared seam — not a mock, so **PA-306 is satisfied**.

**Deliberately not changed:** the `"auto"` port test (never reaches the gate) and the busy-path tests (want a busy port on purpose). Gate coverage is unchanged.

## Verified two ways

| condition | result |
|---|---|
| whole file, normal | 47 passed in 5.39s |
| `_PORT` forced to a socket **held open** — genuinely occupied, the exact condition that reds CI | **8 passed in 0.16s** |

That 0.16s is the evidence: nothing polled the gate at all, where before these tests would have polled and raised `TurnBridgePortBusyError`.

One caveat, stated because it bounds the claim: the `stop_turn_bridge` tests legitimately wait for a port to be *released*, so pinning a port permanently busy contradicts their premise and they poll to timeout. That is correct behaviour under a false premise, not a defect, and it is why the control above is scoped to the eight tests it actually applies to.

### DIFF
--- a/tests/scitex_agent_container/runtimes/test__tui_turn_bridge.py
+++ b/tests/scitex_agent_container/runtimes/test__tui_turn_bridge.py
@@ -17,6 +17,7 @@ assert + STX-TQ003 descriptive names.
 
 from __future__ import annotations
 
+import errno
 import json
 import shutil
 import socket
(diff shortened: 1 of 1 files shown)


ywatanabe1989:
## Correcting the mechanism claim in this PR's description

An adversarial review of the diagnosis (three independent lenses) upheld the conclusion — the failure is pre-existing/environmental, not caused by #1222 — but it flagged that I stated the *mechanism* more confidently than the evidence supports. Correcting that here rather than leaving it in the description.

**What I wrote:** that on a loaded runner "several CI legs and real agents all churning the ephemeral range" take the port before the bind probe runs.

**What is actually established:** the port was **not free** at probe time. That is all. In every one of the three historical occurrences the error reads `Holder PID(s): unknown (install lsof/ss/fuser to identify)` — the holder was **never identified, in any of them**. Two reproduction attempts returned null results (0/25 stolen at a 10s window; 0 re-handouts in 4000 simultaneous binds). Ephemeral re-handout is an inference that fits the evidence, not a measurement.

One reviewer proposed a better-fitting account that needs no external process at all: another xdist worker's own port draw colliding with a stale `_PORT`, then binding a real listener. That is also an inference.

Two further corrections worth recording:

- **Contention is larger than I implied.** `run-in-sif.sh:280-283` documents up to *three* matrix legs co-tenant on one 32-core box — so the concurrent population is legs × workers (~96 processes), not the 8 I described.
- **"Reservation" is the wrong word** for what `_free_a2a_port` does, and it is the false premise the original helper's own docstring encodes. Nothing is reserved; the socket is closed on exit from the `with`.

## Why none of this changes the fix

The fix does not depend on knowing who took the port. It removes the question: the eight tests that are about argv, pid and host no longer consult a live port at all. Whatever held 59935 — kernel re-handout, a sibling worker, or a real agent — an injected predicate is immune to it.

The one nuance I want on the record, because it cuts against me: this PR is not strictly causally inert with respect to scheduling. Adding tests changes the collected item count, which changes interleaving, so "but for that PR, that particular run might have been green" may be true in a narrow counterfactual sense. That is not the same as causing the defect — and relatedly, a reviewer measured that the +4 item delta blamed on #1222 was mostly #1223's three tests merging in between; #1222 contributed one.

PR #1235, opened 2026-08-28
feat(state): comms_grants to PostgreSQL — and a revoke that stops forgetting

ywatanabe1989:
Module 2 of the SQLite eradication, after the diary (#1233). Chosen next on
**blast radius** rather than size: one caller path, against 33 for
`state_db_nodes`.

This is an ACL table, so the migration had to preserve three properties a
round-trip test would not notice losing.

### Revoke is no longer a DELETE

The store's only removal is `hide`, and that is a strengthening. Under
`DELETE`, *"never granted"* and *"granted then revoked"* were
indistinguishable — for an ACL table, the difference between a clean history
and an unanswerable audit question. The row and its history stay readable
through `include_hidden` and in the oplog, while every default read treats it
as absent, so `has_grant` still denies the instant a grant is withdrawn.

That flip is the **first test in the file**: a migration that stored grants
perfectly and forgot to deny would read green on every round-trip assertion
and be a security regression.

### The listing order is the HLC, not `created_at`

The SQLite docstring recorded why it ordered by `rowid`: `created_at` ties on
bulk-imported peer rows and skews across hosts, so a foreign row sorted into a
plausible-looking position instead of standing out — which is what once let a
leaked `-> lead` grant hide inside the listing. `rowid` does not exist here;
its correct successor is the hybrid logical clock, built for exactly this.

`test_listing_order_is_insertion_order_not_created_at` writes a row with an
**older** `created_at` than one already stored and asserts it still lists
last. Ordering by `created_at` fails it.

### Re-granting stays idempotent, and gains a case

A re-grant of a live pair leaves the row and timestamp untouched. A re-grant
of a **revoked** pair un-hides it — impossible to express under `DELETE`, and
treating it as a no-op would leave an operator unable to restore a grant they
had just withdrawn.

### Other decisions

- `MULTI_WRITER`: a grant is created on one host, revoked by an operator from
  another, and bulk-imported from peers. Under `SINGLE_WRITER` the first
  revoke-from-elsewhere would be an illegal write — the store's own policy
  docs give this reasoning for the card store.
- `db_path` is gone from all four signatures. It **stays** in
  `_acl.check_send_acl`, whose six other lookups are still on SQLite.

### Data migration is not optional here

`scripts/migrate_comms_grants_to_postgres.py`, dry-run by default. Measured on
compute-04's host `state.db`: **31 live grants**. Shipping the code without
moving them would not lose history quietly — it would **deny 31 sends that are
allowed today**. Run it before the restart that picks up this code, and twice,
since the spawn path auto-grants.

### Verification is partial

All 12 tests **skip** on this host — its loopback is a read-only replica and
the fixture correctly refuses to create schemas in production. They collect
cleanly and the module imports, but the behaviour is unverified until CI,
which has a throwaway database, runs them.

### DIFF
--- a/scripts/migrate_comms_grants_to_postgres.py
+++ b/scripts/migrate_comms_grants_to_postgres.py
@@ -1,0 +1,183 @@
+#!/usr/bin/env python3
+"""One-shot: copy the SQLite ``comms_grants`` table into PostgreSQL.
+
+Companion to the ``state_db_grants`` port (2026-08-28). The code stopped
+reading SQLite; this moves the rows already there so live permissions are not
+stranded in a file nothing opens any more.
+
+THIS ONE IS NOT LIKE THE DIARY. The diary held history: losing a heartbeat row
+costs an observation. ``comms_grants`` holds AUTHORISATION. A grant that fails
+to migrate does not go missing quietly — it DENIES a send that used to be
+allowed, and the agent on the other end sees a refusal it cannot explain. So
+run this BEFORE the restart that picks up the new code, not after, and read
+the verify line rather than the exit code.
+
+A DRY RUN IS THE DEFAULT. Pass ``--commit`` to actually write.
+
+RUN IT ON THE HOST, NOT IN A CONTAINER
+======================================
+``default_db_path`` resolves differently in the two places: a container gets
+its own per-agent shard under ``/state/<agent>/state.db``, the host gets
+``~/.scitex/agent-container/runtime/state.db``. A container run would migrate
+an empty shard and report success — the same shape that let the state-write
+outage look finished for four days.
+
+RUN IT TWICE
+============
+Grants are written by a live daemon (``_lifecycle/_instances`` auto-grants on
+spawn, and the approval path grants on unblock). Run it once now, restart so
+the daemons pick up the new code, then run it again to sweep anything the old
+path wrote in between.
+
+WHY IT DOES NOT CALL ``grant_send``
+===================================
+``grant_send`` stamps ``created_at`` with ``time.time()``. Calling it here
+would rewrite every grant's age to the migration moment and destroy exactly
+the audit data the column exists for — "since when could this agent send
+there?" would answer "since the migration", for all of them. The rows are
+written through ``Store.put`` with their original timestamps intact.
+
+WHAT IT DOES NOT MIGRATE, and why that is correct
+=================================================
+Nothing. The SQLite table has no revoked rows to carry — a revoke was a
+DELETE, so the history it would have carried is already gone. The new store
+hides instead of deleting, so revocations from here on stay auditable; this
+script cannot retroactively recover the ones that were deleted before it.
+"""
+
+from __future__ import annotations
+
+import argparse
+import os
+import sqlite3
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "src"))
+
+from scitex_agent_container._state.state_db_grants import (  # noqa: E402
+    GRANTS_STORE,
+    _open,
+)
+
+TABLE = "comms_grants"
+COLUMNS = ("sender_name", "target_name", "created_at", "note")
+
+
+def default_db_path() -> Path:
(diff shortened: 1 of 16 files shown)


ywatanabe1989:
**Converted to draft. This should not merge, and my reason for picking this module was wrong.**

## The measurement error

I chose `comms_grants` as module 2 on "blast radius: 1 caller". That number counted references to the **module name** `state_db_grants`. But all four primitives are re-exported from `state_db_nodes`, so real callers import them from there and never mention this module. I measured a proxy that did not represent the thing.

The true figure:

```
81 call sites across 16 files
32 of them pass db_path
```

Not one caller. `_listen/_acl.py`, `_lifecycle/_instances.py`, `cli_pkg/a2a_group.py`, plus ten test files. This is an ACL primitive woven through the listen server, and it was the *least* safe of the remaining five to take second, not the most.

## Why it cannot proceed as-is

CI shows the shallow half: **28 failures**, mostly `TypeError: has_grant() got an unexpected keyword argument 'db_path'`. I updated the two `src` callers and never searched `tests/` — the wrong population again, in the same PR that the ranking error is in.

The deep half is what blocks it. Those 32 `db_path` sites use a **temp SQLite file for isolation**. On PostgreSQL the equivalent is the `pg_schema` fixture, which **skips** where no cluster exists. On the runner that produced this log, `SCITEX_STORE_DSN` is the dead `port=1` guard, so `pg_schema` would skip — converting roughly thirty real ACL and approval-flow tests into skips.

That is buying green with coverage, on the security path. Per the constitution: a gate that cannot fail is not a gate, and it is worse than deleting it because the config still lists it.

## What unblocks it

CI needs a **guaranteed** PostgreSQL, not an incidental one. That work already exists in this tree as `feat/ci-provisions-its-own-database` (whose own last commit is "a PostgreSQL gate that skips silently is a gate that passes", #1220 — the same lesson).

Until that lands, migrating any ACL-adjacent module trades verified security behaviour for a green tick. The diary (#1233) was safe to move because it is telemetry and its failure mode is a lost row; `comms_grants` decides whether a send is permitted, and its failure mode is a denial.

## Order I intend to follow instead

1. land the CI-provisions-its-own-database work
2. then this PR, with all 81 call sites migrated and the ACL tests running for real
3. `state_db_acl_policy` and the rest afterwards

Leaving the branch intact — the module, the 12 flip tests and the 31-row migration script are all still correct work; the prerequisite is what is missing.

ywatanabe1989:
**Retracting the blocker I claimed in the previous comment. It was wrong, and the check that settles it is one I should have run before writing it.**

I said this PR was blocked on `feat/ci-provisions-its-own-database`, and that merging it would convert ~30 ACL tests into skips. Both parts are false.

## What I actually measured

`feat/ci-provisions-its-own-database` is **already merged** — PR #1221, merge commit `6b784860`, 2026-08-26 — and it **is in this PR's base**:

```
merge-base --is-ancestor 6b784860 9a1ee121  ->  YES
```

The worktree branch I looked at is a stale local copy, 13 commits behind develop. I read the branch and inferred the work was outstanding, without checking whether it had landed.

And the skip claim does not survive the log either:

```
grep -c "PostgreSQL fixture cannot use"  ->  0
```

Zero. Every skip in that run is unrelated (`scitex_todo` import, mount namespace). So `pg_schema` **connected** rather than skipping, which means the 12 flip tests in `test_state_db_grants.py` **ran and passed** — they are absent from FAILED and absent from SKIPPED, among 17,828 passed.

I had treated their absence from the log as evidence of skipping. Absence from a pytest log is equally consistent with passing, and I did not distinguish the two before writing a caution that stopped the work.

## What is actually left

The 28 failures are exactly what they look like: `TypeError: ... unexpected keyword argument 'db_path'`, at test call sites I never enumerated because I searched `src` only. 32 sites across ~10 files. Mechanical.

The caller-count error from the previous comment **still stands** and is still worth recording — 81 call sites, not 1, because the primitives are re-exported from `state_db_nodes`. That was a genuine mistake in how I ranked the module. But it is a reason to be careful with this PR, not a reason to stop it.

Taking it out of draft and fixing the call sites.

PR #1242, opened 2026-08-28
feat(state): land the a2a port ledger port — its upstream blocker is fixed

ywatanabe1989:
#1206 wrote this module on 2026-08-24 and held it deliberately. Its own test
carried an xfail saying why:

  "BLOCKED UPSTREAM on scitex_dev.store 0.56.5 ... concurrent same-node writes
   race the oplog (origin, seq) allocation and escape as a raw psycopg
   UniqueViolation ... Do NOT merge this module while this xfails — the SQLite
   version it replaces handles this case, and shipping it would reintroduce the
   bug that killed v0.21.19."

and, crucially:

  "xfail turns green the moment the store is fixed, so this file tells us
   instead of us having to remember."

It did exactly that. scitex-dev 0.56.6 added the bounded (origin, seq)
oplog-allocation retry; 0.56.8 is what the fleet now runs. Re-running that file
unchanged gives 14 passed, 1 XPASSED, the xpassed one being the held test.

So the marker comes off. An XPASS is not an assertion — with strict=False a
regression would quietly read as an expected failure again — and the whole
point of the ratchet was to be removed once it fired.

WHAT THIS IS AND IS NOT. Additive only: it adds port_allocator_pg.py and its
tests. It does NOT touch port_allocator.py, does NOT remove a2a_ports from the
SQLite schema, and has no src/ importers. Landing it is the PORT; the CUTOVER
(moving callers, deleting the SQLite path) is separate and must satisfy the
pin-reclaim invariant merged today as #1239 — a released port leaves a
tombstone that still occupies the store identity, and a naive guard makes a
pinned agent unrestartable.

Opened as a fresh branch off current develop rather than force-pushing #1206,
which is 30 commits behind. #1206 should be closed in favour of this.

VERIFICATION CAVEAT, stated because it matters: the 14-passed/1-xpassed run was
made against the fleet PRIMARY. Current develop's pg_schema fixture now REFUSES
that ("Tests must never create schemas in production; point SAC_TEST_PG_DSN at
a throwaway database") — a guard added after #1206 was written, and a correct
one. The evidence is real but was gathered in a way the repo now forbids; CI,
which has a proper throwaway database, is the authority here.

Supersedes #1206 (30 commits behind develop; opened as a fresh branch rather than force-pushing it).

### DIFF
--- a/src/scitex_agent_container/_state/port_allocator_pg.py
+++ b/src/scitex_agent_container/_state/port_allocator_pg.py
@@ -1,0 +1,342 @@
+"""``a2a_ports`` — the A2A port claim ledger, on PostgreSQL only.
+
+Step 4 of the operator's sqlite→PostgreSQL migration (approved 2026-08-24).
+The move is by ADOPTING :mod:`scitex_dev.store`, the fleet's own store
+primitive, exactly as :mod:`.state_db_incarnations` did — not by sac growing
+a private psycopg layer. That primitive implements the operator's 2026-08-19
+rule ("fail fast, fail loud, no fallbacks"): it resolves ``SCITEX_STORE_DSN``
+or the per-host PostgreSQL and has deliberately NO SQLite fallback, so a host
+whose database is unreachable raises rather than quietly allocating ports
+into a file nobody reads.
+
+``db_path`` IS GONE from every function. It named a SQLite file; there is no
+file. Callers that threaded it through simply stop.
+
+THE IDENTITY IS THE PORT, AND THAT IS THE WHOLE DESIGN
+======================================================
+The obvious port of this module keys the store on ``name``, because
+:func:`get_port` and :func:`release_port` both look up by agent. That would
+be wrong in a way tests would not catch.
+
+The invariant this module exists to hold is ``UNIQUE(port)`` — a DIFFERENT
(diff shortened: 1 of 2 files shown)


ywatanabe1989:
Closing in favour of #1243 (moves the a2a port ledger to PostgreSQL), which supersedes this.

This PR is +617/-0: it ADDS `port_allocator_pg.py` and its suite and deletes nothing. `port_allocator.py` keeps `CREATE TABLE IF NOT EXISTS a2a_ports` and every reader keeps calling it, so landing this would leave one table with two implementations and readers split across them — the half-migrated state that raises nothing and is worse than not starting. It also does not reduce the SQLite table count, which is the actual goal.

#1243 does the whole cutover: it deletes the DDL, the index and the SQL, converts `_instances.py`, `_registry_endpoints.py`, `_peer_resolve.py` and `ports_cmds.py`, ships `migrate_a2a_ports_to_postgres.py`, and updates `test_sqlite_footprint_frozen.py`.

The one thing worth carrying across from here is the test fix in 12328957: `test_init_reports_where_the_state_went` asserted `'55432' in locator` — the fleet's per-host port — which fails on CI's ephemeral port. If #1243's suite pins the port rather than the `postgres[` prefix, it will hit the same failure.

PR #1243, opened 2026-08-28
fix(state): move the a2a port ledger to PostgreSQL — a released port must stay re-claimable

ywatanabe1989:
Moves `a2a_ports` — the per-agent A2A port claim ledger — off `state.db` onto
per-host PostgreSQL via `scitex_dev.store`. Every reader moves in this PR.

## The trap this had to survive, and the red control that proves it is real

The store's only removal is `hide`, and `put` reads with `include_hidden=True`,
so a released claim leaves a **tombstone that still occupies the store
identity**. `get` answers ABSENT while `put(NEW_RECORD)` answers TAKEN.

Handled naively that is not a wrong answer, it is fleet-down. Every ordinary
restart of a pinned agent runs `release_port` then
`claim_port(explicit=…, explicit_is_pin=True)`, so the agent is told its own
port is already claimed and never comes back.
`tests/.../test_port_allocator_pin_reclaim.py` (#1239) measured that hazard
before this migration existed; this is the migration it was written for.

Measured, not assumed — replacing the tombstone branch of `try_claim` with the
naive guard and re-running the merged test:

```
RuntimeError: a2a port 21500 already claimed by 'another agent'; cannot pin for 'alpha'
1 failed, 3 passed
```

Restored, the same test passes. The `unhide` is load-bearing and the test is
not vacuous.

## The identity is the PORT, not the agent

The obvious port keys the store on `name`, because `get_port` and
`release_port` both look up by agent. That is wrong in a way tests do not
catch: the invariant this module exists to hold is `UNIQUE(port)`, a
**different column from the lookup key**, and keyed on `name` two agents
claiming one port are simply two records — mutual exclusion gone, every unit
test still green. That is the collision that killed v0.21.19.

So `port` is the sole IDENTITY and the store's own uniqueness carries the
invariant structurally. The price is stated rather than hidden: "which port
does this agent hold" inverts into an O(n) scan over a ledger bounded by the
configured range (1,000 records at the default; tens in practice).

## The order of two reads in `try_claim` is load-bearing

The revision is read **before** the row it decides on. That token is the
compare-and-swap for the tombstone takeover, and capturing it first is what
makes every interleaving safe. Read the row first and a competitor who takes
the port after our read leaves us with a stale-but-matching token: our write
lands on top of theirs and *both* callers are told they hold the port — exactly
what the SQLite single-statement insert existed to prevent.

`Store.revision` is the accessor, not `Row.seq`: those are different columns
(`_revision` vs `_seq`) and only the former is what `check_revision` compares.

## Scope

`db_path` is gone from `get_port` / `claim_port` / `release_port` /
`list_claims` and from `cli_pkg.collect_ports_data`; it named a SQLite file and
there is no file. 74 call sites across six test modules were rewritten
**AST-guided rather than textually**, because several of those modules also
pass `db_path` to state_db helpers that are still SQLite-backed and a blind
strip would have broken them silently.

`port_allocator.py` splits: the storage adapter becomes
`port_allocator_store.py` (per-file line cap, and the same responsibility split
`state_db_grants` made out of `state_db_nodes`); `port_allocator.py` keeps the
allocation policy and re-exports the store surface, so no import site changed.

`claimed_at` becomes epoch REAL, matching every other migrated timestamp in
`_state`. Its only consumer is `sac ports --json`, which passes it through.

**Not touched, deliberately:** `state_db_export.py` and `cli_pkg/db_group.py`.
`a2a_ports` is absent from `KNOWN_TABLES` and from the export path, and another
change is in flight on those files.

## The migration script was run, not merely written

`scripts/migrate_a2a_ports_to_postgres.py` — #1237 exists because two sibling
scripts could never write a row, so this one was exercised end to end: 2
written, read back through `list_claims` / `get_port`, idempotent on re-run,
unparseable timestamps reported rather than costing the claim, and a duplicate
port in the source refused rather than arbitrated.

An unmigrated claim is not a lost observation: the allocator would see a bound
port as free and hand it to the next agent, whose runner then loses the bind.
**Run it before the restart that picks up this code, on the host, not in a
container.**

## The isolation canary moved, and had to

`test_state_db_test_isolation.py` used `claim_port` to prove the per-test
`state.db` isolation is FUNCTION-scoped. `claim_port` no longer touches
`state.db`, so left alone those two tests would have gone on passing while
measuring a different database — the exact vacuous test that file exists to
prevent. They now use `record_instance_start`, which is still SQLite; the
docstring records that the canary is weaker in *consequence* while testing the
same guard.

Two assertions in the concurrency module were **reworded rather than
re-pointed**, for the same reason: "no raw `sqlite3.IntegrityError` escaped" is
trivially true once the driver is psycopg (it is now backend-neutral: everything
that escapes must be the diagnostic `RuntimeError`), and a direct
`SELECT … FROM a2a_ports` cannot run at all (now a `list_claims()` read).

## Verification

Against a **real PostgreSQL 18 provisioned for the run**. Loopback on this host
is a read-only replica of the fleet cluster, so `pg_schema` would otherwise have
skipped every one of these — and a skip is not a pass.

| suite | result |
|---|---|
| `test_port_allocator_pin_reclaim.py` | 4 passed |
| `test_port_allocator.py` | 20 passed |
| `test_port_allocator_concurrency.py` | 9 passed (16-thread contention) |
| `_state` + `develop` + `cli_pkg/test__send.py` | 1455 passed, 3 skipped |
| `_lifecycle` + `_listen` + `cli_pkg` sweep | 125 passed |

AST walk over `ast.Call` matching both `ast.Name` and `ast.Attribute`: `open_db`
call sites under `src/` go **52 → 47**, the five that left all in
`port_allocator.py`.

`ruff` on the changed files: the only findings are two `E731` lambdas that are
byte-identical on `origin/develop`.

## Two corrections, in the second commit

**A `--dry-run` really does claim a port, and one failure was mine.**
`test_fleet_template.py::test_start_params_file_dry_run_exits_zero` fails on
this branch. `sac agents start --dry-run` still reaches `resolve_a2a_port` →
`claim_port`, so a dry run takes an a2a claim — already true under SQLite, where
the claim just landed in the per-test `state.db` and nothing said so. The
fixture now takes `pg_schema`. That a dry run mutates state at all is
pre-existing and left alone.

I nearly shipped that as "pre-existing", because **my baseline control was
lying**: it ran `PYTHONPATH=<origin/develop src>` and pytest ignored it —
`pyproject.toml` sets `pythonpath = [".", "src"]` and its own comment says that
entry beats "BOTH site-packages and any inherited `$PYTHONPATH`". The
"baseline" therefore executed the branch under test and reproduced the identical
failure, which reads exactly like a pre-existing one. The valid control puts
develop's file **in the tree**, because the tree is what pytest loads: that way
the module is 20 passed, and with mine 1 failed.

Re-run correctly, the three `test__start.py::TestGroupOption` timeouts **are**
pre-existing — same three tests, same `_start_parallel.py:149` subprocess wait,
with develop's allocator in the tree.

**A worry I raised against my own design did not reproduce.** `name` was the
SQLite PRIMARY KEY, so one agent could not hold two ports; keying on `port`
removes that, and in principle two concurrent auto-claims for one agent could
each win a different port and both return happily — a leak that raises nothing.
Measured through the public surface on real PostgreSQL 18: **15 runs at 2, 4, 8,
16 and 32 threads, all barrier-released, every run ended with exactly one claim
and zero errors** (the store serialises `Store.__init__` per schema behind its
own advisory lock, so racers queue at construction and every later thread finds
the claim in the idempotent fast path).

I wrote the settlement logic and a test, then **deleted both**: the test passed
*without* the fix, so it was a control that could not fail. What ships is the
measurement, in `port_allocator_store`'s docstring, so the next reader who does
see one agent holding two ports knows the question was asked and what held it.

## What CI caught that local runs could not

`agent_start` claims an a2a port **unconditionally** — on `--dry-run` too — so
every test that reaches it needs a database once the ledger is PostgreSQL. That
set is far wider than the modules naming the allocator, and it is not
enumerable by reading imports.

A hosted CI runner has no PostgreSQL, so the suite's deliberately-unreachable
DSN turns each of those into a `StoreTargetError`, and CI's log prints
`FAILED <id> - <exception>`. Filtering that on the store error gave the exact
set — **45 of 60 failures across 13 modules**; the other 15 are runner
artefacts and were deliberately left alone, because adding the fixture there
would buy a green build by skipping tests that never touch a store.

That attribution had to come from CI. My own local full-tree run reported the
same 60 failures, but under `--tb=line` the FAILED lines carry no reason, so
locally the two groups were indistinguishable.

The scripted edit then introduced one bug of its own — it inserted the fixture
as the *first* parameter, which on two class-based tests put it ahead of
`self`, and pytest reported `fixture 'self' not found` as an **error**, easy to
skim past next to "2261 passed". Fixed by hand, with the package re-run
afterwards specifically to watch the error count reach zero.

Final local verification against a real PostgreSQL 18 — `_lifecycle` +
`_state` + `develop` + `smoke`: **3846 passed, 6 skipped, 1 failed**. The one
failure is `test_audit_all_clean`, a >90s timeout in the audit CLI inside this
container, carrying no store error and untouched by this branch.


ywatanabe1989:
Merged `origin/develop` into this branch (963af9ea) to clear a conflict I caused.

This was green at 07:15 and then went `DIRTY` — not from anything you did. Three migrations landed on develop underneath it in the following half hour (#1244 dispatches, #1241 relocation trio, #1245 node_comms_policy) and I merged all three, so unblocking this is mine to do. Merged rather than rebased, since a rebase would need a force-push.

**Every one of the 22 conflicts was the same shape**: this branch added the `pg_schema` fixture to the `agent_start` tests for `a2a_ports`, and develop added it to the same tests for `node_comms_policy`. The two sides differ only in the `: str` annotation and in parameter **order** — fixtures resolve by name, not position, so both spellings are the same test. Resolved to develop's form.

I did not hand-pick a side. A script compared the two halves of each hunk as a **set** of parameter tokens and auto-resolved only where the sets were equal, leaving anything genuinely different in place: 22 hunks, 22 equal, 0 left over. So no behavioural difference was silently dropped.

One thing I deliberately did **not** change: ruff offered an isort fix on `test__start_verdict.py` and I reverted it, because `origin/develop`'s own copy fails that same default-config check. It is pre-existing and not this merge's business.

Your 4 commits are untouched; this is a merge commit on top. CI is re-running.

ywatanabe1989:
⛔ **Do not merge yet** — I merged develop in to clear the conflict, so this is now mergeable, but it should not be merged. This branch's own adversarial review returned NOT SOUND, and I independently reproduced the central claim.

**The design claim is false.** `port_allocator_store.py:47-50` says keying the store on `port` makes `UNIQUE(port)` hold "STRUCTURALLY: one record per port, by construction, with no secondary constraint to remember", and :282 says `put(..., NEW_RECORD)` means "either our record lands or it does not, with no window".

I read `Store.put` (`scitex_dev/store/_store.py:212-247`). It is:

```python
with self._lock:                      # threading.RLock -- IN-PROCESS ONLY
    current = self._read(record, include_hidden=True)              # SELECT
    check_revision(record, current, self._revision(record), expected_revision)   # in PYTHON
    entry = self._append(record, OpKind.UPSERT, {...}, actor)      # INSERT
```

Read, then compare in Python, then write. `self._lock` is a `threading.RLock`, so it serialises threads inside one process and nothing else. Port claims race across **processes and hosts** — that is the whole point of the ledger — so the lock does not apply and there is a window between the SELECT and the INSERT.

**And the failure is silent, which is a regression.** Because identity *is* the port, two agents that both pass the check do not collide — the second UPSERTs the first under the same record key. One record, two agents each believing they hold port N. The SQLite version this replaces failed **loudly** with `sqlite3.IntegrityError: UNIQUE constraint failed: a2a_ports.port`, which is how the v0.21.19 incident was found at all. Trading a loud failure for a silent double-claim on the table that answers "where do I POST to reach agent X?" is worse than not migrating.

The review also reports that the test which would have falsified this was rewritten into one that cannot, and that the justification for not investigating cites a lock released before the code it supposedly protects.

**None of this is an argument against the migration** — a2a_ports should leave SQLite. It needs a claim path whose mutual exclusion is enforced by the database rather than by a Python-side comparison. Full review is in the workflow result; happy to help once there is a mechanism to review.

Everything else here looks good: the AST count is honest (52→47, all five in `port_allocator.py`), the migration script ships, and the tombstone/`unhide` branch has a real red control.

ywatanabe1989:
## Measured at source: why the race is real, and the claim protocol that works anyway

I re-read `scitex_dev.store` internals (0.56.8) to turn my earlier objection into something actionable. Three facts, then the design.

**1. Identity IS the primary key — but that does not protect a claim.**
`record_key()` (`_guards.py:76`) joins the identity-field values into `_record`, and the rows table declares `_record TEXT PRIMARY KEY`. So with `port` as the identity field, two claims of one port name the same row.

**2. The rows write is an UPSERT, so the PK never fires.**
`Store.put` appends an oplog entry (`_store_op` INSERTs into the **oplog** only; the unique-violation retry at `_store.py:448` handles origin/seq collisions, nothing else) and the row materializes through `_apply._apply_upsert` — per-field HLC merge. Two processes racing NEW_RECORD on the same port: both read no-row, both pass `check_revision` (Python-side, under a `threading.RLock` that no other process can see), both apply. The second **merges silently**. `RevisionMismatchError` only catches the sequential case. This is by design — the store is convergent, not exclusive — and it is why keying on `port` alone cannot make `UNIQUE(port)` structural.

**3. IMMUTABLE is the loud path the schema already owns.**
`MergeRule.IMMUTABLE` resolves a differing concurrent value by reporting a `MergeConflict (kept/rejected/reason)` instead of quietly picking one — the same mechanism `_store_plugin.py` already relies on for `sac_comms_nodes.registered_at` collisions.

### The claim protocol

```
schema: port = identity; claimed_by = IMMUTABLE, required; claimed_at = IMMUTABLE
claim(port):
    put({port, claimed_by=me, claimed_at=now}, expected_revision=NEW_RECORD)
      -> RevisionMismatchError        # sequential loser: port already claimed, pick another
    row = read(port)                   # THE VERIFIER — never trust the put alone
    if row.claimed_by != me:           # concurrent loser: merge kept the first claimant
        pick another port
release(port): hide()                  # tombstone, auditable; re-claim = unhide + verify
```

Put-then-read-back makes the concurrent loser *learn* it lost, deterministically, from the same store both writers converge on. No unique constraint needed, no hand-rolled SQL, no advisory locks — the primitive as-is.

Two requirements on the implementation for this to hold:
- The read-back is **not optional**. A put that "succeeded" proves nothing under merge semantics.
- `claimed_by` must be IMMUTABLE, not LWW — LWW would let the *later* claimant win the merge and the read-back would then lie to the first.

With this, I withdraw my block on the *approach* (the DB does provide the guarantee — through convergence plus IMMUTABLE, not through UNIQUE) while keeping the objection to the earlier draft's Python-side comparison as the sole guard. If the author of this PR wants, I can push the protocol as a commit; a2a_ports is the last SQLite table outside `state_db_schema.py` and I would like it to move soon either way.

ywatanabe1989:
The settled claim protocol (comment 5451759350) is implemented in #1253, which is this branch (head `963af9ea`) plus the protocol commit on top — your migration and its commits are contained, not redone.

One measured correction to the protocol as written: `claimed_by = IMMUTABLE` cannot hold at scitex-dev 0.56.8 source. `_merge.merge_field` keeps the first-stamped value on **every** later differing write, sequential included (stamps are never consulted — "First value wins forever"), and the losing write is only *reported* in `PutResult.conflicts`, nothing raises. Verified against a real store: after `hide` + `unhide`, a re-claim by a different agent reads back the FIRST claimant's name forever — so under IMMUTABLE, `claim -> release -> claim-by-someone-else` can never succeed, and every released port is burned to its first holder. That collides with this PR's own title requirement, which #1253 pins with an explicit claim-by-someone-else test (red under IMMUTABLE).

So #1253 keeps LWW for the lease fields and carries the loudness in the protocol's other half, made unconditional: the **mandatory read-back** after every put — which works under any merge rule and is what the two-process fork race in `test_port_allocator_two_process_race.py` exercises (exactly one winner; the loser's diagnostic names the winner it read back). Full measurement and test numbers are in #1253's description.

ywatanabe1989:
**Rebase hazard — this branch predates two fixes that landed in the same files tonight (2026-09-07).** Flagging before anyone resolves the conflicts, because the obvious resolution silently reverts a live fix.

This PR was last updated `2026-08-28`. Since then, on `develop`:

| commit | PR | file this PR also touches |
|---|---|---|
| `6a2149ab` | #1328 | `src/scitex_agent_container/_state/port_allocator.py` |
| `f54c9c76` | #1329 | `tests/develop/test_delivery_claims_name_their_sink.py` |

**#1328 is the one that matters.** It added `port_is_bindable()` and made `claim_port()`'s auto-scan skip any candidate that cannot be bound *right now*, whatever the claim table says. That fix exists because of a measured misdelivery on compute-03: `agents stop` released the CLAIM while leaving the bridge listening, a second agent was handed the same port with `a2a.port: auto`, and a turn dispatched to one agent arrived **verbatim in the other agent's pane** — with `{"ok": true, "http_status": 200}` returned to the sender. A task that edited files would have edited the wrong tree, authored as the wrong agent.

This PR rewrites the ledger that `claim_port()` reads. Resolving the conflict in `port_allocator.py` by taking this branch's side drops the socket probe and reopens that hole — and it would look clean, because the claim-table logic this PR rewrites is *correct on its own terms*. The probe is deliberately additional to it: **the claim table is not the port.**

Concretely, after rebasing, the auto-allocation loop must still refuse an unbindable candidate:

```python
for candidate in range(lo, hi + 1):
    if candidate in held:
        continue
    if not port_is_bindable(candidate):   # <- must survive the rebase
        continue
    if try_claim(...):
        return candidate
```

and `tests/scitex_agent_container/_state/test_port_allocator_socket_probe.py` (new on develop, real sockets, no mocks) must still pass. If it is deleted or skipped by the rebase, that is the signal the probe went with it.

**#1329's overlap is mechanical**, not semantic: `FROZEN_UNNAMED_CLAIMS` is keyed by `FILE:LINE`, and an insertion in `build_cmds.py` moved one entry `:316 → :368`. Take develop's side there; the claim itself is byte-identical and was neither fixed nor re-reviewed.

Separately, and not for me to decide here: this overlaps #1206 ("[HOLD — blocked upstream] feat(state): port the a2a port ledger to PostgreSQL"). Two open PRs moving the same ledger, one held and one dirty, is worth collapsing to one before either is rebased across 34 files.

*Not rebasing this myself right now — noting the hazard so the fix cannot be reverted by an unremarkable conflict resolution.*

PR #1248, opened 2026-08-28
fix(cli): give the sac process the SCITEX_STORE_DSN it hands to containers

ywatanabe1989:
## Symptom

`sac agents restart <name>` on scitex-compute-04 (2026-08-28):

```
Error: Cannot connect to Postgres store 'node_comms_policy' at postgres: connection
failed: connection to server on socket "/home/ywatanabe/.scitex/pg/run/.s.PGSQL.55432"
failed: fe_sendauth: no password supplied.
```

## Root cause (measured, not inferred)

`_start.py` → `persist_acl_policy` → `open_policy_store` → `scitex_dev.store.host_store()`, which resolves `SCITEX_STORE_DSN` if set, else the local UNIX-socket DSN.

- The fleet default `SCITEX_STORE_DSN=postgresql://scitex-primary:55432/scitex` (`_fleet_env.FLEET_DEFAULT_ENV`) was only ever rendered into `apptainer --env` flags. **The host-side `sac` process itself never received it**, so the resolver took the socket fallback.
- `pg_hba.conf`: `local all all scram-sha-256` → the socket demands a password.
- `~/.pgpass` is keyed by `127.0.0.1` / `localhost` / `scitex-primary`. libpq matches on the literal host string, which for a non-default socket dir is the directory path → no row → `fe_sendauth: no password supplied`.

Reproduced with psql: socket DSN fails identically; `localhost:55432` as the same role connects.

## Why not a `.pgpass` row for the socket dir

The local `:55432` cluster is a streaming **standby** (`pg_is_in_recovery() = t`; `scitex-primary` reports `f`). A pgpass row would only trade a loud connect refusal for silent read-only write failures — the exact "connects fine, cannot write" failure `_fleet_env.py` already documents.

## Fix

`apply_fleet_defaults_to_process()` in `_fleet_env.py` applies `declared_fleet_defaults()` (sac's constant + the operator's `config.yaml` `spec.fleet_default_env`) to `os.environ` at `cli_entry_point`, **only for keys not already present**. A DSN exported in the operator's shell still wins — the same spec.env-over-fleet precedence the module already has. One source of truth; container behaviour unchanged.

## Verification

- Two-arm live check from the worktree source with `SCITEX_STORE_DSN` unset:
  - without the call: locator `postgres[host=/home/ywatanabe/.scitex/pg/run …]`, `open_policy_store()` raises the reported `StoreTargetError`
  - with the call: locator `postgres[host=scitex-primary db=scitex port=55432]`, `open_policy_store()` connects
- `sac agents restart -y scitex-agent-container` with the DSN present: exit 0, new run verified.
- `tests/…/runtimes/test__fleet_env.py` + `test__pg_identity_env.py`: 58 passed (worktree source on PYTHONPATH). Seven new single-assert tests cover both arms, the config.yaml layer, and idempotence against the real `os.environ`.
- Linter: zero errors on the three files; remaining warnings are pre-existing lines.

## Operator workaround until merged

```bash
SCITEX_STORE_DSN=postgresql://scitex-primary:55432/scitex sac agents restart -y <name>
```

### DIFF
--- a/src/scitex_agent_container/cli_pkg/_main.py
+++ b/src/scitex_agent_container/cli_pkg/_main.py
@@ -432,6 +432,16 @@ def cli_entry_point() -> None:
 
     warn_once()
 
+    # The fleet defaults this process hands to every container, applied to
+    # ITSELF — sac opens the same Postgres stores its agents do, and until
+    # 2026-08-28 it was the one process in the fleet launching without
+    # ``SCITEX_STORE_DSN``. A variable exported in the operator's shell still
+    # wins; see ``apply_fleet_defaults_to_process`` for the failure this
+    # closes.
+    from ..runtimes._fleet_env import apply_fleet_defaults_to_process
+
+    apply_fleet_defaults_to_process()
+
     # Cheap pre-scan: don't import the heavy ``host_group`` module
     # unless ``--on`` is actually present on the command line. Importing
     # ``host_group`` (state.host_config + subprocess + the full click
--- a/src/scitex_agent_container/runtimes/_fleet_env.py
+++ b/src/scitex_agent_container/runtimes/_fleet_env.py
@@ -51,7 +51,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import Any, Mapping
+from typing import Any, Mapping, MutableMapping
 
 logger = logging.getLogger(__name__)
 
@@ -378,9 +378,58 @@ def fleet_env_flags(
     ]
 
 
+def apply_fleet_defaults_to_process(
+    environ: "MutableMapping[str, str] | None" = None,
+    *,
+    config_path: Path | None = None,
+) -> dict[str, str]:
+    """Give sac's OWN process the fleet defaults its containers get.
+
+    Everything above renders the defaults into an agent's ``apptainer --env``
+    flags and stops there. But the host-side ``sac`` process opens the SAME
+    stores the containers do — ``persist_acl_policy`` on every start,
+    instances, dispatches, the diary — and it received none of them. On a
+    shell with no ``SCITEX_STORE_DSN``, ``scitex_dev.store.host_store()`` fell
+    through to its UNIX-socket fallback, ``pg_hba`` asked the socket for a
+    password, and ``.pgpass`` — keyed by ``127.0.0.1`` / ``localhost`` /
+    ``scitex-primary``, never by a socket directory — had nothing to offer:
+
+        Cannot connect to Postgres store 'node_comms_policy' ...
+        socket "~/.scitex/pg/run/.s.PGSQL.55432" failed:
+        fe_sendauth: no password supplied
+
+    MEASURED on compute-04, 2026-08-28, on ``sac agents restart``. A
+    ``.pgpass`` row for the socket directory is the WRONG fix: that socket is
+    the local cluster, which is now a streaming STANDBY
+    (``pg_is_in_recovery() = t``), so the row would only trade a loud connect
+    refusal for a quiet read-only write failure. The right store is the one
+    :data:`FLEET_DEFAULT_ENV` already names — the gap was that only the
+    containers were told.
+
+    Same precedence as everything else in this module: a key already present
+    in ``environ`` wins, untouched — a default exists in order to be
+    overridden, and an operator who exported ``SCITEX_STORE_DSN`` in their
+    shell has overridden it. ``environ`` defaults to ``os.environ`` and is an
+    injection seam for tests. Returns the keys it actually set.
+    """
+    import os
+
+    target = os.environ if environ is None else environ
+    injected: dict[str, str] = {}
+    for key, val in declared_fleet_defaults(config_path).items():
+        if key in target:
+            continue
+        target[key] = val
+        injected[key] = val
+    if injected:
+        logger.debug("fleet_env: process env gained %s", sorted(injected))
+    return injected
+
+
 __all__ = [
     "CONFIG_SECTION",
     "FLEET_DEFAULT_ENV",
+    "apply_fleet_defaults_to_process",
     "declared_fleet_defaults",
     "effective_env",
     "fleet_env_flags",
--- a/tests/scitex_agent_container/runtimes/test__fleet_env.py
+++ b/tests/scitex_agent_container/runtimes/test__fleet_env.py
@@ -23,6 +23,7 @@ import pytest
 from scitex_agent_container.runtimes._fleet_env import (
     CONFIG_SECTION,
     FLEET_DEFAULT_ENV,
+    apply_fleet_defaults_to_process,
     declared_fleet_defaults,
     effective_env,
     fleet_env_flags,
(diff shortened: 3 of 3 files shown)


ywatanabe1989:
Additional evidence the fix belongs at `cli_entry_point` rather than at `persist_acl_policy` alone — `sac peer post-turn` from the same shell (no `SCITEX_STORE_DSN`) hits the identical fallback on a different store:

```
dispatch-ledger record failed: Cannot connect to Postgres store 'dispatches' at
postgres[host=/home/ywatanabe/.scitex/pg/run db=scitex port=55432]: ... fe_sendauth: no password supplied.
dispatch-ledger status update failed: (same)
```

Every host-side verb that opens a store was affected; one injection point at the entry covers them all.

ywatanabe1989:
Reviewed. **The root cause is right and I reached it independently** — host-side `sac` never got `SCITEX_STORE_DSN`, `host_store()` fell back to the local socket, and that socket is the read-only standby. Agreed on all of that.

**One objection, and it is load-bearing: the DSN you inject has no role, and that cannot authenticate host-side.** Measured on scitex-compute-04 just now, from an ssh login shell:

```
PGUSER=[UNSET]  PGSERVICE=[UNSET]  whoami=ywatanabe

postgresql://scitex-primary:55432/scitex
  -> FAIL StoreTargetError: Cannot connect to Postgres store 'node_comms_policy'
postgresql://ywatanabe__scitex-agent-container@scitex-primary:55432/scitex
  -> OK rows=50
```

The reason is in `.pgpass`. It has rows for `scitex-primary:55432:*:ywatanabe__scitex-agent-container`, `ywatanabe__cli`, `svc_notifyd`, `svc_gui`, `svc_cards_sync`, `svc_alert` — and **no row whose user is the bare OS user `ywatanabe`** (I checked all 522 lines; every one is a service role or a per-agent role). A roleless DSN connects as the OS user, finds no credential, and fails.

So `FLEET_DEFAULT_ENV["SCITEX_STORE_DSN"] = "postgresql://scitex-primary:55432/scitex"` works **inside a container**, where the per-agent role is supplied, and does not work in the host process this PR is fixing. The same applies to the workaround in your message — `SCITEX_STORE_DSN=postgresql://scitex-primary:55432/scitex sac agents restart -y <name>` is exactly the roleless form, and it is what I measured failing.

**Suggested fix, verified working:** inject a role-qualified DSN for the host process. `ywatanabe__cli` already exists and has a `.pgpass` row, and it looks purpose-built for this:

```
postgresql://ywatanabe__cli@scitex-primary:55432/scitex
  -> READ OK rows=50
  -> WRITE OK        (the put that agent_start actually needs)
```

Please confirm your 58 tests and the live two-arm check would still pass with the roleless DSN — I suspect the live arm passed because the shell that ran it had a role in scope, which is the vantage trap here.

**Second issue, separate from this PR but on the same path.** Fixing the connection does not finish the restart. `Store.__init__` runs `_apply_additive_migrations` on every open, which issues DDL, and DDL requires table OWNERSHIP. After connecting successfully you hit:

```
Error: must be owner of table node_comms_policy_oplog
```

40 store tables are owned by per-agent roles instead of `scitex_owner` (created by whichever agent first opened each store during the migration). It is bidirectional — business's start also failed to write `incarnations_oplog` and `cct_token_owner_oplog`, which `scitex_owner` owns and no agent role does. Tracked on card `store-tables-owned-by-agent-roles-block-agent-start-20260828`, with the ALTER script at `/home/ywatanabe/fix-store-table-ownership.sql`. Worth knowing so this PR is not reported as fully restoring restarts.

Everything else looks good — absent-keys-only with the exported DSN winning is the right precedence, and the symptom write-up is accurate.

ywatanabe1989:
**Correction and strengthening of my earlier comment — your diagnosis is more right than I first credited, and one of my claims needs narrowing.**

I said the "must be owner of table node_comms_policy_oplog" failure was a *separate* problem from this PR, about table ownership on the primary. That framing was wrong. Measured on **scitex-compute-01** just now:

```
# what sac actually resolves with no env:
host_store(pkg=..., name='node_comms_policy').locator
  -> postgres[host=/home/ywatanabe/.scitex/pg/run db=scitex port=55432]   # LOCAL socket

# the same call, with a role-qualified primary DSN:
SCITEX_STORE_DSN=postgresql://ywatanabe__cli@scitex-primary:55432/scitex
  record_comms_policy(name=..., ...)  -> OK
```

So the ownership error is a **symptom of resolving the wrong database**, not of ownership on the primary. Host-side sac lands on that host's *local* PostgreSQL, whose tables have their own owners. **This PR is the fix for it**, which I did not say clearly enough the first time.

**My roleless objection stands, and now matters more**, because it decides whether the fix lands or just changes the error. On scitex-compute-04, from an ssh login shell:

-  (your  value) -> **FAIL, cannot connect**
-  -> **READ OK, WRITE OK**

 has 522 rows and not one has the bare OS user  as its user field; a roleless DSN connects as that user and finds no credential. Please inject a role-qualified DSN.

**One more thing to check on your side, which I could not explain.** Setting the env externally does **not** rescue the restart:

```
SCITEX_STORE_DSN=postgresql://ywatanabe__cli@scitex-primary:55432/scitex sac agents restart business --yes
  -> Error: must be owner of table node_comms_policy_oplog
```

…while a direct  in the same shell, same env, on the same host, **succeeds**. So something in the restart path is not seeing that variable — a re-exec, or a subprocess whose environment is rebuilt. If that is so, the workaround in your message cannot work for anyone, and it is worth confirming your live two-arm check was not measuring a shell that already had a role in scope.

Practical note for whoever tests this:  stops before it starts, so a failing start **leaves the agent down**. It took business down twice for me today. Prefer a stopped canary over a live agent.

ywatanabe1989:
⚠️ **Still blocking on c8716ec3 — the pytest gate is green but the fix does not remove the error it targets.** Line 211 is unchanged and still roleless:

```python
"SCITEX_STORE_DSN": "postgresql://scitex-primary:55432/scitex",
```

Ran exactly that value host-side on scitex-compute-04, and it produces **the same `fe_sendauth` your PR description quotes as the symptom** — only against the primary rather than the socket:

```
StoreTargetError: Cannot connect to Postgres store 'node_comms_policy' at
postgres[host=scitex-primary db=scitex port=55432]: connection failed:
connection to server at "100.64.0.5", port 55432 failed:
fe_sendauth: no password supplied.
```

Compare, same host, same shell, same moment:

```
postgresql://ywatanabe__cli@scitex-primary:55432/scitex   -> READ OK, WRITE OK
```

So this moves the failure from "socket, no password" to "primary, no password". The host-side process still has no usable identity, which is the thing the PR set out to give it.

**Why the tests do not catch it.** `fe_sendauth` is a *credential-resolution* failure, not a DSN-shape one. `.pgpass` is matched on (host, port, db, **user**); a DSN with no user connects as the OS user `ywatanabe`, and none of that file's 522 rows names that user. A unit test asserting the env var is now present in the process passes regardless — it never opens a connection as the OS user against the primary. That is also the most likely reason the live two-arm check looked green: any shell that already had a role in scope would succeed.

**Minimal change:** make the fleet default role-qualified, e.g. `postgresql://ywatanabe__cli@scitex-primary:55432/scitex`. `ywatanabe__cli` already exists and already has `.pgpass` rows on the hosts I checked, and it both reads and writes the stores `agent_start` touches.

If you would rather keep `FLEET_DEFAULT_ENV` roleless because containers resolve their own per-agent role from it, then the host-side value needs to be a *separate* constant rather than the same one — the two contexts genuinely need different identities, which is worth stating explicitly in the module docstring so the next reader does not re-unify them.

Happy to be wrong here — if you have a host where the roleless form authenticates, tell me which and I will re-measure. My check was a login shell on compute-04 with `PGUSER` and `PGSERVICE` both unset.

ywatanabe1989:
## Correcting my own objection — do NOT put a role in the DSN

My earlier comment asked for a role-qualified `SCITEX_STORE_DSN`. **That advice was
wrong and I am withdrawing it.** Implementing it would have broken per-agent
attribution across the whole fleet. The blocker itself is real; the remedy I named
was not. Details, measured rather than argued:

### Why the DSN is roleless on purpose

From inside a running agent container:

```
DSN     : postgresql://scitex-primary:55432/scitex     <- roleless
PGUSER  : ywatanabe__scitex-agent-container            <- the role lives HERE
CONNECT OK as: ('ywatanabe__scitex-agent-container', 100.64.0.5)
```

`_pg_identity_env.py` injects `PGUSER=<host_user>__<agent>` for every container when
no layer declares one, and `_fleet_env.py:182` says so explicitly: the DSN "names the
host by name and OMITS the user, letting `PGUSER=ywatanabe__<agent>` supply the role".
Confirmed independently: **0 of 132 specs set `SCITEX_STORE_DSN` themselves.**

A role written into `FLEET_DEFAULT_ENV` is explicit in the DSN and therefore takes
precedence over `PGUSER`, so it would collapse 132 distinct per-agent roles into one.
Every store write would attribute to the same identity, and per-agent grants would
stop meaning anything. That is a worse bug than the one this PR fixes.

### What is actually broken

The container path has two halves — roleless DSN **plus** injected `PGUSER`. This PR
gives sac's own host process the first half and not the second:

```
host scitex-compute-04:  PGUSER=[UNSET]   ~/.pgpass exists: yes
```

With `PGUSER` unset libpq falls back to the OS user `ywatanabe`, and `.pgpass` matches
on (host, port, db, **user**) — there is no row for bare `ywatanabe` among its 522.
Hence the `fe_sendauth: no password supplied` this PR set out to eliminate, reproduced
in the previous comment.

### The fix

`apply_fleet_defaults_to_process` should supply `PGUSER` for sac's own process — the
host-side counterpart of `_pg_identity_env`, keeping the DSN untouched and preserving
the declared-anywhere-wins precedence (an existing `PGUSER` must not be overwritten).

`ywatanabe__cli` is the natural value for the host process. Verified to authenticate
on every host that runs agents:

```
scitex-compute-01    OK ywatanabe__cli
scitex-compute-03    OK ywatanabe__cli
scitex-compute-04    OK ywatanabe__cli   (READ OK, WRITE OK)
scitex-nas-03        OK ywatanabe__cli
scitex-compute-02    not tested - no psycopg installed (and no agent specs)
```

I have not pushed to this branch: it is checked out in another session's worktree and
I do not want to collide with work in progress. Happy to send it as a commit if you
would rather not carry it — say the word and it is yours either way.

ywatanabe1989:
Heads-up from another session: #1250 is branched from this PR's head, so it
contains your commit verbatim and supersedes it — merging #1250 merges both, and
this one can be closed once it lands.

What it adds is the second half of the same identity. Your DSN is roleless by
design, and containers pair that with `PGUSER=<host_user>__<agent>` from
`_pg_identity_env`; the host process still had no `PGUSER`, so libpq fell back
to the OS user `ywatanabe`, and `.pgpass` matches on user — compute-04 holds 522
rows and none for the bare OS user, so it still ended in `fe_sendauth: no
password supplied`. #1250 injects `PGUSER=<host_user>__cli` in
`apply_fleet_defaults_to_process` after your cascade (declared-anywhere still
wins), reusing `derive_pg_role`/`PG_USER_ENV` rather than putting a role in the
DSN, which would override all 132 containers' per-agent roles. Control and fix
are both measured on the compute-04 host in the PR body.

https://github.com/scitex-ai/scitex-agent-container/pull/1250

ywatanabe1989:
Closing as contained: #1250 (merged) was based on this branch's head c8716ec3, so every commit here is now in develop with this PR's DSN work intact, plus the host-side PGUSER injection that the review found missing. Nothing here was discarded — the branch's history merged verbatim. Thank you for the root-cause work on the DSN cascade; the two-halves analysis in the review comments has the full story.

PR #1258, opened 2026-08-28
feat(state): move instances to the shared store

ywatanabe1989:
# feat(state): move instances to the shared store

`instances` was the last table in `state.db` that both a writer and a reader
actually reached — 605 rows on compute-04 across nine caller modules. It now
lives in the shared PostgreSQL store via `scitex_dev.store`.

**`init_schema` on a fresh database now creates NOTHING.** Measured on this
branch after merging #1255 / #1256 / #1257:

```
KNOWN_TABLES = ()
CORE_TABLES  = ()
--- sqlite_master after init_schema ---
  (nothing)
```

sac's SQLite schema is empty. 16 tables at the start of 2026-08-28; zero now.

## Row counts, for planning the data migration

Measured read-only against each host's
`~/.scitex/agent-container/runtime/state.db`:

| host | total | live (`ended_at IS NULL`) |
|---|---|---|
| scitex-compute-04 | 605 | 7 |
| scitex-compute-03 | 174 | 1 |
| scitex-nas-03 | 36 | 2 |
| scitex-compute-01 | 15 | 0 |
| scitex-compute-02 | — | no state.db |

**830 rows total, 10 live.**

Within compute-04's 605, the `host` COLUMN spreads across four values —
`scitex-compute-04` 597, `scitex-02` 3, `scitex-compute-03` 3,
`ywata-note-win` 2 — and 5 rows carry `remote=1`. Those cross-host rows are
NOT duplicates of the peers' own rows: the identity is `{id, host}` with a
uuid7 minted at the origin, so the lead's mirror and the peer's own record are
different records by design and the migration must not deduplicate them.

Zero rows anywhere have `a2a_port IS NULL AND bound_port IS NOT NULL`, so the
`COALESCE` fold below changes no existing value — it removes the ability for
the two to diverge in future.

## What moved, and what did not

`_store_plugin.INSTANCES` declared this schema while it was still a plan.
Checking it against the real DDL and against every reader in `src/` found gaps
in **both** directions; both are closed here.

**Three columns dropped** — no reader anywhere, no writer ever set them:
`definition_id` (a FK to `definitions`, a table nothing ever INSERTed into and
which itself left in #1255), `scope` (the literal `'global'` on every row,
alive only inside `idx_instances_active`), `ppid` (a parameter with no call
site).

**Three columns added** — production code reads them: `screen`
(`_restart_verify` takes the tmux handle off an ENDED row), `workdir` (never
read back, but the rename verb REWRITES it — dropping it would have retired
live coverage), `remote` (the authoritative locality flag, deliberately not a
hostname compare; five readers plus the GC branch on it).

**`bound_port` is a FOLD, not a drop.** The DDL had it beside `a2a_port` and
every writer set both from ONE value. Two columns holding one fact is how the
two drift, and they had: `state_db_forward`'s docstring records a live row on
ywata-note-win where the split answered "where do I send this" two different
ways. The store keeps one port field; `instance_as_dict` mirrors it back out
under **both keys**, so the seven readers that prefer `bound_port` are
shape-stable.

## `spawned_by` is now LAST_WRITER_WINS, and the plugin says so

The plugin declared it IMMUTABLE. Measured against the real rename path, that
is **not implementable**: `merge_field` freezes an IMMUTABLE field at its first
stamped value and reports a `MergeConflict` in `PutResult.conflicts` **without
raising**, which no caller inspects. Renaming a parent agent would therefore
leave every child's `spawned_by` on the dead name while the rename reported
success — the exact silent no-op the rename step exists to prevent.

Excluding `spawned_by` from the rename was the alternative and was rejected:
`_lifecycle/_status` reads it, and `sac agents rename` already covered it under
SQLite, so excluding it would retire live coverage inside a storage migration.

IMMUTABLE was never buying replication safety **here**: `instances` is PER_HOST
truth with `host` in the record identity under `SINGLE_WRITER`, so two hosts
cannot write one record at all — the merge rule can only ever decide a
SAME-HOST rewrite, which is exactly this rename. `sac_lineage` is where a
contradicted parent still surfaces as a `MergeConflict`. Same shape, same class
of reason, as `state_db_incarnations`' documented divergence.

## The GC branch that became fleet-wide, and is now scoped

Under per-host SQLite the heartbeat-staleness branch scoped **only** by
`remote=0`, and that was sufficient BY ACCIDENT: each host owned its own file,
so "not a cross-host mirror row" also meant "written here". On the shared store
it means neither — a peer's own local record is `remote=0` too, and every host
can see it.

Left unscoped, one host running `sac db tick` would have tombstoned LIVE agents
fleet-wide. Nothing would have raised: the victims simply stop being returned
by `list_active_instances`, which every reader takes as "not running". The
branch now carries `host == canonical_host` as well and keeps `remote` — they
are different claims (`remote` written by the dispatcher, `host` by the
observer) and losing either would be a silent widening.
`test_state_db_gc_host_scope.py` pins all three branches and the scope.

## Two hazards this port had to measure

1. **An IMMUTABLE field is frozen by its FIRST WRITE, and writing `None`
   counts.** `merge_field` takes the incoming value only while
   `current_stamp is None`. A `put` carrying `ended_at=None` on a live record
   stamps that field at None, and every later tombstone is then a silently
   rejected `MergeConflict` — no exception, just a record that can never be
   ended. Every write path goes through `strip_unset` for this reason, and
   every write is a PARTIAL put rather than a whole-record rewrite.

2. **The `id DESC` tiebreak is DETERMINISM, not RECENCY.** `new_uuid7` falls
   back to `uuid4` on every Python below 3.14, including the 3.12 in the sac
   image. So two starts inside one `started_at` second tie and the tiebreak
   resolves them randomly. This is INHERITED — the SQLite
   `ORDER BY started_at DESC, id DESC` had the same property against the same
   ids — but it is written down now, and one test states it outright instead
   of relying on it.

## Three constants went EMPTY, and each needed a rework rather than a value change

`instances` was the last name in all three, so the code consuming them lost its
referent rather than an entry.

**`KNOWN_TABLES = ()`** — every generic reader (`table_counts` behind
`sac db show`, `export_state` / `import_state`, the `click.Choice` for
`sac db query`) now has nothing to be right about. `--table` rejects every
value, and that refusal is the intended end state: an accepted-but-empty answer
reads as "this agent has no rows" when the truth is "you are asking the wrong
database".

**`CORE_TABLES = ()`** — the health probe's premise ("this file is OURS,
identified by the tables it carries") has no referent. Padding it with a name
nothing creates would be *worse* than emptying it: `any()` over such a name is
false for every store, so every healthy `state.db` would classify `schemaless`
— the probe's own "this is a DIFFERENT database" verdict, manufactured by our
own migration, with `describe()` sending an operator looking for a table that
cannot exist. So `inspect_store` treats an empty core as "no signature to
check" and reports a readable file as `populated`; `absent` and `empty` are
untouched and are the two verdicts that still carry real information. The
`schemaless` remedy clause is kept and tested by constructing the state
directly, so re-adding a table is a one-line change rather than a silent
regression.

**`PATH_COLUMNS = ()`** — `instances.workdir` was its last entry. The path
rewrite is real and moved to the store (`state_db_instances_rename`).

The tests that encoded "there is at least one table" were **reworked, not
deleted**. `test_the_core_is_not_empty` is the positive control that CAUGHT
this — it existed precisely because a membership test over an empty tuple is
vacuously true — and it is inverted rather than dropped: the guard moved from
the constant to the classifier, and the test now asserts an empty core does not
make every SQLite file on the machine classify as ours. Likewise the
export/import round-trip tests assert the empty *wire shape* (`tables == {}`,
distinguishable from `{"instances": []}`), the `--tables` acceptance cases
became refusal cases, and the eight `sac db query` rendering tests collapsed
into two refusal tests with a comment recording what they used to cover and
where the surviving guarantees are exercised.

## One module DELETED, not ported

`_state/state_db_migrations.py` is gone. Its last function,
`migrate_instances_add_family_tree_cols`, ALTERed `instances` to add
`bound_port`/`remote`/`spawned_by` — and that table moved here, so the
migration became a permanent no-op. Its two predecessors had gone the same day
with `instance_heartbeats` and `node_comms_policy`.

What was left was a module of departure notes with **no code and no importer**,
whose `import sqlite3` survived only to keep its `FROZEN_SQLITE` entry honest —
a module kept alive BY the freeze list, which is the opposite of a ratchet. The
notes are carried into `state_db.py`'s docstring (including the one real ruling
among them: the two heartbeat migrations could still FIRE on an old enough
database, re-creating a table the schema had just declared it does not
maintain), and the freeze entry shrinks, which that list is allowed to do.

## Data migration

`scripts/migrate_instances_to_postgres.py`, on `scripts/_migrate_lib.py`.

**Every row moves, ENDED ones included.** `last_known_instance`,
`cli_pkg/lifecycle/_restart_verify` and `_reconcile/_rule` all read an ended row
as evidence, and for the last of those a missing record is a DIFFERENT verdict
(`NEVER_STARTED`), not merely less detail. `started_at` / `ended_at` /
`exit_reason` / `last_heartbeat_at` are carried VERBATIM — never re-stamped.

**The verify had to be replaced.** `_migrate_lib`'s shape compares the source
row count against whatever `verify()` returns, and every earlier consumer let
that be a GLOBAL count of the store. For a SHARED store that cannot hold: once
compute-04 has migrated, compute-03's run reads its own 174 rows and sees 779 in
the store. A check whose verdict depends on run order is not a check. This
script's `_verify` counts the SOURCE IDS IT CARRIED, read back through the
production scan.

## The migration answers three incidents from tonight's sibling migrations

The operator hit all three running #1256 and #1257 by hand; each is guarded
here rather than left to be rediscovered.

**1. The migration CREATES the tables, so they are owned by whoever runs it.**
Running channel_events as `ywatanabe__cli` left the new tables owned by that
role with grants to nobody; every agent connects as `ywatanabe__<agent>`, so the
fleet's message channel failed with `InsufficientPrivilege` three minutes later
and stayed broken ~6 minutes. `_preflight_ownership` now runs BEFORE any write
and distinguishes three cases: already `scitex_store_owner` (proceed); a MEMBER
of it (pin `role=` in the DSN so every CREATE lands owned correctly); neither
(REFUSE, naming the `GRANT` to run). A role that does not EXIST on the server is
its own message with its own remedy, because `pg_has_role` raises on an unknown
role and "not provisioned" is not "not a member". `SAC_MIGRATE_OWNER_ROLE`
overrides the name, and `=''` opts out — loudly, naming the outage.

**2. Verify with the CONSUMER's credential, not the writer's.** The
post-migration check passed while the fleet was broken, because it ran as the
migrating role — the one identity guaranteed to work. `SAC_MIGRATE_VERIFY_ROLE`
re-reads as a named consumer and FAILS the run if it cannot; without it the
script prints a warning saying in as many words that the verification proves
nothing about the fleet.

**3. A cross-host ordering assumption can be structurally unsatisfiable.**
channel_events' guard refused on rows newer than the import, assuming they were
post-cutover writes; 145 such rows for one target turned out to be another
host's legitimately imported history, because two hosts served the SAME targets
over OVERLAPPING periods. **This migration is immune by construction, and that
is stated in `main()` rather than left unexamined:** the record identity is
`{id, host}` with a uuid7 minted at the origin, so two hosts' rows are different
records that never compare, and no rule here reads one host's timestamps against
another's. An interleaved history is simply two sets of records. There is also
no collision check for the same reason — two hosts' rows cannot collide.

## The migration, exercised end to end

Run for real against a throwaway PostgreSQL, on a SQLite fixture holding the
three shapes that matter (a live row with only `bound_port` set, an ENDED row,
and the lead's `remote=1` mirror of an agent on a peer):

```
=== DRY RUN ===
  ENDED 2026-08-19T00:00:00Z  alpha@host-a:8002  id=i-ended
  live  2026-08-20T00:00:00Z  alpha@host-a:8001  id=i-live
  live  2026-08-20T00:00:01Z  beta@peer-b:8100  id=i-remote
DRY RUN — 3 row(s) would move.
=== OWNERSHIP REFUSAL (no scitex_store_owner on this server) ===
  REFUSING: role 'scitex_store_owner' does not exist on this server ...
rc: 1                                    <- refused BEFORE writing
=== COMMIT (owner check opted out) ===
  3 written, 0 already present and left untouched
  verify: 3 record(s) visible through the production reader
  WARNING: the verification above ran as the MIGRATING role ...
=== RE-RUN (idempotence) ===
  0 written, 3 already present and left untouched
=== READ BACK ===
live: [('alpha', 8001, 8001, 0), ('beta', 8100, 8100, 1)]
ended alpha exit_reason: stopped         <- carried verbatim
ended alpha ended_at: 2026-08-19T02:00:00Z
last_known alpha: i-live
dropped keys absent: True                <- no `scope` key re-emitted as None
workdir: /home/u/proj/alpha
screen: sac-alpha
counters: 3 10
```

The `bound_port`-only row (`a2a_port` NULL) comes back as `8001` under BOTH
keys — the fold and the mirror, measured.

## Merges with `origin/develop` — four times, mid-flight

Merged `c428c7bc` (#1255, three dead tables), `8c6ad37a` (#1256, lineage),
`fb5592b0` (#1257, channel_events) and `40779a58` (#1232). Eight files
conflicted on the first, six on the second, ten on the third; all resolved by
hand, keeping every removal from every PR.

The rename/export seeds moved three times as their subject tables left —
`instances` → `lineage` → `channel_events` → nothing. Each move is recorded in
the seed's own comment, because each was a table LEAVING SQLite and a seed
naming a departed table RAISES, which is the loud failure those files' own
docstrings ask for. The `_rename_db` identity/history halves finally had no
SQLite table to seed at all; `_rename` runs the `instances` step against the
store with a key-scoped inverse, covered in
`_state/test_state_db_instances_rename.py`.

## `_store_plugin` was split, and why the refusals went under `_state/`

`_store_plugin.py` was 529 lines against a 512 cap **before** this change, and
the measured `sac_instances` reasoning pushed it further. `NEVER_SYNCED` moved
to `_state/_store_refusals.py` (re-exported, so every import keeps resolving).
It sits under `_state/` rather than beside its parent for a budget reason
stated in its docstring: the package root is capped at 15 flat `.py` files
(PS-108b) and was exactly at it — a sixteenth failed `audit-all`.

https://claude.ai/code/session_019PCMkub2ZBiL4huQUxMDDb


ywatanabe1989:
## Picked up at `7663bbd1`; fixed in `eb4aaf83`

### The failure, measured — not the one the brief predicted

Both matrix legs, 147 errors each, **identical**: 17818 passed / 84 skipped / 147 errors on
py3.11 (scitex-03-org-cpu-01) and py3.13 (scitex-01-org-cpu-01). Same count, same three files,
so this was **not** the runner-hostname collision that has split legs here before — one cause,
both legs.

```
116  tests/scitex_agent_container/_lifecycle/test__rename.py
 23  tests/scitex_agent_container/cli_pkg/lifecycle/test__rename.py
  8  tests/scitex_agent_container/_lifecycle/test__rename_db.py
```

All `sqlite3.OperationalError: no such table: instances`, all from one line:

```
tests/scitex_agent_container/_helpers/fleet_root.py:289: in seed_db_rows
    conn.execute(sql, args)
```

The predicted `test_state_db_export_tables.py` rework was **already done** in this branch
(`KNOWN_TABLES` is empty, and that file's only entry in the run is one SKIP at line 84). So were
`state_db_export`, `state_db_gc`, `_lifecycle/_instances` and `state_db_health.CORE_TABLES`
(now `()`, guarded by `if CORE_TABLES and ...`). The break was entirely in the **test seam**.

### What changed

1. **`_helpers/fleet_root.py`** — `seed_identity_and_history` now writes both halves through
   their real production writers, `record_instance_start` and `persist_event`, instead of
   hand-INSERTing. `seed_db_rows` goes with its last caller: `init_schema` creates no tables, so
   a seeding helper can only raise. `make_state_db` stays and is now the one place these suites
   prove a fresh state.db still opens cleanly while defining nothing.

2. **`_lifecycle/test__rename.py`** — `_db_names(path)` becomes `_carried_names()`, reading the
   identity half through `list_active_instances` (the same accessor `sac agents list`, the start
   preflight and the reconciler use) and the history half through the channel store.

3. **`_rename_plan.build_plan`** — merges `count_instance_rename_rows` into `plan.db_counts`.
   **This is a real bug the test failure was hiding.** Every surviving `NAME_COLUMNS` pair names
   a table `init_schema` no longer creates, so the SQLite half of that count is permanently `{}`
   — `sac agents rename --dry-run` was about to print `0 column(s)` for an agent with hundreds
   of recorded lifetimes. `count_instance_rename_rows`'s own docstring already promised to
   report under the same `table.column` keys; nothing was calling it. An unreachable store now
   appends a stated warning instead of contributing a silent zero.

4. **`_lifecycle/test__rename_db.py`** — rewritten to measure what the module can now do rather
   than what it used to. Hand-creating the departed tables to keep the old assertions running
   was considered and rejected in the docstring: it would measure a schema the test file wrote.
   A positive control (`test_the_module_still_declares_pairs_to_look_for`) keeps an emptiness
   that came from an empty constant from passing as one that came from an empty database.

### On `_rename_db` — a documented limitation, stated rather than left to be rediscovered

After this PR, **`_rename_db` reaches nothing on any state.db sac creates.** `PATH_COLUMNS` is
empty and `NAME_COLUMNS`' only survivors are `("comms_grants", "sender_name")` and
`("comms_grants", "target_name")` — a table whose DDL was already deleted (it is
`sac_comms_grants` in the store now). `rename_rows` skips absent tables, so the `state-db` step
is a correct no-op rather than a silent wrong answer, and the rows it used to carry are carried
by `rename_instance_rows` as its own step with a key-scoped inverse.

Deleting the module and its step is the honest follow-up and is deliberately **not** in this PR:
it touches `STEPS` (which the rollback matrix parametrises over), `RenamePlan.db_counts`, the CLI
dry-run output and `FROZEN_SQLITE`. Worth its own change with its own argument.

Also stale and not edited: `_rename.py`'s docstring still says "16 name columns + 2 path
columns". `_rename.py` is 572 lines, over the 512-line cap, so the pre-edit hook refuses any
edit to it until it is split — flagging rather than triggering a module split inside this PR.

### Verification

* **Targeted:** 147 collected / **147 passed** / 0 skipped — exactly the 147 that errored in CI.
* Run against a **throwaway PostgreSQL provisioned the way `exec-in-sif.sh` provisions CI's**
  (its own `initdb`, `system_identifier` 7679233114159680002, `pg_is_in_recovery()=false`), so
  the `pg_schema` tests **execute** rather than skip. Loopback on a fleet host is a read-only
  replica; without this every rename test would have skipped and looked identical to passing.
* **Fresh-DB listing through sac's own `init_schema`: `sqlite_master` = 0 entries, 0 user
  tables.** Zero `CREATE TABLE` remain.
* `ruff check` clean on all five changed files.

PR #1274, opened 2026-08-30
refactor(sqlite-out): delete the storage engine, the one-shot carriers, and the allowlists

ywatanabe1989:
Stage F2, and the end of the SQLite footprint in `sac`.

Matching lines outside `docs/adr/`: **1011 → 160**, measured on the pushed
commit against the merge-base `afe3129f`. `import sqlite3` across
`src/` + `tests/` + `scripts/`: **1 → 0**.

## The engine

`_state/state_db.py` kept a connection factory, `init_schema`, `open_db`,
`table_counts` and the `KNOWN_TABLES` whitelist every generic reader walked;
`_state/state_db_schema.py` held the DDL. Both are deleted.

The shape they had reached is the argument for deleting rather than keeping
them. `KNOWN_TABLES` was already an EMPTY tuple and both schema constants pure
SQL comments, so `init_schema` issued no DDL, `open_db` handed out a connection
to a database with no tables, and `table_counts` could only return `{}`.
**Nothing in `src/` called any of the three.** What remained created an empty
file on disk and answered every question with a plausible zero — the exact
success-shaped wrong answer each departing table was removed to avoid, one
level up, in the accessor that was supposed to be the safe one.

`state_db.py` is now a namespace: `now_iso`, `new_uuid7`, and the accessor
re-exports.

## The dead pin

`_lifecycle/_instances.py` threaded a `db_path` from `make_restart_callback`
through `restart_and_record` to `record_local_instance`, where nothing read it.
It was added 2026-07-14 for a real incident: the health monitor runs on a
background daemon thread, and without a pin its writes followed a mutable
process-global into another test's database under pytest and into the live
fleet file on a real host — both observed. Every write on that path now
addresses the store through `SCITEX_STORE_DSN`, so the pin had stopped
selecting anything. A pin that pins nothing reads as an isolation guarantee
that is not being made.

## The one-shot carriers

Fourteen `scripts/migrate_*_to_postgres.py`, five shared helpers, nine tests —
deleted. A one-shot migration does not earn permanent tooling.

**No carrier logged its own run**, so completion was established externally:

* The source files are gone fleet-wide. A manifest of 33 `state.db` files was
  taken at `~/.scitex/backups/sqlite-eradication-20260829/manifest.tsv` on
  2026-08-29 before deletion. Probed six hosts: compute-01/-02/-03/-04 have
  none; nas-03 and ywata-note-win keep a file with **0 rows in every legacy
  table**.
* Each destination store holds rows stamped **before** its cutover — history
  only a migration could have written: `instances_rows` 887 with `started_at`
  back to 2026-06-01, `incarnations_rows` 488 from 2026-08-14,
  `dispatches_rows` 3159 from 2026-08-09, `comms_grants_rows` 50,
  `comms_nodes_rows` 53, `lineage_rows` 20, `verdict_delivered_rows` 779,
  `sac_channel_events` 9180.
* `sac_channel_import` keeps **53 import records** naming source host, source
  path and id window, across three hosts.

### One is recorded as a loss, not a completion

`migrate_diary_to_postgres.py` measured **32 heartbeats** on compute-04 the
night of the port; the destination holds **1**, on an append-only store where
every tick is its own record. Those rows were not carried, and the source is
now destroyed. The script is deleted because it has nothing left to carry on
any host — **not** because it finished.

## The gate keeps its scans and loses its allowlists

`tests/develop/test_sqlite_footprint_frozen.py` carried four frozen
inventories (`FROZEN_SQLITE`, `_SCRIPTS`, `_TESTS`, `_DDL`). All four
populations are empty, and their staleness tests were vacuous with them — a
test that cannot fail is worse than no test, because the file still lists it.

The predicate *"every module importing sqlite3 appears in FROZEN_SQLITE"*
collapses, once the set is empty, into *"nothing imports sqlite3"* — the rule
the operator actually asked for, with no list between the reader and it.

**Kept:** the import scan, the `CREATE TABLE` scan (with `POSTGRES_DDL` for
legitimate PostgreSQL definers and `SCAN_EXEMPT` for prose), the VENDOR scan,
and every planted-file positive control. The vendor scan is the one that must
outlive the others: `SQLiteSession` is the shape this comes back in once nobody
writes the import, and no import check can see it. 11 tests, all passing.

## Absence tests

Tests asserting a table was NOT in the local database are deleted rather than
re-pointed: an absence test needs somewhere its subject could still be, and
there is none. Their reasoning is preserved as departure notes where it was
specific to the table — an empty `lineage` reads as "every agent is a ROOT",
and a root may spawn.

## What is deliberately left

160 lines remain outside `docs/adr/`, in five categories, none of which is sac
choosing a storage engine:

| lines | what |
|---|---|
| 82 | the gate file itself — the scan patterns, and the operator's ruling quoted verbatim |
| 21 | `CHANGELOG.md` — historical entries plus this one. Rewriting past entries would be dishonest |
| 11 | the card id `sqlite-out-per-call-connect-cost-20260828`, cited for a live perf measurement. Mangling an external identifier breaks the lookup that makes the number checkable |
| ~20 | other packages' storage: `scitex-cards`' shadow and its retired `SCITEX_CARDS_READ_BACKEND`, `scitex-db`'s test suite (an exempt database *library*) |
| ~26 | residual test prose, reducible but low-value; the operator's Japanese ruling in `config/_group_authority.py` is among it and is worth keeping verbatim |

## Verification

* **Baseline control:** `git grep -i sqlite afe3129f -- . ':(exclude)docs/adr'`
  returns exactly **1011**, and exactly one `import sqlite3` in `src/`
  (`state_db.py`) — both matching the brief.
* **Ruff control:** 34 findings (`F401,F811,F821,E9`) on this branch and
  **34 on the merge-base tree**, in the same files — none of which this PR
  touches. No new lint.
* **Targeted suite** (`_state/` + `_lifecycle/` + `test__store_plugin.py`,
  the areas this PR changes): **2807 passed, 957 skipped**.
* **Full suite:** see the comment below — the PostgreSQL-backed tests cannot
  run from this container.


ywatanabe1989:
## Verification, and what this container could not verify

### Counts, on the pushed commit `5a60ca64`, with a positive control

```
$ git grep -i -I 'sqlite' afe3129f -- . ':(exclude)docs/adr' | wc -l
1011                      # baseline, matching the brief exactly
$ git grep -i -I 'sqlite' 5a60ca64 -- . ':(exclude)docs/adr' | wc -l
160
```

`import sqlite3` / `from sqlite3` across `src/` + `tests/` + `scripts/`:
**1 → 0**. The one at the baseline was `_state/state_db.py`, as the brief said.

The control is load-bearing here: the same scan on the parent returns the
figure the work was scoped against, so the "after" number is measured with an
instrument that demonstrably reports the "before".

### Lint

```
ruff check --select F401,F811,F821,E9  →  34 findings
```
on this branch, **and 34 on the merge-base tree** extracted with `git archive`,
in the same files — `_freshness/__init__.py`, `_state/snapshot/__init__.py`,
`config/_parsers/__init__.py`, `_state/state_db_comms_nodes.py` (TYPE_CHECKING
`Store`), `cli_pkg/test__dev_jobs_backend.py`. This PR touches none of their
code. **No new lint.** CI's own `ruff-on-ubuntu-latest` gate is green.

Nine findings this PR *did* introduce (unused imports left behind by deleted
tests, plus a `Path` import orphaned by the `db_path` removal) were fixed
before the commit.

### Tests

Post-commit, on exactly the files this PR edits:

```
49 passed, 157 skipped in 5.21s     (exit 0)
```

The retargeted gate on its own, scanning the whole tree:

```
tests/develop/test_sqlite_footprint_frozen.py ...........  11 passed
```

Earlier in the change, the full `_state/` + `_lifecycle/` + `test__store_plugin`
slice ran **2807 passed, 957 skipped, 0 failures**.

### What this container could NOT verify, stated plainly

**The 157 skips above are the PostgreSQL-backed tests, and they cannot run
here.** `tests/_store_isolation.py` refuses two targets by design, and both are
what this host offers:

* every loopback `:55432` is a **read-only replica** of the fleet cluster;
* `scitex-primary:55432` is the **fleet PRIMARY**, and the fixture hard-fails
  on it: *"Tests must never create schemas in production; point
  `SAC_TEST_PG_DSN` at a throwaway database."*

I hit the second one first — a full-suite run against the primary returned
`15671 passed, 59 skipped, 2348 errors`, where **every one of the 2348 errors
was that refusal**, not a defect. `.github/ci/exec-in-sif.sh` has already ruled
on exactly this: *"Pointing the tests at the primary instead was considered and
rejected… Tests get their OWN database or none."*

No PostgreSQL server binaries exist in this container **or on the bare host**
(`command -v initdb pg_ctl postgres` is empty in both), so a throwaway cluster
cannot be provisioned from here. CI provisions its own and is the authority for
that slice.

Separately, a full in-container suite run stalls near the end regardless of
this PR — `tests/develop/test_audit.py` and a lifecycle test hang in-container
while passing in CI, and the host is currently saturated by other agents' test
runs. `pytest-matrix` is queued at 0s for the same reason.

### The residue is not zero, and here is what it is

160 lines outside `docs/adr/`. None of it is sac choosing a storage engine:

* **82** — the gate file itself. Its matches *are* the scan patterns, plus the
  operator's 2026-08-18 ruling quoted verbatim. Irreducible without deleting
  the guard, which the file's own prose forbids.
* **21** — `CHANGELOG.md`, historical entries plus this one. Rewriting past
  entries would be dishonest.
* **11** — the card id `sqlite-out-per-call-connect-cost-20260828`, cited where
  a live perf number is justified. Mangling an external identifier breaks the
  lookup that makes the number checkable.
* **~20** — other packages' storage: scitex-cards' shadow and its retired
  `SCITEX_CARDS_READ_BACKEND`, and scitex-db's suite (an exempt database
  *library*).
* **~26** — residual test prose, reducible but low-value. The operator's
  Japanese ruling in `config/_group_authority.py` is among it and is worth
  keeping verbatim.

ywatanabe1989:
## CI closed the gap: full suite green on a real PostgreSQL

`pytest-matrix-on-ubuntu-py3.11` on `5a60ca64`:

```
17994 passed, 84 skipped, 400 warnings in 307.21s (0:05:07)
```

Zero failures, zero errors. This is the slice the container could not run: CI
provisions its own throwaway cluster, so the PostgreSQL-backed tests that
skipped locally (157 in the edited-files run, 957 in the wider slice) actually
executed here — the skip count falls to **84**, and those are the
SLURM/e2e/psutil exclusions that skip everywhere.

So the earlier caveat is now discharged rather than merely stated: the store
accessors, the `instances` lifecycle, the grants/lineage/channel paths and the
`record_local_instance` signature change are all verified against a live
PostgreSQL, not inferred from a green subset.

`ruff`, `import-smoke`, `rtd-sphinx-build` and the hosted-runner guard are
also green. `pytest-matrix-on-ubuntu-py3.13` is still queued on the
self-hosted pool.

PR #1277, opened 2026-08-30
fix(rename): carry comms grants, and delete the SQLite rename path

ywatanabe1989:
`sac agents rename` silently failed to carry an agent's cross-group comms grants.

`comms_grants` moved to PostgreSQL on 2026-08-28, but its two pairs were the ONLY ones left behind in `_rename_db.NAME_COLUMNS`, and `rename_rows` SKIPS a table absent from `sqlite_master`. Measured on a state.db built the way the package builds one:

```
NAME_COLUMNS targets:         ['comms_grants']
_existing_tables sees:        (none)
count_rows('old-agent') ->    {}
```

`{}` is the defect in one line — even `--dry-run` reported nothing to carry.

**Not dead code.** `list_comms_grants()` returns real rows today; these are the explicit per-pair authorisations `_listen._acl` reads. A renamed agent starts answering to the new name while every grant still names the old one, so permitted sends begin to be DENIED — surfacing later as a permission error nobody connects to the rename. Nothing else renames them.

## The fix

`state_db_grants_rename`, shaped like `rename_comms_node` rather than `rename_instance_rows`: the directed pair IS the record identity, so this is copy + withdraw, not an in-place update.

- `created_at` carried forward — a renamed agent is the same agent
- a LIVE occupant is **folded, not refused**: a grant is not exclusive, so `new -> target` already existing means the authorisation is already in force, and refusing would block a legitimate rename over a no-op
- a REVOKED occupant is un-hidden, so a live grant is never silently downgraded
- the undo is key-scoped, because `new` may legitimately have held grants beforehand

## Why the module goes too

With those two pairs gone, `NAME_COLUMNS` and `PATH_COLUMNS` were both empty, so `_rename_db` could not touch anything — `rename_rows` looped over an empty tuple and reported success. `test__rename_db.py` asserted exactly that condition and went red the moment the last pair left; it was written for this moment. Module and test deleted, both leaving the SQLite ratchet with reasons recorded.

`_verify_nothing_left_behind` read `state.db` too. It now reuses `_count_rows_everywhere`, so the postcondition asks the STORE rather than a database that since 2026-08-28 could only answer `{}` — the check guarding the rename was itself vacuous.

## Why the file split

`_rename.py` was 571 lines against a 512 cap. The seam is not arbitrary: it held the forward orchestration AND the inverse of every step, so each store migration grew it twice. Inverses moved to `_rename_undo`; `move_path` moved to `_rename_plan` (which owns `Move`), so neither module imports the other.

## Verification

- 16 new tests against a real writable PostgreSQL, including the regression pin that `count_grant_rename_rows` reports the grant where the old path returned `{}`
- **2326 passed / 2 skipped / 0 failed** across `_lifecycle`, `_state` and the SQLite ratchet
- 705 insertions, 718 deletions — the fix is smaller than what it removes


ywatanabe1989:
Closing as DUPLICATE. develop already carries the complete fix — `state_db_grants_rename.py` with `rename_comms_grants` / `undo_rename_comms_grants` / `count_grant_rename_rows`, its test, `STEP_GRANTS` wired into `_rename` with an undo, and `_rename_db.py` plus the SQLite ratchet already deleted. Another agent shipped this while I was working on it; I did not check develop before starting.

Their design is BETTER than mine on the one point where we diverged, and that is worth recording rather than glossing. A hidden record at the destination is a REVOKE somebody performed deliberately. I chose to un-hide it, which silently reinstates a revoked authorisation as a side effect of a rename. They REFUSE, with one exception: the destination is revived only when its `created_at` is exactly the one being carried, which is evidence the two records are the same authorisation descended from an earlier forward rename — so a rename-back still works without making an ACL grant come back from the dead. That distinction did not occur to me.

Nothing here is worth salvaging over what is already merged.

PR #1286, opened 2026-09-03
feat(scratch): bind /uvwork from the host scratch volume, not the overlay upper

ywatanabe1989:
## The mechanism

Ninety of the 123 agent specs point `TMPDIR`, `UV_CACHE_DIR`, `UV_INSTALL_DIR`
and the agent venv (`/uvwork/venv-agent`) at `/uvwork`. The base image *creates*
that directory (`containers/apptainer-base.def`) and **nothing bound it** — so
an apptainer container's writes to it go to the overlay's **upper layer**, and
every agent's overlay lives on the host's **root logical volume**.

Measured on `scitex-compute-04`, 2026-09-03, in `overlays/<agent>/upper/uvwork`:

| agent | size |
| --- | --- |
| `sac` | 11.7 GB |
| `scitex-dev` | 3.3 GB |
| `scitex-hub` | 3.0 GB |
| `scitex-cards` | 2.5 GB |
| `scitex-storage` | 1.9 GB |

That root LV **filled to 0 four times on 2026-09-02**. `/scratch` on the same
host is a separate 3.0 T volume with 2.8 T free (compute-03 3.0 T / 2.8 T,
compute-01 295 G / 243 G), and **114 of the 123 specs already bind
`/scratch:/scratch:rw`**. The volume was mounted, bound, and unused for the one
thing filling the disk.

Operator directive, Telegram 2026-09-02: 「ディスクは /scratch 使ってくださいね」.

## What this does

1. **One knob, `scratch_root:` in the per-host `config.yaml`** — an absolute
   path, or the literal `none` (this host keeps `/uvwork` in the overlay), which
   is **refused without a `scratch_root_reason:`** beside it. No env var
   selects it. Absent, the default `/scratch` is probed — compute-01 and
   compute-03 have no `config.yaml` at all. With neither a declaration nor a
   `/scratch`, sac **refuses to start the agent**, naming the missing path, the
   key, the file and all three fixes. Falling back to the overlay silently is
   the bug being replaced.
2. **The bind.** Each start creates `<root>/sac/agents/<agent>/uvwork` (0700,
   idempotent — an existing directory keeps its mode and contents) and appends
   `--bind <that>:/uvwork:rw` in `_apptainer_argv_finalize`, after every
   spec-declared bind so a spec that binds `/uvwork` itself still wins, and
   before the secret lift so the credentials bind keeps its last-bind position.
   **No spec is edited**; `binds:` and `startup_commands` are untouched fleetwide.
3. **`sac agents scratch-migrate`** moves each STOPPED agent's stranded overlay
   copy across. Dry-run by default; `--apply` is the deliberate act; copy →
   **verify** every path, size and symlink target → only then remove the source.
   Exit codes: 0 sound, 1 the plan does not describe the sweep, 2 refused.
4. **ADR-0024** records the decision, the rejected alternatives, and the honest
   edge (a `scratch_root:` under `/home` gives jailed capsules a bind
   `enforce_jail` does not police, because sac's own binds never were in scope
   for that check).

## Two defects the first real dry-run found

Running it for real is why this PR is what it is.

**(a) The probe called the agent running it "stopped."** From inside the
`scitex-agent-container` agent, `is_running` (a pid file plus
`os.kill(pid, 0)`) reported the agent *executing the probe* as stopped and
offered its 10.3 GiB. The recorded pid was `3190806`; that container's `/proc`
tops out at `74275`, because the container has its own PID namespace. Neither
the pid file nor the adapter is wrong — the **vantage** is. So the verb now
abstains outright when `APPTAINER_CONTAINER` / `SINGULARITY_CONTAINER` is set:
every agent comes back UNKNOWN, which is a refusal, and the message says to run
it on the host. Sizes still print, because the overlays are read through a bind
mount and *that* reading is faithful.

**(b) One tree, listed as movable twice.** `scitex-hub` and
`scitex-hub-mobile-ux` declare the **same** `--overlay`, as do `scitex-cards` /
`scitex-todo` and eight `handyman-*` specs sharing `local-coder`. Applying that
would move the tree for the first agent, hand the second a source that no longer
exists, and file a shared tree into one agent's private directory. Which agent
owns a shared overlay is a fleet decision sac may not make, so **both** rows
refuse, each naming the others. Ownership is computed over the whole roster,
never the `--agent` subset.

Together they took the headline from a wrong **28.1 GiB across 17 agents** to a
truthful **8.1 GiB across 12**.

## The real dry-run (read-only), from the host vantage

`sac agents scratch-migrate` over the live 116-spec roster on
scitex-compute-04. `nothing` rows elided; nothing was written.

```
sac agents scratch-migrate  dry-run (read-only)

scratch root: /scratch (default: /scratch is a mount point on this host and
/home/ywatanabe/.scitex/agent-container/config.yaml declares no scratch_root:)
116 spec(s) under /home/ywatanabe/.scitex/agent-container/agents

  MOVE    canary-resume-test                 53.7 MiB  stopped; 56350248 bytes in 33 file(s) move to /scratch/sac/agents/canary-resume-test/uvwork
  MOVE    claude-code-telegrammer            89.7 MiB  stopped; 94069889 bytes in 1129 file(s) move to /scratch/sac/agents/claude-code-telegrammer/uvwork
  MOVE    figrecipe                         566.5 MiB  stopped; 594006840 bytes in 20763 file(s) move to /scratch/sac/agents/figrecipe/uvwork
  REFUSED scitex-agent-container             10.3 GiB  RUNNING (TuiSessionRuntime); stop it first: sac agents stop scitex-agent-container
  MOVE    scitex-agent-container-04         534.7 MiB  stopped; 560716126 bytes in 27551 file(s) move to /scratch/sac/agents/scitex-agent-container-04/uvwork
  MOVE    scitex-app                         64.6 MiB  stopped; 67748138 bytes in 1985 file(s) move to /scratch/sac/agents/scitex-app/uvwork
  REFUSED scitex-cards                        2.3 GiB  RUNNING (TuiSessionRuntime); stop it first: sac agents stop scitex-cards
  MOVE    scitex-cards-gui                  214.5 MiB  stopped; 224971247 bytes in 25486 file(s) move to /scratch/sac/agents/scitex-cards-gui/uvwork
  MOVE    scitex-db                           1.6 GiB  stopped; 1734531125 bytes in 122972 file(s) move to /scratch/sac/agents/scitex-db/uvwork
  MOVE    scitex-dev                          2.6 GiB  stopped; 2826297786 bytes in 209435 file(s) move to /scratch/sac/agents/scitex-dev/uvwork
  MOVE    scitex-hpc                        137.9 MiB  stopped; 144617303 bytes in 14439 file(s) move to /scratch/sac/agents/scitex-hpc/uvwork
  REFUSED scitex-hub                          2.6 GiB  /home/ywatanabe/.scitex/agent-container/containers/overlays/scitex-hub/upper/uvwork is ALSO declared by scitex-hub-mobile-ux; moving it into /scratch/sac/agents/scitex-hub/uvwork would take it away from them. Give each agent its own overlay, or move this tree by hand.
  REFUSED scitex-hub-mobile-ux                2.6 GiB  /home/ywatanabe/.scitex/agent-container/containers/overlays/scitex-hub/upper/uvwork is ALSO declared by scitex-hub; moving it into /scratch/sac/agents/scitex-hub-mobile-ux/uvwork would take it away from them. Give each agent its own overlay, or move this tree by hand.
  MOVE    scitex-live-paper                 592.8 MiB  stopped; 621556616 bytes in 4116 file(s) move to /scratch/sac/agents/scitex-live-paper/uvwork
  MOVE    scitex-storage                      1.6 GiB  stopped; 1667501297 bytes in 236732 file(s) move to /scratch/sac/agents/scitex-storage/uvwork
  REFUSED scitex-todo                         2.3 GiB  /home/ywatanabe/.scitex/agent-container/containers/overlays/scitex-cards/upper/uvwork is ALSO declared by scitex-cards; moving it into /scratch/sac/agents/scitex-todo/uvwork would take it away from them. Give each agent its own overlay, or move this tree by hand.
  MOVE    scitex-ui                          91.0 MiB  stopped; 95406435 bytes in 15385 file(s) move to /scratch/sac/agents/scitex-ui/uvwork

12 agent(s), 8.1 GiB (8687773050 bytes) would move from the overlay upper to
/scratch
5 refused: scitex-agent-container, scitex-cards, scitex-hub,
scitex-hub-mobile-ux, scitex-todo

Nothing was moved — this is a dry-run. To act:
    sac agents scratch-migrate --apply
```

`--json` on the same run: `scratch_source: "default"`, `liveness_vantage: ""`
(host), `roster: "populated"`, `specs: 116`, `safe_to_apply: true`,
`total_bytes: 8687773050`, `exit_code: 0`.

**`--apply` was NOT run against the real overlays.**

The same command from *inside* an agent container, for contrast — the guard
doing its job:

```
LIVENESS UNREADABLE FROM HERE — this process runs INSIDE a container
(APPTAINER_CONTAINER=…/sac-base-2026-0902-195723.sif), which has its own PID
namespace, so an agent's recorded host pid resolves to nobody here and every
agent reads as 'stopped'. Run `sac agents scratch-migrate` on the HOST instead.

0 agent(s), 0 B (0 bytes) would move from the overlay upper to /scratch
17 refused: canary-resume-test, claude-code-telegrammer, figrecipe,
scitex-agent-container, …
```

## Positive controls

Every new behaviour was shown to FAIL without its production code. Each control
patched one thing, ran the suite, and reverted; the summaries are the red ones.

| control | change | red |
| --- | --- | --- |
| A | drop `argv += uvwork_bind_flags(...)` | `test_build_run_argv_binds_uvwork_from_the_host_scratch_root` — 1 failed, 18 passed |
| B | resolver falls back to the overlay instead of refusing | 5 failed, 16 passed (all four refusal-message rows) |
| C | `scratch_root: none` no longer needs a reason | 2 failed, 13 passed |
| D | undeterminable liveness read as "stopped" | 2 failed, 32 passed |
| E | remove the overlay copy without verifying | 3 failed, 31 passed |
| F | CLI applies by default | 4 failed, 15 passed |
| G | create the bind source 0755 | `test_the_created_directory_is_owner_only` — 1 failed, 18 passed |
| H | the container vantage guard never fires | 8 failed, 24 passed |
| I | a shared overlay is moved for whoever comes first | 4 failed, 19 passed |

Control A's companion red control (`…emits_no_uvwork_bind_on_a_none_host`)
stays green under A by design — it is the control, not the assertion.

## Tests

New: `_state/test_host_scratch.py` (the resolution table, all four rows plus the
answer shape), `_state/test_host_config_scratch.py` (block parsing, the
reason-less `none` and the orphan reason), `runtimes/test__apptainer_scratch.py`
(layout, 0700, idempotence, spec-bind-wins, and the bind in a real
`build_run_argv` argv with a red control), `_maintenance/test__scratch_migrate*.py`
(planning, applying, measuring, liveness) and
`cli_pkg/test__agents_scratch_migrate.py` (dry-run default, exit codes, the
vantage). No mocks anywhere: real specs, the real loader, real trees, the real
`copytree` and `rmtree`. Only liveness is injected, through the module's
documented seam — a test cannot make a real agent run.

`_scratch_migrate.py` crossed the 512-line ceiling while gaining the two guards
and was split by responsibility into `_scratch_migrate_measure.py` (`tree_size`
/ `verify_copy`) and `_scratch_migrate_liveness.py` (`agent_liveness` /
`liveness_vantage`), both re-exported so the import site is unchanged. The test
files split to match, sharing `tests/.../_helpers/scratch_fleet.py`.

## Note on `tests/conftest.py`

The suite floor now writes a per-worker `config.yaml` declaring a `scratch_root`
inside the sandbox. Without it, ~20 test modules that assemble a launch argv
would each create `sac/agents/agt/uvwork` on the operator's real `/scratch`
(present on every compute host and inside every agent container), and on a box
*without* `/scratch` the same tests would refuse the launch instead — the
resolver doing its job on the wrong population.

https://claude.ai/code/session_01Adzed4bZzRxEDbMh9QoRtV

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,59 @@ versioning follows [SemVer](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- **`/uvwork` is bound from the host scratch volume instead of accumulating in
+  the apptainer overlay upper on the root LV** (ADR-0024; operator directive
+  2026-09-02 「ディスクは /scratch 使ってくださいね」). Ninety of the 123 agent
+  specs point `TMPDIR`, `UV_CACHE_DIR`, `UV_INSTALL_DIR` and the agent venv at
+  `/uvwork`; the base image creates that directory and **nothing bound it**, so
+  every byte went to `overlays/<agent>/upper/uvwork` on the host's ROOT volume.
+  Measured on `scitex-compute-04` 2026-09-03: 11.7 GB (`sac`), 3.3 GB
+  (`scitex-dev`), 3.0 GB (`scitex-hub`), 2.5 GB (`scitex-cards`), 1.9 GB
+  (`scitex-storage`) — and the root LV filled to 0 **four times** on
+  2026-09-02, while `/scratch` on the same host is a separate 3.0 T volume with
+  2.8 T free.
+  Each start now resolves the host scratch root once, creates
+  `<root>/sac/agents/<agent>/uvwork` (mode 0700, idempotent across restarts)
+  and appends `--bind <that>:/uvwork:rw` in the argv finalize layer, after
+  every spec-declared bind so a spec that binds `/uvwork` itself still wins.
+  No spec changes: `binds:` and `startup_commands` are untouched fleetwide.
+- **`scratch_root:` (and `scratch_root_reason:`) in the per-host
+  `config.yaml`.** An absolute path, or the literal `none` — the written
+  decision that this host keeps `/uvwork` in the overlay, which is **refused
+  without a stated reason**, as is a `scratch_root_reason:` with no
+  `scratch_root:` beside it. With no declaration the default `/scratch` is
+  probed (compute-01 and compute-03 have no `config.yaml` at all, so the
+  default has to work without one); with neither a declaration nor a
+  `/scratch`, sac **REFUSES to start the agent** naming the missing path, the
+  config key, the config file and all three fixes. There is no env-var knob:
+  falling back to the overlay silently is the bug this replaces.
+- **`sac agents scratch-migrate [--agent NAME]… [--apply] [--json]`** — moves
+  each STOPPED agent's `overlays/<agent>/upper/uvwork` to
+  `<scratch_root>/sac/agents/<agent>/uvwork`, so the historical copy stops
+  occupying the root LV and the next start finds uv and the venv already in
+  place. **Dry-run by default**, printing per-agent sizes, per-agent decisions
+  and the total it would move. A RUNNING agent is refused by name (its
+  container has the overlay mounted) and so is one whose liveness the runtime
+  adapter could not determine — "unknown" is not "stopped". Applying copies,
+  **verifies** every path, size and symlink target against the source, and only
+  then removes the overlay copy; a verification mismatch keeps the source and
+  says so. Exit codes: 0 sound, 1 the plan does not describe the sweep, 2
+  refused.
+  Two refusals come straight out of the first real dry-run, which found both:
+  **(a)** run from inside an agent container, the probe called the agent
+  executing it "stopped" and offered its 10.3 GiB — `is_running` is a pid file
+  plus `os.kill(pid, 0)`, the recorded pid was 3190806 and the container's
+  `/proc` topped out at 74275, because the container has its own PID namespace.
+  The verb now abstains whenever `APPTAINER_CONTAINER` /
+  `SINGULARITY_CONTAINER` is set, naming the vantage and saying to run it on
+  the host; sizes still print, since the overlays are read through a bind mount.
+  **(b)** `scitex-hub` and `scitex-hub-mobile-ux` declare ONE `--overlay`, as do
+  `scitex-cards` / `scitex-todo` and eight `handyman-*` specs, so one 2.6 GiB
+  tree was listed as movable twice. Shared sources are now refused on every
+  claiming row, each naming the others, with ownership computed over the whole
+  roster rather than the `--agent` subset.
+
 ### Removed
 - **`state_db.DEFAULT_DB_PATH`, and the ~4900-times-per-run test ceremony that
   rebound it.** The engine deletion below left the PATH behind — a `Path`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,19 @@ versioning follows [SemVer](https://semver.org/).
   (`scitex-storage`) — and the root LV filled to 0 **four times** on
   2026-09-02, while `/scratch` on the same host is a separate 3.0 T volume with
   2.8 T free.
-  Each start now resolves the host scratch root once, creates
-  `<root>/sac/agents/<agent>/uvwork` (mode 0700, idempotent across restarts)
-  and appends `--bind <that>:/uvwork:rw` in the argv finalize layer, after
-  every spec-declared bind so a spec that binds `/uvwork` itself still wins.
+  Each start now resolves the host scratch root once and appends
+  `--bind <root>/sac/agents/<agent>/uvwork:/uvwork:rw` in the argv finalize
+  layer, after every spec-declared bind so a spec that binds `/uvwork` itself
+  still wins. The directory (mode 0700, idempotent across restarts) is created
+  on the REAL launch path only — `build_run_argv` is also what `sac agents
+  explain` and `sac agents start --dry-run` call, and a read-only command must
+  neither write to the host nor fail on a launch-time host condition; on a
+  host with no scratch root those two show the refusal a start would hit
+  instead of raising it. The per-agent directory is derived by ONE function
+  (`scratch_uvwork_dir_for`, keyed on the effective `config.name`) that the
+  launch bind and `scratch-migrate` both call, so the two cannot disagree —
+  they would for every `hosts:` spec, whose effective id carries a
+  `-<hostname>` suffix its spec directory name does not.
   No spec changes: `binds:` and `startup_commands` are untouched fleetwide.
 - **`scratch_root:` (and `scratch_root_reason:`) in the per-host
   `config.yaml`.** An absolute path, or the literal `none` — the written
--- a/docs/adr/0024-uvwork-on-the-host-scratch-volume.md
+++ b/docs/adr/0024-uvwork-on-the-host-scratch-volume.md
@@ -111,14 +111,39 @@ write the `none` decision down).
 
 ### 2.3 The bind
 
-When a root resolves, `runtimes/_apptainer_scratch` creates
+When a root resolves, `runtimes/_apptainer_scratch` appends
+`--bind <that>:/uvwork:rw` to the flag argv, where *that* is
 
 ```
 <scratch_root>/sac/agents/<agent-name>/uvwork      mode 0700
 ```
 
-before exec and appends `--bind <that>:/uvwork:rw` to the flag argv. Notes
-that are decisions rather than details:
+and creates it — but the two halves happen at different moments, which is a
+decision, not an implementation detail. Notes that are decisions rather than
+details:
+
+* **The `<agent-name>` in that path is the EFFECTIVE id (`config.name`), and
+  exactly one function spells it** — `scratch_uvwork_dir_for`. The launch
+  bind and `sac agents scratch-migrate` both call it. They must, because for
+  a `hosts:` spec the effective id carries a `-<hostname>` suffix that the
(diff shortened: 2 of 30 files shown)


ywatanabe1989:
Both review findings are fixed in `7e2730bf`.

### 1. The read-only surfaces are read-only again

`uvwork_bind_flags` sat on `build_run_argv`, which `sac agents explain` and `sac agents start --dry-run` also call — so both created a directory on the host, and on a host with no resolvable scratch root both raised. The file's history already records that exact regression (it is why `verify_tmpfs_headroom` was lifted out of `tmpfs_workdir_flags`, and why `reconcile_overlay_venv_for_launch` sits past the `dry_run` return), so the fix reuses that mechanism rather than inventing one.

- `plan_uvwork_bind` decides the bind as a **value**, touching nothing and never raising on a host condition — a `ScratchRootError` comes back in the value's `error` field.
- `uvwork_bind_flags` is now a read-only wrapper over it: on an unresolvable host it logs at `WARNING` (so it reaches stderr through the last-resort handler even with no logging configured) and emits an argv with no bind.
- The mkdir and the raise moved to `ensure_uvwork_for_launch`, called past the `dry_run` return from **both** real launch paths — `_apptainer_runtime.start` (which `claude_session` and `openai_session` both delegate into) and `tui_session.start`. **The refusal is unchanged**: same error, same message, same three fixes named.
- It creates the directory *this argv will actually mount* — the bind source is read back out of the finished argv — so a spec that declared its own `/uvwork` bind is left to its owner. Same derive-from-what-launches rule `tui_session` already uses for the SIF.
- `explain` renders binds by walking the argv, so the one outcome it could not show was the one that emits no bind — precisely where a start *will* refuse. `render_plan` now carries an explicit `/uvwork:` line stating the resolved source, or the refusal.
- Landing the TUI half needed one line more than `tui_session.py` had under the 512-line cap, so its three launch-time steps (uvwork, overlay-venv reconcile, entry-point probe) moved to `_tui_launch_gate.run_launch_gate`. They share one placement invariant — each writes to the host or refuses to start — so this follows the module's own delegation pattern. Net −20 lines there.
- The two `test__apptainer_runtime.py` tests that had to declare their own `scratch_root:` to survive the raise are reverted; the workaround is unnecessary now and its comment had become false.

### 2. The migration destination and the launch bind can no longer disagree

`_scratch_migrate` keyed on `spec_path.parent.name` (spec **directory** name) while the bind keyed on `config.name` (**effective** id). Identical for a `host:` spec, different for every `hosts:` one (`-<hostname>` suffix) — so for any multi-instance agent the migration copied gigabytes to a path no launch would ever mount, verified the copy, deleted the overlay original and reported success. Nothing failed; the next start just rebuilt uv and the venv from nothing.

Both sides now call `scratch_uvwork_dir_for(root, config)`, the single place an agent's identity becomes a path component.

### Positive controls (red → green)

**Finding 1** — restored the old behaviour inside `plan_uvwork_bind` (resolver raises, `ensure_scratch_uvwork` on the emit path):

```
FAILED ...test__apptainer_scratch.py::test_emitting_the_flags_creates_NOTHING
FAILED ...::test_build_run_argv_creates_nothing_under_the_scratch_root
FAILED ...::test_build_run_argv_does_not_raise_when_no_scratch_root_resolves
FAILED ...::test_an_unresolvable_root_is_a_WARNING_not_a_silent_absence
FAILED ...::test_the_explain_line_names_the_refusal_a_start_would_hit
5 failed, 26 passed in 0.46s
```

**Finding 2** — reverted `_row_for` to `scratch_uvwork_dir(scratch_root, agent)`:

```
FAILED ...test__scratch_migrate.py::test_the_destination_for_a_hosts_spec_is_the_launch_bind_source
E   AssertionError: assert '/uvwork/tmp/.../multi/uvwork' == '/uvwork/tmp/...ute-04/uvwork'
E     - ents/multi-scitex-compute-04/uvwork
E     + ents/multi/uvwork
1 failed, 56 passed in 0.96s
```

The `host:`-spec companion row stayed green in that control, so the test discriminates rather than merely being sensitive.

Restored, both green.

### Suites

`runtimes`, `config`, `_state`, `_maintenance`, `cli_pkg`:

```
7071 passed, 1457 skipped, 16 deselected, 1 warning in 321.51s
```

One note worth recording: the stderr-visibility rows use `capfd`, not `caplog`. Measured twice, both `caplog` and a handler attached to the logger by name reported an **empty** record list in the full suite while pytest's own captured-output sections printed the very `WARNING` they claimed was absent — an instrument that denies a message it is simultaneously displaying is the one thing a visibility test must not use. Stderr is also the property actually being claimed.

PR #1313, opened 2026-09-06
feat(ratelimit): switch a Fable-capped agent to opus instead of leaving it silent

ywatanabe1989:
## What

A **fourth agent-liveness shape** for `_ratelimit`, and the remedy for it: an agent frozen behind a **model cap** is moved onto `opus[1m]` and kicked, instead of being waited on forever.

Four new modules, keeping this package's existing separation (pure parser / pure rule / one mutation / IO pass) rather than growing a subsystem:

| module | what it is |
| --- | --- |
| `_ratelimit/_modelcap.py` | PURE parser — both measured banners, a three-valued `readable`/`capped` observation, and `verify_switch()` |
| `_ratelimit/_switch_rule.py` | PURE rule — `SWITCH_MODEL(target="opus[1m]")` or today's verdict, nothing in between |
| `_ratelimit/_switch.py` | the mutation — the operator's three steps, three seconds apart, then a verifying capture |
| `_ratelimit/_switch_pass.py` | its own ledger (`SAC_MODEL_SWITCH_HISTORY`) and perform leg |

Wired as `sac agents resume-rate-limited --switch-model`, **OFF by default**. No new timer is armed.

## Why

**Measured 2026-09-06.** The operator sent two messages to a Fable-family agent and BOTH were answered by the harness with:

```
You've reached your Fable limit. Run /usage-credits to continue or switch models with /model.
```

The agent went silent to him. Three workflow subagents died in the same window with `You've hit your session limit · resets 2am (UTC)`.

sac's rate-wall reviver recovered neither, and it was not wrong to:

* `src/scitex_agent_container/_ratelimit/_banner.py:66` — `LIMIT_RE` anchors on `hit your <word> limit`. The Fable banner says **reached**, so it is invisible to the matcher.
* the Fable banner publishes **no reset clause at all**, so even a matcher that saw it would have nothing to wait for. `src/scitex_agent_container/_ratelimit/_rule.py:203` therefore returns `NOT-LIMITED` — "nothing for this pass to do" — about a visibly capped agent.
* the session-limit shape *does* publish an end, hours away.

A rate wall is a pause with a published end and waiting is the only correct action. A model cap is a wall to **walk around**, and the provider's own banner says how.

### Three decisions a reviewer should judge

**1. Steps 1–2 do not go through `_delivery.deliver`, and that is measured, not assumed.** `_resume.py` uses `deliver` and says why (an unverified send leaves a prompt unsubmitted at `❯`). That reasoning holds — for prose. It cannot carry a slash command: `_delivery/_deliver.py:127` calls `format_payload(message, token)` unconditionally, which returns `f"[sac-deliver:{token}] {message}"`. The payload reaches the composer as `[sac-deliver:a1b2c3] /model opus[1m]` — first character `[`, not `/` — which the TUI submits as prose about a model rather than an instruction to change one. There is no flag to suppress the token; it is the arrival matcher's whole mechanism. So steps 1–2 use the package's tmux helper, and **step 3 (the kick) IS prose and does go through `deliver`**, inheriting its proof that the payload left the compose box.

**2. Nothing is claimed from a send that returned 0.** `tmux send-keys` exiting 0 means tmux accepted a keystroke. The last thing `switch_model_now` does is capture the pane again and ask `verify_switch`, three-valued: `SWITCHED` / `SWITCH-FAILED` / `SWITCH-UNVERIFIED` (exit 2, alongside the other unresolved readings). The verifier also **subtracts the target-model occurrences sac itself typed into that pane** — a verifier that simply searched for `opus` would find its own keystrokes and certify every attempt, successful or not.

**3. The default is load-bearing.** With `--switch-model` unset the pass is byte-for-byte the pass it was, down to the exit code; `test_the_flag_off_changes_nothing` is the red half of a control whose only variable is the flag.

Also note `_account/rate_limit_signals.py:167` (#1288, merged onto develop while this branch was in flight) matched the same sentence for a **different** consumer — account rotation, over error text, deliberately firing on the whole cap family. Reusing it here would fire a model switch on every weekly-window phrasing and move agents off models that were never capped. The distinction is written into `_modelcap.py`'s docstring rather than left for the next reader to re-derive.

## Test plan

```
PYTHONPATH=<worktree>/src /opt/venv-sac/bin/python -m pytest <worktree>/tests/scitex_agent_container/_ratelimit -q -p no:cacheprovider
  -> 111 passed (44 new)

... tests/scitex_agent_container/{cli_pkg,_jobs} tests/scitex_agent_container/test_cli.py
  -> 2893 passed, 856 skipped, 16 deselected

uv tool run ruff check src/scitex_agent_container/_ratelimit src/.../cli_pkg/_agents_resume_rate_limited.py tests/.../_ratelimit
  -> 2 I001 findings, BOTH pre-existing on develop (_pass.py:31, test__pass.py:26) in blocks this PR does not touch;
     every line this PR changed is clean
uv tool run ruff check --select T201,T203 src/   ->   All checks passed
```

New coverage, and what each half is for:

* **parser** — both real banner strings (parametrised over the verbatim specimens), a **negative control** (an ordinary pane, including an agent talking *about* the Fable limit in prose, must not match), and an unreadable pane that reports `readable=False` rather than `capped=False`.
* **rule** — fires on the measured case; `ALREADY-ON-TARGET` for an agent already on opus; `COOLING-DOWN` for a second fire in one incarnation; `MOVING` for an advancing pane; a non-Fable agent behind the session wall keeps today's verdict.
* **mutation** — the three steps in the operator's order, `[0.0, 3.0, 6.0, 9.0]` on an **injected clock** (the fake sleep advances it, so no test sleeps), the verification capture runs, and all three of `True`/`False`/`None` come back from the pane rather than the send.
* **pass** — the flag-off / flag-on control, read-only stays read-only, the ledger stops a second switch, the resume ledger is *not* spent, and an unverified switch reaches the event log as degraded.

https://claude.ai/code/session_01AaiWJHYdhALg2wB4hRNft1

### DIFF
--- a/src/scitex_agent_container/_ratelimit/__init__.py
+++ b/src/scitex_agent_container/_ratelimit/__init__.py
@@ -31,6 +31,24 @@ Layout — the rule is separated from the doing, as in the siblings:
 Driven by ``sac agents resume-rate-limited`` (read-only by default) and
 scheduled as the ``sac.resume-rate-limited-agents`` JobSpec.
 
+THE FOURTH SHAPE — a wall that WAITING cannot end (2026-09-06)
+    A rate wall publishes an end. A MODEL cap does not. Measured tonight: the
+    operator sent two messages to a Fable-family agent and both were answered
+    by the harness with ``You've reached your Fable limit. Run /usage-credits
+    to continue or switch models with /model.`` — no reset clause anywhere, so
+    the machinery above has nothing to wait for and correctly reports
+    ``NOT-LIMITED`` while the agent stays silent to the operator. Three
+    workflow subagents died in the same window behind ``You've hit your
+    session limit · resets 2am (UTC)``, which does publish an end — hours
+    away.
+
+    Both are walls a MODEL SWITCH ends in seconds, so the same four-part
+    separation is repeated for that remedy: :mod:`._modelcap` (the pure
+    parser), :mod:`._switch_rule` (the pure rule), :mod:`._switch` (the one
+    mutation — the operator's three steps, three seconds apart, then a
+    verification) and :mod:`._switch_pass` (its own ledger and perform leg).
+    It is OFF by default and armed with ``--switch-model``.
+
 WHY THIS CANNOT HOT-LOOP AGAINST A LIVE LIMIT
     Structurally, not by tuning. The resume branch of :func:`._rule.decide`
     is unreachable until ``now >= reset_at``, where ``reset_at`` is read from
@@ -44,6 +62,7 @@ from __future__ import annotations
 
 from ._alarm import SUBSYSTEM
 from ._banner import LimitObservation, observe_pane
+from ._modelcap import ModelCapObservation, observe_model_cap, verify_switch
 from ._pass import (
     DEFAULT_INTERVAL,
     DEFAULT_PASS_CAP,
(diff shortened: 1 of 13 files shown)


ywatanabe1989:
Correction to the commit message on 13775817.

It says "Negative control run: with the rung deleted, all three fail". That is wrong, and I wrote it before reading the control output rather than after. The measured result is **1 of 3**:

```
rung removed:  1 failed, 19 passed
  FAILED ...::test_it_outranks_the_echo_arithmetic
rung restored: 20 passed
```

So only `test_it_outranks_the_echo_arithmetic` uniquely pins the new rung. The other two (`test_the_confirmation_proves_the_switch`, `test_a_confirmation_naming_another_model_is_not_our_switch`) pass with the rung deleted because the older presence-gated-arithmetic rung already reaches the same verdict on those two panes. They are still worth keeping as regression cover for the phrase matching, but they are not evidence that this rung does anything.

The one test that does discriminate is the one that matters: target id present ONLY as sac's own echo, kick not submitted. Every older rung refuses that pane as unknown; this rung reads the session's own "Set model to" line and decides. That is the behaviour the operator asked for.

No code change - the rung and the tests are as reviewed. This corrects the claim, not the work.

PR #1335, opened 2026-09-11
fix(hermes): restrict each generated provider to its selected model

ywatanabe1989:
## Summary
- disable endpoint model discovery in SAC-generated Hermes providers
- expose exactly the selected configured model in the provider catalog
- prevent a paid provider's other models from appearing as runtime choices

## Verification
- 21 targeted tests passed
- Ruff passed
- diff check passed

This is a model-selection guard. Response-reported model verification belongs in the explicit paid engine probe and is intentionally not run during ordinary starts.


ywatanabe1989:
Independent review HOLD at exact head `5a752188764e7a1703067ba907cf2e9c4d021b87`. `discover_models: false` narrows picker discovery but is not an outbound allowlist. Two reproduced bypasses: (1) because the container receives the canonical `DEEPSEEK_API_KEY`, Hermes also authenticates its built-in `deepseek` provider, which exposes Flash and Pro; (2) explicit `switch_model("deepseek-v4-pro", explicit_provider="sac-paid-flash")` succeeds against a catalog endpoint even though the generated provider maps only Flash. This PR must not be merged as a never-Pro guarantee. Required end state: remove built-in credential collision and enforce the exact model at the outgoing request/gateway boundary. Focused 95 tests otherwise pass.

ywatanabe1989:
Hermes foundation prerequisite has since landed on `origin/develop` (`d88f00caad325c6e4a40a1e75c0c0a1151c798f6`). Current develop now has:

- `AGENT_HARNESSES` including `hermes` from the central harness registry;
- `HermesTuiSessionRuntime` selected for `harness=hermes,runtime=tui`;
- `build_run_argv` accepting the Hermes TUI path;
- pinned Hermes source/image staging;
- real v3 spec-load → engine selection → runtime materialization → `build_run_argv` integration coverage.

Reconciliation evidence on clean `origin/develop`: **104 targeted tests passed**, and live import readback reports Hermes registered with `HermesTuiSessionRuntime`.

This PR is now only a two-file provider-discovery guard, but a synthetic merge into current develop has a content conflict in `tests/scitex_agent_container/config/test__hermes_config.py`; source `_hermes_config.py` auto-merges. The head is authored/committed by `ywatanabe`, so the fleet agent has not modified or pushed it. It needs a history-preserving merge of current develop by its owner, conflict resolution in that test file, exact-head tests/CI, and fresh review.

ywatanabe1989:
Closing as unsafe/incomplete rather than leaving a known-failing green-looking PR in inventory. Independent review reproduced that `discover_models: false` is not an outbound model allowlist and that built-in credentials/explicit provider selection can bypass the intended never-Pro guarantee. Exact-model enforcement must live at the outgoing gateway/request boundary; this two-file discovery guard must not be merged.

Issue #1340, opened 2026-09-11
TUI delivery verifier misclassifies active Hermes turn as pasted-but-unsent


ywatanabe1989:
## Observed incident

On compute-03 on 2026-09-11, sac agents deliver submitted a Hub turn. The live Hermes pane showed a reasoning spinner and Ctrl+C to interrupt, and scitex-genai gateway health increased in_flight from 3 to 4. However, delivery returned exit 1 and claimed the prompt was still unsent because the submitted prompt text remained visible in the transcript.

The raw pane itself contained both the prompt and an active busy marker. Treating prompt-text persistence as an unsent composer produced a false negative.

## Required behavior

- Distinguish transcript/rendered user-message text from the live composer.
- An observed busy marker after Enter should prove submission unless stronger contrary evidence exists.
- Use gateway/turn-bridge acknowledgement when available, while keeping the public delivery result harness-neutral.
- Add a Hermes Ink TUI regression where submitted text stays rendered while the agent is actively reasoning.
- Never resend solely because this ambiguous display state was observed.

ywatanabe1989:
Closed by the SciTeX 72-hour freshness policy: this issue had no activity for more than three days. If it is still relevant, create a fresh issue with current evidence and scope.

PR #1342, opened 2026-09-11
fix(hermes): recover stale provider latch in place

### LINKED ISSUE #1339: Hermes sessions stay latched after inference gateway recovers
## Observed incident

On compute-03 on 2026-09-11, multiple Hermes/Qwen application sessions accumulated five stale provider attempts while the single Qwen upstream was saturated. After scitex-genai gateway PR #56 was deployed and the gateway health endpoint reported active=true with available admission capacity, ordinary autonomous turns still aborted immediately without a network request:

    Provider has been unresponsive ... for 5 consecutive stale attempts — aborting this call

The installed Hermes source documents that the cross-turn stale streak resets only after a successful call or provider swap. SAC health and autonomous idle kicks therefore cannot recover the agent after the external provider becomes healthy.

A same-provider TUI command restored the session without clearing its history:

    /model qwen38-27b --provider sac-qwen38-27b

After that command, the next Hub turn reached the gateway and preflight compression started.

## Required behavior

- Detect that a configured inference provider has recovered before submitting an autonomous kick.
- Reset/rebind the same Hermes provider without clearing the persistent session, or use a supported Hermes runtime API that performs the equivalent.
- Bound probes and retries so a fleet cannot create a recovery stampede.
- Preserve the agent's session/context and incarnation identity.
- Expose the stale-latched state separately from ordinary READY/idle in SAC status and the Agents GUI.
- Add an integration regression covering: stale latch -> provider health recovery -> same session resumes and performs a real turn.

The fix should remain harness-neutral at the SAC state/control surface; Hermes-specific recovery belongs in its runtime adapter.

ywatanabe1989:
Closes #1339.

## What changed

- adds a harness-neutral `runtime_control.turn_admission` status/control boundary, surfaced by CLI/listener status and the Agents GUI
- adds one bounded Hermes recovery observer for autonomous TUI sessions
- recognizes only Hermes 0.21.1's exact stale-provider breaker error, then pauses the native heartbeat before any recovery probe
- probes the configured provider's non-generating `/health` endpoint on a 30-second cadence with a stable per-agent phase; recovery requires an active member with immediate admission capacity
- uses Hermes' documented session-scoped controls to rebind the same model/provider and resume heartbeat scheduling, without stop/start, `/new`, or context deletion
- lets authoritative dead/not-started liveness outrank persisted stale/recovering control observations, and resets the observation when the owning mux naturally exits
- records only adapter control state/PID/log artifacts, preserving Hermes session/context and SAC incarnation identity
- documents the boundary and explicitly leaves #1340's composer/transcript delivery classification out of scope

## Verification

- HOLD regression suite (projection, monitor, synthetic integration, audit): 31 passed
- proportional Django/lifecycle/listener/Hermes suite: 2,816 passed, 650 skipped
- original full suite: 16,906 passed, 2,453 skipped; 2 unrelated existing failures in `config/test__resolve.py` (stderr capture assertions), reproduced in isolation
- repository audit: passed
- Ruff on all changed Python files: passed
- `git diff --check`: passed

The integration regression uses a behavioral fake Hermes session around real local HTTP health and completion boundaries. It verifies adapter orchestration plus session-id, incarnation-id, and `state.db` persistence invariants; it is not production proof of the pinned Hermes TUI path. Validation in a disposable real Hermes environment remains operational follow-up evidence.


ywatanabe1989:
Independent re-review PASS at exact head `a105e44a43c8bd91a4f072a910ff5a12d19ce6fc`. Prior liveness-overlay HOLD is fixed: dead/not-started state dominates persisted recovery markers; alive+stale overlays correctly; natural mux disappearance resets control state. Focused 34 passed and wider relevant 132 passed; Ruff/diff clean. Hermes pinned source confirms `--session`, same session DB route, live runtime rebuild, and stale-streak reset. Operational caveat remains explicit: behavioral fake + real HTTP is not production Hermes recovery proof; validate in a disposable live session after merge before claiming deployment success.

PR #1343, opened 2026-09-11
Fix active Hermes TUI delivery verification

### LINKED ISSUE #1340: TUI delivery verifier misclassifies active Hermes turn as pasted-but-unsent

ywatanabe1989:
## Summary

- classify a delivered token by whether it is in the bottom-most live composer or an earlier transcript row
- preserve placeholder/short-prompt fallback behavior and never treat a busy marker as stronger than payload visibly remaining in the live composer
- recognize the observed `Ctrl+C to interrupt` active-generation cue
- add an end-to-end Hermes Ink regression reproducing the compute-03 pane shape from issue #1340

## Observed basis

A read-only capture of `tui-scitex-hub` showed the submitted `[sac-deliver:...]` text retained under an earlier `❯`, active reasoning output, and a bottom-most `❯ Ctrl+C to interrupt…`. The old generic non-empty-prompt check therefore returned pasted-but-unsent even though the turn was running.

## Verification

- `189 passed in 7.20s` across `_delivery`, compose-verifier regressions, startup-prompt regression, and liveness probe tests
- Ruff check passed
- Ruff format check passed
- pre-commit hooks passed

Closes #1340


ywatanabe1989:

PR #1350, opened 2026-09-11
fix(hermes): outlive local inference deadline

### LINKED ISSUE #1348: fix(hermes): long local inference must not be abandoned before the configured upstream deadline
## Observed fleet failure (2026-09-11)

The compute-03 Qwen Hermes fleet repeatedly showed `Provider has been unresponsive ... for 5 consecutive stale attempts`, then returned to `ready`. At the same time the scitex-genai gateway remained saturated with abandoned work.

Measured at one point:

- TUI state: only 3 agents still busy, 8 ready
- client TCP: 3 ESTABLISHED and 10 CLOSE_WAIT connections to port 18772
- gateway accounting: in_flight=8, queued=38, capacity=8
- gateway logs: legitimate 2.7 MB requests completed in 1117–1129 seconds
- configured gateway upstream deadline: 1800 seconds

This demonstrates a timeout-contract mismatch: Hermes/TUI declares the provider stale and disconnects before legitimate long-context local inference completes. That wastes H100 computation, produces retries/duplicates, and strands fleet capacity.

## Required behavior

SAC must make the effective end-to-end provider/stream-idle/stale policy explicit in the built spec/runtime state. For a local Qwen engine, client abandonment must be later than the gateway/upstream deadline (or disabled while transport liveness continues). The generated Hermes configuration and image must honor it.

Add an integration test with a slow-but-live provider whose response arrives after the old stale window but before the configured deadline. Assert one upstream request, no retry/duplicate, no early ready transition, and successful completion. Keep provider failure recovery bounded for genuinely dead endpoints.

ywatanabe1989:
Closes #1348

## Observed contract

The local Qwen gateway may legitimately take 1117–1129 seconds before completing a long-context request, while its upstream deadline is 1800 seconds. Hermes 0.21.1 defaults local stale detection to 900 seconds, so the client can abandon healthy work first.

## Change

- add neutral per-engine `timeouts.upstream_deadline_seconds` and `client_abandonment_seconds`
- reject incomplete, non-positive, or non-strictly-ordered contracts
- compile client abandonment through Hermes supported per-model `timeout_seconds` and `stale_timeout_seconds` seams
- preserve the resolved values in AgentConfig/birth state and show them in `sac agents explain`
- set the tracked Qwen policy to 1800s upstream / 1860s client
- document the effective behavior

## Evidence

- 112 focused SAC tests passed
- pinned Hermes 0.21.1 integration: real slow SSE HTTP provider returns after the old deadline; one request, no early return, success; dead accepted connection exhausts three bounded attempts
- `scitex-dev ecosystem audit-all ... --new-only --severity error`: 0 net-new findings

No image build, deployment, or agent restart was performed.


ywatanabe1989:
Local verification on exact head `f45d9801ec6d9277c39d95065be7f41a2022daa5`:

- SAC focused suite: `112 passed`
- Actual pinned Hermes 0.21.1 integration with real HTTP/SSE provider: `2 passed in 9.38s`
  - live response at 1.25s, after modeled 1s upstream window and before 2s client bound: exactly one upstream POST, call remained in-flight at 1.05s, successful `done` response
  - accepted-but-silent provider: Hermes exhausted exactly three bounded 2s attempts and returned an error; it did not hang indefinitely
- scitex-dev new-only audit at error severity: `0 net-new finding(s)`

The integration test uses the official pinned Hermes package and real loopback HTTP sockets; it does not mock Hermes or its client. Normal SAC-only environments skip it because Hermes is intentionally an image-local staged dependency. No image/deployment/restart was performed.

ywatanabe1989:
Independent-review HOLD corrected at exact head `5c2bd97f2af8765998c5bb3d3b6b099e1e816732`.

The gap was real: fleet `engines:` entries were parsed without the validator used for spec-local entries. This head extracts one shared `validate_engine_entry()` and applies it at both boundaries before parsing.

New regression evidence:
- invalid fleet timeout mapping, incomplete pair, non-positive value, misordered deadlines, and unknown timeout field all populate `FleetLibrary.errors` and block a dependent agent through `validate_raw`;
- a valid fleet timeout reaches loaded `AgentConfig` (`1800`/`1860`) and compiled Hermes (`timeout_seconds == stale_timeout_seconds == 1860`);
- spec-local entry still wins a valid fleet-key collision.

Verification on this exact tree:
- `uvx ruff check ...`: passed
- focused engine/Hermes unit suite: `122 passed in 0.65s`
- fleet boundary module after adding the invalid-mapping case: `19 passed in 0.12s`
- canonical config → real Hermes profile integration: `2 passed in 0.14s`
- actual pinned Hermes slow-live/dead-provider integration: `2 passed in 9.49s` (one POST/no duplicate/no early completion/success for slow-live; dead provider remains bounded)
- scitex-dev audit gate: `2 passed in 56.97s`
- pre-commit: passed
- `git diff --check`: passed

No image build, deployment, or restart was performed.

PR #1364, opened 2026-09-11
fix(mcp-channel): dedupe re-served msg_id before dispatch (stuck CI re-fire)

ywatanabe1989:
## What

The sac MCP channel consumer re-wakes the agent's own /v1/turn for a **replayed** durable-log row (same msg_id), which is the mechanism of the stuck CI re-fire: one dead run (be31e307 / msg_id ad7db0e5, row id=9309, delivered_at set) was re-delivered ~100+ times into the same session.

## Root cause (source-verified)

- On reconnect, `_listen/_node_channel_stream.py:137` calls `list_since_id`, which replays `id > cursor` **regardless of `delivered_at`**. A consumer whose in-memory SSE cursor (`_mcp/_channel_sse.py:167`, resets to None on process restart; advanced at :224-225 only after a clean dispatch) is stuck below an already-delivered row re-serves that row on every reconnect.
- `_push_channel_event` (`_mcp/channel.py`) then re-wakes /v1/turn via `_wake_turn` -> `_channel_wake.py:116` `_wake_text` (this is the <channel msg_id=...> tag; it is SAC-rendered, not the Hermes runtime).
- The store contract **mandates** the consumer de-duplicate on the envelope msg_id: `state_db_channel_store.py:347-352` — 'The SSE cursor is NOT a dedupe key and never was — it is a resume position.' The consumer never did: `_recent` (200-cap ring, channel.py:49-50) was append-only and only consulted by a2a_reply/a2a_ack, never checked before dispatch.

## Fix

One seam, `_push_channel_event` in `_mcp/channel.py`: before wake/notification dispatch, if the event's msg_id is already in `_recent`, return (pure replay — the row is already delivered and the cursor already advanced, so dropping it loses nothing; first deliveries still pass). No new state; reuses the existing ring.

## Evidence

- New test `test_push_same_msg_id_replayed_drives_turn_once`: **fails without the fix, passes with it** (re-delivered same-msg_id drives exactly one turn).
- 465 passed / 30 skipped across tests/.../_mcp/ (the package the change lives in); audit 0 unmasked errors (scitex-dev 0.59.0).

## Note on deploy

This PR fixes the code. The *live* loop keeps firing until the running sac mcp channel process is restarted with the merged code (it runs the pre-fix version). Restart is a production step — I am running inside that process, so it must be an operator/authorized action, not a self-restart.


ywatanabe1989:
Closing as DEFECTIVE (scitex-agent-container, self-correction). This PR put the msg_id dedup inside _push_channel_event, which the inbox-bridge 429-busy-retry (_hermes_inbox_bridge.on_event) re-calls for the same event. The dedup therefore suppresses the legitimate busy-retry and the SSE cursor advances/acks the row as delivered WITHOUT it being delivered -> data loss for any live message that arrives while the gateway is busy (max_concurrent_runs=1). The root-cause analysis (channel-consumer SSE replay) was also mis-targeted: the loop driver is the HERMES inbox bridge's 429-retry, confirmed by the gateway log (1851x429, 373x202, 243 distinct idempotency keys for msg_id ad7db0e5). The correct seam is the inbox bridge's delivered-set (keyed on msg_id, recorded only after a successful dispatch, checked before re-dispatching a re-served row) — NOT _push_channel_event. Rebuilding with the correct fix before re-opening. Immediate operator relief for the stuck replay: restart the running inbox-bridge process (its cursor resets past the dead row) — this is independent of any code change.

PR #1434, opened 2026-09-14
fix(hermes): make idle admission event-driven

ywatanabe1989:
## Observed behavior

Read-only inspection on compute-03 found all six live Hermes sessions with native heartbeat status paused. Their last_fired_at values predated the current owner generations. Recent expensive turns in Hermes session state were instead preceded by durable Cards notifyd or direct user/agent envelopes. Therefore the recent turns were valid event wakes, not the configured 600-second idle kick.

The same inspection found two detached scitex-hub stale-recovery observers: an older dotfiles-spec process and the current authority-snapshot process, both reparented to PID 1. The old PID-file plus exact-config-path ownership rule could not see both incarnations.

## Change

- Clear persisted Hermes heartbeat state through session.control heartbeat.clear at exact owned-session attachment, before publishing attached.
- Keep the TUI alive and retry on transient control failure.
- Preserve event-driven Cards, CCT, SAC, A2A, and human wakeup.
- Give recovery observers explicit agent names and reconcile exact-module same-agent observers across spec incarnations before spawn.
- Preserve PID-reuse safety with PID plus process creation time; never match another or substring agent name.
- Supersede ADR-0028 with the evidence-grounded event-driven ADR-0030.

## Verification

- 59 focused tests passed.
- Runtime suite: 2443 passed, 24 skipped; the same three environment failures reproduce on untouched origin/develop (missing sac executable on PATH and local scitex-dev 0.59 lacking StoreProvisionError).
- Ruff passed on all touched Python and test files.
- git diff --check passed.
- scitex-dev ecosystem audit-all: 0 unmasked errors.

No live agents were restarted and no SGLang setting was changed.


ywatanabe1989:
CI follow-up (2026-09-14): the first Python 3.11 run failed only the repository audit. Its authoritative log reported PA-306 at the two new tests because they used the monkeypatch fixture, and PA-307 because each had multiple assert statements; no runtime test failed. Rebased onto develop at 7417b172 (PR #1432), replaced monkeypatching with explicit config-loader dependency injection and a pure recovery_monitor_argv test, and consolidated each verdict into one assert. Reverification: 59 focused tests passed; runtime suite 2443 passed / 24 skipped with only the same three local environment failures documented in the PR; current scitex-dev PA-306/PA-307 audit reports no violations. Updated head: f08b4cd007d42a785a2adcc7bd0e81561984439a.

PR #1444, opened 2026-09-14
fix(tui): handle Claude 2.1.197 default-No trust prompt

ywatanabe1989:
## Summary
- detect the unnumbered, default-No workspace-trust picker observed in Claude Code 2.1.197 and current 2.1.270
- move from `No, exit` to `Yes, I trust this folder` with `Down`, then confirm with `Enter`
- ensure trust-modal detection outranks generic compose-pending detection while preserving the older numbered layout (`1`, then `Enter`)

## Live verification
On scitex-compute-03, the exact v2.1.197 modal was reproduced with `No, exit` selected by default. Applying the new detector and `Down`/`Enter` response dismissed it successfully. A fresh Hub start then completed under Claude `--model opus[1m]`. The same unnumbered trust layout is covered for current v2.1.270.

The separate behavior where a second startup prompt can remain queued while the first turn is busy is intentionally out of scope.

## Tests
- `PYTHONPATH=src /home/ywatanabe/proj/scitex-agent-container/.venv/bin/pytest -q tests/scitex_agent_container/runtimes/test_prompts.py tests/scitex_agent_container/runtimes/test__tui_drain.py` (70 passed)
- `ruff check` on changed files
- `git diff --check`


ywatanabe1989:

PR #1464, opened 2026-09-15
fix(config): map workdir hook through bind

ywatanabe1989:
## Summary

- treat `spec.workdir` as the in-container path it declares
- translate the workdir through the most-specific explicit writable bind before creating a host-side `.claude` directory
- skip host mutation when the workdir is unmapped or read-only
- shell-quote the mapped host path

## Root cause

The config loader unconditionally generated `mkdir -p <spec.workdir>/.claude` as a host pre-start hook. Capsule specs commonly declare `workdir: /work` and bind a real host directory to `/work`, so SAC tried to create `/work/.claude` on the host and emitted `Permission denied` even though the container launch itself was valid.

For the measured Hermes+Codex canary, the hook now resolves from `/work` to `/scratch/ywatanabe/canaries/hermes-codex-channel-canary/workdir/.claude` through the declared `:rw` bind.

## Verification

- config, loader, and integration suite: 157 passed
- exact canary-spec load produced only `mkdir -p /scratch/ywatanabe/canaries/hermes-codex-channel-canary/workdir/.claude`
- regressions cover exact mapping, nested/most-specific mapping, unmapped paths, read-only binds, and shell quoting
- Ruff and `git diff --check`: passed
- `scitex-dev ecosystem audit-all`: 0 unmasked errors


ywatanabe1989:
## Coordinator review fix

New exact head: `14e5bf80995a8b9440505c9ef9397ebc169bc0ea`

The original mapping correctly used explicit writable binds and shell quoting, but accepted parent segments. A spec workdir such as `/work/../outside` could remain relative to the `/work` destination and map the host mkdir outside the declared bind source.

This head refuses `..` segments in both workdir and bind destination before any host hook is generated.

Verification:

- 93 latest-slice tests passed with 8 xdist workers
- two adversarial parent-segment tests
- Ruff and diff checks passed

All CI/review evidence for `7b8b2c37...` is invalidated. Await current-head CI before merge.

— Agent: scitex-hub-review · Dispatch: coordinator-review-1464

PR #1469, opened 2026-09-15
fix: scope Hermes sessions by engine identity

ywatanabe1989:
## Problem
Hermes sessions were keyed only by agent name. Switching an agent from Qwen to Codex resumed the old Qwen-bound session, so the process argv named the new engine while the live TUI and requests still used the old one.

## Change
Use `sac:<agent>:<engine>` consistently for Hermes continuation and the gateway session header. This preserves prior transcripts while preventing cross-engine resume.

## Verification
- 38 Hermes TUI/config tests passed
- 83 related launch/RPC/owner tests passed


ywatanabe1989:
Closing as semantically integrated by newer develop work. Current `config/_hermes_session.py` defines `hermes_session_key(agent_name, engine_key) -> sac:<agent>:<engine>`, and `_harness_callables.py` uses it for automatic continuation. The cross-engine resume bug this PR fixes is already unrepresentable; this dirty stacked branch is duplicate inventory.

PR #1472, opened 2026-09-16
fix: never resume a Hermes session from another engine

ywatanabe1989:
## Root cause
Current `develop` launches the selected model/provider but resumes `sac:<agent>`, an engine-neutral historical session. For Hub this reopened a Qwen session even though its spec and argv both selected GPT-Sol.

## Fix
- make Hermes continuation identity `sac:<agent>:<engine>`
- refuse an absent engine key instead of falling back to the model label
- add regression coverage for the exact session key and fail-closed behavior

## Evidence
The observed broken live argv combined `--model gpt-5.6-sol` with `--continue sac:scitex-hub`, while the TUI restored `qwen38-27b`.

## Tests
34 focused harness/config tests passed.


ywatanabe1989:
Closing as already integrated: `git cherry origin/develop origin/pr/1472` marks exact head commit `8fcaf26b25b3784ca1507af726ba51a8d35b35f6` with `-`, proving a patch-equivalent change is present in current develop. Keeping this open would duplicate engine-scoped Hermes resume semantics.

PR #1474, opened 2026-09-16
feat: gate agent tasks with neutral worktree policy

ywatanabe1989:
## Summary
- treat the operator-owned `~/.dotfiles/src/.bin/scitex-worktree-policy` CLI plus manifest/hash as the sole policy authority; SAC only implements lifecycle mechanics
- deterministically plan, provision, and resume an agent-owned linked worktree per agent identity, replace the in-memory runtime workdir, and authorize both provisioning and final edit context through that CLI
- preserve unfinished agent work; refuse dirty authority checkouts, unowned dirty linked worktrees, orphan branch/owner records, changed Git identity, stale projections, malformed responses, and hash drift
- persist ownership with spec/session/resume/incarnation identity and record `policy_sha256` / `projection_sha256` on incarnation state plus compiled launch artifacts
- make dry-run and `agents explain` show authored/resolved paths, action, branch, and hashes without provisioning
- remove the independent worktree-policy skill/prompt copy; launch accepts host AGENTS/CLAUDE/HERMES projections only after executable `check-projections` hash validation

## Validation
- focused policy/birth/start/explain suite: **41 passed, 16 skipped**
- incarnation returned-row contract: updated intentionally for the two persisted hash fields (PostgreSQL test skipped locally)
- `tests/develop/test_audit.py::test_audit_all_clean`: **1 passed** (74 s)
- targeted `ruff check`: passed
- `git diff --check`: passed
- real authoritative CLI dry-plan: returned `planned-linked-worktree`, policy SHA `53964ab...f0797`, projection SHA `a51234fa...07c3`, and confirmed the planned path was not created
- executable fixture drives real subprocess and Git worktree operations; it models the CLI protocol, not policy rules

## Compatibility / migration findings
- current Hub/application specs that author a main `develop` checkout no longer launch there: a clean authority checkout automatically resolves to its repository-local owned worktree
- current whole-home container binds cover repository-local `.worktrees` paths. Narrow/container-only mappings must be adapted; `agents explain` visibly reports when the resolved path is not bind-backed
- a dirty main checkout is deliberately preserved and refused. Operators must commit/move its existing work before retry; SAC does not copy, reset, delete, or silently claim it
- same-owner dirty linked worktrees resume in place, preserving unfinished agent changes
- remote hosts must expose the authoritative CLI/config/projections at the neutral host path; containers do not need dotfiles or a Claude-specific hook
- reviewed PR #1473 conceptually for lifecycle/state ordering; this branch remains based only on merged `origin/develop`

[...]

PR #1475, opened 2026-09-16
fix: verify prompt projection provenance

ywatanabe1989:
## Outcome

Prompt assembly now derives from explicit recipe state and fails closed when the source, runtime projection, or harness consumer disagrees.

- removes the implicit Claude/scitex-todo startup prompt; explicit `startup_prompts: []` is a real no-op
- makes declared `to_home_layers` control the general materialization walk
- writes and verifies `$HOME/.sac/prompt-projections.json` with SHA-256 hashes for startup turns, declared instruction/skill/command sources, runtime projections, and harness consumers
- requires Hermes recipes to provide neutral root `AGENTS.md`; SAC verifies every possible home backing and embeds those exact bytes through Hermes `agent.system_prompt`
- refuses legacy `CLAUDE.md` translation and divergent/mutated projections with an actionable migration error
- updates runtime-facing docs and skills to describe executable behavior rather than Claude-specific fallback behavior

This deliberately does not migrate fleet specs or deploy agents. Specs selecting Hermes need an explicit `AGENTS.md` in a declared `to_home_layers` source before this is deployed.

## Validation

- `1550 passed, 9 skipped` across the complete config suite plus relevant to_home / projection / Hermes profile / Hermes TUI suites
- Ruff format and lint pass for all changed Python files
- `git diff --check` passes

## Coordination

Separate from #1474: this PR owns prompt source/projection/consumer integrity. #1474 may record the resulting manifest digest/path in lifecycle identity later, but no lifecycle/birth-certificate files overlap here.


ywatanabe1989:

PR #1494, opened 2026-09-17
feat: native mid-turn steer for managed Codex TUI

ywatanabe1989:
## Summary
- own one authenticated loopback Codex app-server per managed Codex TUI
- attach the visible TUI to that server with `--remote`
- route channel arrivals through `turn/steer` while active and `turn/start` while idle
- fail loud with no terminal-key or end-of-turn FIFO fallback

## Contract
The app-server RPC ACK is the transport admission point. The visible TUI is a client of the same server, while the agent semantic read/reply remains a later receipt.

## Evidence
- installed Codex 0.154.0 WebSocket initialize verified with capability-token auth
- 130 focused launcher/native/bridge tests pass
- ruff and diff checks pass

## Dependency
Stacked on #1493, which prevents false channel ACK when legacy TUI submission is dropped. Retarget to develop after #1493 merges.


ywatanabe1989:

PR #1496, opened 2026-09-17
fix(hermes): preserve engine-scoped session identity

ywatanabe1989:
## Summary
- preserve compute-03 local commit `bbfb37ee` through merge ancestry instead of squashing/recreating it
- scope automatic Hermes continuation by agent + selected engine
- refuse unresolved engine model/key instead of opening Hermes Setup Required
- update runtime fixtures and explicit engine-switch regression

## History contract
The branch merges current `origin/develop` into the compute-03 lineage. `bbfb37ee` remains an unchanged ancestor. Merge commit: `bf985d29`; focused follow-up: `aebdc89a`.

## Validation
- 69 focused config/runtime/provider/audit tests passed
- focused Ruff passed
- `git diff --check` passed
- compute-03 host sync canary: DeepSeek real key HTTP 200/model match; invalid control 401

## Rollout gate
Do not start Scholar until this PR is green and merged with a history-preserving merge commit, compute-03 resolves canonical develop at ahead=0/behind=0, and live module provenance matches.


ywatanabe1989:
Independent review verdict: DO NOT MERGE at aebdc89a, despite green CI.

Ancestry and merge strategy PASS: bbfb37ee is preserved unchanged; after fixes use **Create a merge commit only**. Never squash/rebase/cherry-pick.

Blocking defects:
1. `_hermes_profile._launch_plan` still sets agent-only `sac:{config.name}`. Native Hermes uses `sac:<agent>:<engine>`, but the OpenCode `x-opencode-session` header aliases different engine conversations to the same agent-only identity.
2. Profile and stale-recovery paths accept `config.engine_key or config.model`, while TUI now requires an explicit engine key. All launch/profile/recovery paths must refuse missing/unresolved engine keys consistently.
3. Tests mask this: OpenCode integration still expects agent-only identity, and a Hermes test checks obsolete substring `--continue sac:scholar`, which passes accidentally for longer engine-scoped values.

Required before re-review: align profile/header/native continuation identity; remove model-as-engine fallback everywhere; test exact argv/header values across engine switches, non-Hermes roundtrip, fresh sessions, and missing engine keys. Keep the PR lineage intact by adding child commits and merging new develop into the branch if needed—no rebase.

PR #1504, opened 2026-09-17
fix(ssh): ignore inherited forwarding config

ywatanabe1989:
## Summary
- add `-o ClearAllForwardings=yes` to the shared SAC SSH argv builder
- prevent unrelated user `LocalForward` bind collisions from consuming fleet-probe timeouts
- cover both preamble and bare peer command paths

## Root cause
`ssh scitex-nas-03` inherited several user-configured local forwards. Existing listeners held ports such as 31291, so SSH emitted bind failures before the SAC command and could consume the bounded fleet timeout. These control-plane invocations need no port forwarding; inheriting it is unrelated state.

Callers may still explicitly override the default through later `extra_opts` when a command intentionally needs forwarding.

## Verification
- RED: both preamble and bare-peer tests failed because `ClearAllForwardings=yes` was absent
- GREEN: 50 host-SSH and fleet-probe tests passed
- audit gate: 1 passed
- Ruff and `git diff --check`: clean
- live read-only probes with `ClearAllForwardings=yes` reached compute-02, nas-01/02/03, and Spartan without local-forward bind noise; all five currently lack a noninteractive `sac` executable, which is a separate immutable-runtime deployment dependency

No merge, deployment, host install, or running-agent restart was performed.


ywatanabe1989:

PR #1502, opened 2026-09-17
feat(hermes): expose pending clarification RPC

ywatanabe1989:
## Summary
- observe live Hermes clarification waits through JSON-RPC (`session.active_list` + `session.activate`), without reading tmux
- expose typed request/question/choice state through SAC runtime control/status as `waiting_for_choice`
- answer one exact request through native per-question `clarify.respond`, with request-id/qid validation and post-response closure proof
- refuse stale request IDs, malformed payloads, partial answer sets, and wrong answer shapes

## Live proof
On the existing `scitex-hub-gui` session, without any pre-start configuration or mutation:
- `session.active_list` returned `status=waiting`
- `session.activate(omit_messages=true)` returned `pending_clarify`
- payload included request_id `376702d4`, two qids, question text, choices, and multi_select flags
- the current request was deliberately not answered; it remains a read-only canary

## Verification
- RED first: five missing-API tests and one missing-control-state test
- focused Hermes RPC/runtime/status suite: 68 passed
- Ruff and `git diff --check`: passed

## Scope
This PR provides the programmatic detection and native response primitives plus status projection. A notification policy consumer can deduplicate and route unresolved requests to scitex-notification without any TUI scraping. It does not auto-answer any live request and does not restart agents.

PR #1505, opened 2026-09-17
feat(gui): SAC agent activity timeline + safe control surfaces

ywatanabe1989:
## What this is

The Agents web console's activity surface, built as a **projection** of signals the SAC control plane already publishes. The GUI owns no store, derives no lifecycle state, reads no runtime directory and never touches tmux — SAC stays the single source of truth.

Card: `sac-agent-activity-timeline-dashboard-20260917`

## Contents of this PR — corrected

This branch is stacked on the baseline fleet-read fix, so its history is **four commits**:

| SHA | commit |
|---|---|
| `4d1e8fbf` | fix(gui): stop losing the whole fleet to one slow control-plane read — **belongs to PR #1506** (baseline card) |
| `6b955b89` | project a bounded activity timeline from published SAC signals |
| `56e87c07` | render the timeline with a bounded, deduped poll |
| `01e2c571` | message/steer and UI-control surfaces through existing SAC APIs |
| `57d5934d` | fix(gui): clear the audit gates on the timeline branch |

**Correction, twice-tested:** an earlier revision of this body said this PR "deliberately excludes" `4d1e8fbf`. That was wrong — the baseline fix is in this branch's history because the timeline work was branched from it. I then recommended rebasing this branch onto `develop`, and **that recommendation was also wrong**: dropping `4d1e8fbf` breaks the suite, because this branch's `test__remote.py` asserts that commit's 60s ceiling (`test_client_timeout_outlasts_the_measured_control_plane_latency` fails; the ceiling reverts to 8.0s). Verified by actually running it.

**Merge guidance (tested, not assumed):** this PR and #1506 **both** carry the same test relocation into `test__remote.py`, so **either merge order conflicts in that one file** — the conflict is mechanical (comment wording and one test name), not semantic. Resolving `test__remote.py` to this branch's version produces a merged tree that **passes: 145 tests, 0 audit findings, delivery-claims clean, 60s ceiling intact**. Recommended: merge #1506 first, then resolve that single file. Nothing has been merged or force-pushed.

## Timeline projection (`_timeline.py`)

Flattens the same `/agents/<name>/status` activity blocks the detail view already reads.

Four facts kept distinct, because collapsing them hides real breakage:

| state | meaning |
|---|---|
| `observed` | the runner published this signal |
| `stale` | it published before and has since gone quiet — real history, marked old |
| `unreachable` | we could **not ask**. An outage, not "no news" |
| `unknown` | asked successfully, nothing published |

Dedup keys on `(agent, kind, value, state, at)` — **value included**, so a `busy` → `idle` change survives as its own row while a re-polled unchanged observation collapses. Window capped; filters applied server-side.

## Rendered surface (`/timeline/`)

Server-rendered so it is complete on first paint and works with JS off. Filters are a real GET form applied **server-side**, and the JS poll forwards the same query string — the two paths cannot diverge, and an unmatched filter returns **empty, never everything**.

The poll is deliberately conservative against a control plane this fleet measures at 5–60s per read:

- **one request in flight at a time** — the next tick is scheduled only after the previous settles, so the interval is a *floor*, not a rate
- pauses while the tab is hidden
- on failure it **stops** and offers Retry rather than hammering a listener that is already unhappy

## Control surfaces

Dials the agent's **own published `turn_url`** — the same endpoint the fleet's A2A send dials:

```
turn / steer   POST {turn_url}          {"text": ...}
UI control     POST {turn_url}/control  {"key": Enter|Escape|C-c}
```

**Idempotency is the point of this surface.** A duplicate submit would inject the operator's text into a *live* agent's turn twice. Every mutation carries a dispatch id; a repeat is reported as `replay` and not re-sent. A *failed* delivery deliberately does not poison its id, so a corrected retry lands.

Control keys are exactly the three the bridge accepts. An invented fourth (`C-d`) is refused **locally**, without contacting an agent, rather than rendering a button that can never work.

Authorization is identical to `lifecycle_action`. Refusals audit as `message_denied`; deliveries as `message_action` with the dispatch id but **not** the message text — the audit trail must not become a transcript.

## Verification

**Tests:** GUI suite **145 passed**, ruff clean, RED-first throughout.

**Audit gates (the reason for `57d5934d`):** CI was red on both Python legs with PA-307 (21 findings), PS-204 and an unnamed delivery claim. All three cleared at the root — no rule suppressed, no assertion dropped, no allowlist widened. `audit-all` now reports **0** findings; `test_audit.py` + delivery-claims pass.

**Live canary** on a *disposable* agent (`/home/agent/sac-gui-evidence-20260917/control/control-canary.json`) — all 7 checks pass, including the message reaching the agent **exactly once** and the real fleet untouched (26/26 before and after).

**Viewport evidence** at 1440×900 and 390×844: 118 rows, **118 distinct keys** (no duplication on render), `states {unknown 112, unreachable 6}`, zero horizontal overflow at both widths.

**CI at `57d5934d`:** tests (pytest-matrix py3.11 + py3.13 + verdict), lint, docs, import-smoke, no-hosted-runners, cla — all success.

## Known gaps — stated, not hidden

1. **`stale` has never fired from real data.** `/status` publishes only the current activity block with no `observed_at`, so every entry stamps at read time. The state is built and tested; the data cannot produce it. This is why the surface is honestly a *bounded live projection*, not "near-real-time with history". Flagged upstream — it needs one typed event contract.
2. **Pending-choice response is UNAVAILABLE — `reason=upstream review failed`.** *(Corrected.)* An earlier revision of this body said no such API exists. That was imprecise: it **does** exist, in PR #1502 (head `b121ec63`), and is **independently rejected**. `observe_pending_clarification()` documents itself as reading "without attaching to or mutating the TUI" while its body calls RPC `session.activate` — the observation mutates, which is disqualifying for a GUI control. Response atomicity, concurrency, privacy and idempotency are all still open. Not wired, deliberately. Evidence: `/home/agent/sac-gui-evidence-20260917/pending-choice/availability-probe.txt`
3. **Control surfaces have no screenshot** — verified by tests and the live canary, not visually.

## Not in this PR

No merge. No deploy. No lifecycle mutation was performed against any real agent.

Upstream dependency (blocks items 1 and 2): a typed activity/message event contract carrying `event_id`, `agent`, `kind`, `observed_at`/`opened_at`+`closed_at`, `source`/`target`, and a secret-free summary. One contract unblocks this card and the bubble-visualization card.


ywatanabe1989:

PR #1506, opened 2026-09-17
fix(gui): stop losing the whole fleet to one slow control-plane read

ywatanabe1989:
## What this is

One bounded GUI-layer fix: the Agents console was **rendering zero agents and a false alarm** against a healthy control plane.

Card: `gui-baseline-scitex-agent-container-20260917` — kept **separate** from the activity-timeline PR per the operator's instruction.

## The defect

`GET /` rendered **0 rows** plus the banner *"could not reach the SAC listener … timed out"* — at **both** 1440×900 and 390×844 — against a listener that was up, authenticated, and answering `200` with 19 own-scope agents.

The whole fleet table, the harness/engine columns, and the entire lifecycle surface vanished with it.

## Root cause (measured, not inferred)

`RemoteFleet` capped every exchange at **8.0s**, while the real control plane on scitex-compute-03 took longer:

| read | measured |
|---|---|
| `GET /agents` warm | 5.11s / 5.23s / 6.91s |
| `GET /agents` cold | 18.99s |
| one excursion | 117.8s |
| `GET /agents/<n>/status` | 2.1s / 8.5s / 14.1s / 18.5s |

The latency sat **on** the 8.0s wall, so the same healthy node rendered agents or a false alarm depending on the moment. Not a token or connectivity problem — verified separately.

## The fix

- `DEFAULT_TIMEOUT_SECONDS = 60.0` — clears the **cold** latency of a real fleet (a cold read is the normal case after a restart), so "slow" stops meaning "absent".
- `PER_AGENT_TIMEOUT_SECONDS = 12.0` — the fleet is read together, so no single agent may spend the page's whole budget. One wedged agent degrades to its own explicit row.
- `socket.timeout` is classified as `FleetUnavailableError`, so an over-budget read lands in the **same explicit per-row state** as an unreachable one instead of escaping raw.

## Verification

**3 new tests** (`test__fleet_read_budget.py`), RED before the fix — 2 of 3 assertions failed. Full suite **94 passed** at that commit; ruff clean.

Real loopback control planes, no mocks/monkeypatch, per the suite's own honesty rule.

**Live:** 19 rows at both viewports, no banner, no horizontal overflow.

Evidence: `/home/agent/sac-gui-evidence-20260917/` — `REPORT.md`, before/after PNGs and HTML, JSON measurements.

## Honest limits

- **This is containment, not a cure.** After the fix a live fleet GET still hit exactly **60.06s** and `/agents` hit 42.4s, so the ceiling is reached in normal operation. The durable fix is a faster control plane — caching/parallelizing `/agents` — which is the control-plane owner's slice and is flagged on `sac-listener-agents-endpoint-latency-20260917`. I did not touch it.
- **`harness`/`engine` render `unknown`** for every agent because the control plane publishes no such fields. Not a GUI defect; reported as an upstream data-availability gap.

## Not in this PR

No merge. No deploy. No lifecycle mutation — every measurement was a GET; all 19 real agents untouched.


ywatanabe1989:
CI VERIFIED AT THIS EXACT HEAD - 6/6 WORKFLOWS GREEN. Requesting independent review.

Head 668771b6 (local == remote == PR head: 668771b655d77801cc0705032290a4f115376d52). mergeable: MERGEABLE. 0 reviews so far.

  tests (pytest-matrix) ... success  [35214059668]  py3.11 + py3.13 + verdict all success
  lint ................... success  [35214059778]
  docs ................... success  [35214059681]
  import-smoke ........... success  [35214059650]
  no-hosted-runners ...... success  [35214059864]
  cla .................... success  [35214059129]

This CI verdict is NOT a merge signal and I am not merging it.

TWO THINGS FOR THE REVIEWER, flagged rather than left to be found:
1. The PR body describes the fleet-read fix; the SECOND commit (668771b6) is a test relocation + conformance change with no behaviour change. PS-204 orphan test file: tests/.../test__fleet_read_budget.py mirrored no src module, so its tests moved into test__remote.py (which mirrors _remote.py) and the orphan was deleted. While moving, they were made to conform to the quality rules rather than carry findings across (one assertion per test, fixture yields, AAA markers).
2. A PRE-EXISTING FLAKE is visible on this branch's first CI run and is NOT fixed here: tests/scitex_agent_container/runtimes/test__tui_turn_bridge.py::test_post_inject_failure_concludes_ordinary_exchange_with_502 -> KeyError: 'error' at line 735. `_wait_exchange` polls 100x with a 1ms sleep (~100ms budget) and returns as soon as status_code != 202; under CI xdist contention the exchange is still RETRYABLE when that budget expires (the captured log says "turn exchange remains retryable", hint "retry that notification"), so the receipt carries no `error` key and indexing it raises. NOT this branch's regression: the diff does not touch the turn bridge or its test, and _wait_exchange is byte-identical on origin/develop; the file passes 12/12 isolated and 3/3 under -n 8 locally. Fixing it means widening a timing budget in turn-bridge tests, which was excluded from this branch's scope. Flagging it so it is owned somewhere rather than silently recurring.

Local gates before push: audit-all 0 PA-307/PS-204 (was 1); test_audit.py + delivery-claims 6 passed; GUI suite 96 passed; ruff clean.
Evidence: /home/agent/sac-gui-evidence-20260917/

No merge, no deploy.

ywatanabe1989:
Closing as superseded by PR #1512. #1512 carries this PR’s `_remote.py` 60s/12s bounded read and relocated 174-test slice, then completes the product fix with a non-blocking identity-scoped cache, honest loading/cached states, and removal of endpoint/token/setup leakage. Merging this containment separately would add no unique behavior and would preserve the 60s first-paint defect that #1512 removes.

PR #1509, opened 2026-09-17
feat(gui): five-state fleet surface with Retry and a diagnostic ID

ywatanabe1989:
## What this is

The Agents console's **five-state fleet surface** — the states told apart instead of blended into one banner, with a Retry that works and a sanitized diagnostic ID.

Card: `agents-gui-live-inventory-mvp-20260917`

## The problem

The fleet page had **one** blended failure banner. A healthy-but-slow listener, an unconfigured deployment, an empty scope and a refusal all rendered through the same path — which sent operators to the wrong fix ("set the token" for an outage; "retry" for a deploy gap).

## The five states

Each resolves to exactly one `data-fleet-state` on the app root, so they are machine-checkable and visually distinct:

| state | meaning | the right response |
|---|---|---|
| `unavailable` | configured, listener silent | an **outage** — retry / escalate |
| `setup-required` | no listener chosen | a **deploy gap** — configure it |
| `empty` | answered, nothing in scope | ordinary, not an error |
| `denied` | refused on authorization | an **access request** |
| `ok` / `loading` | rows render; sketch covers navigation only | — |

**`denied` is deliberately NOT resolved from the fleet-read path.** "What I may read" and "what I may control" are different questions, so manufacturing `denied` from an empty read would either mask a working fleet or leak its size. The read path resolves the other four; authorization decides `denied` on its own surface, where it is a real outcome.

## Diagnostic ID

`diag-<12 hex>`, **derived** rather than quoted. A raw exception string is exactly what carries a Bearer token, a credentialed URL or a spec path — and this app already drops those on purpose in `_projection`. The id is a digest of a canonical cause *category*, so two hosts hitting one failure correlate and neither is recoverable from it.

## Evidence

Live at both viewports, `/home/agent/sac-gui-evidence-20260917/mvp/`:

| state | desktop 1440×900 | mobile 390×844 |
|---|---|---|
| `ok` | rows, no overflow | rows, no overflow |
| `empty` | renders | renders |
| `unavailable` | renders | renders |

6 PNGs, 6 HTML dumps, `mvp-report.json`. `setup-required` and `denied` are covered by **tests** — forcing them needs a controlled listener/identity, and I won't claim screenshot evidence I didn't take.

## Two real bugs found and fixed while building this

1. **`views.index` referenced `API_URL_ENV` without importing it.** `_is_configured()` raised `NameError` into the view's broad `except`, and **every** fleet state silently collapsed to `ok`. Invisible until the new state tests existed.
2. **My own first design resolved `denied` from the read path** — wrong for the security reason above. Corrected.

## Audit gates (the reason for the last commit)

This branch had never been pushed, so it had never faced the audit gate. Auditing **before** the first push found **8 findings** — two orphan test files and six test-quality findings. All cleared at the root, no rule suppressed:

- `test__fleet_read_budget.py` → merged into `test__remote.py` (the module it exercises)
- `test__fleet_states.py` → renamed `test__diagnostics.py` (the module it mirrors)
- multi-assertion tests **split**, never trimmed

A real bug surfaced while merging: this branch's bounded-read fixture was an older, broken shape (referenced non-existent attributes; called `socket.gethostname()` without importing `socket`). Pushing unexamined would have shipped a broken test file.

## Known gap — `engine/model` is UNKNOWN, deliberately

The acceptance criterion asked for `engine`/`model` in the inventory. **The control plane publishes no such fields** — `/agents` rows carry none, and across all 19 own-scope agents `/status` returns non-null counts of `liveness 0, status 5, runtime 0, harness 0, engine 0, model 0, pid 0, session_id 0`. The columns render and read `unknown` because that is honestly all the committed contract serves. Per the operator's instruction, values are left unknown rather than invented. Flagged upstream on `sac-listener-agents-endpoint-latency-20260917` — the same typed-event gap that blocks the bubble-visualization card.

## Verification

- GUI suite **112 passed**; `audit-all` **0 findings** (was 8); `test_audit.py` + delivery-claims pass; ruff clean.
- No backend, runtime or infrastructure mutation. No agent started, stopped or restarted. Every measurement was a GET; all 19 real agents untouched.

No merge. No deploy.


ywatanabe1989:

PR #1511, opened 2026-09-17
fix(tmux): bracket-paste long TUI turns atomically

ywatanabe1989:
## Summary
- replace streamed `tmux send-keys -l <payload>` with a unique named tmux buffer loaded from stdin
- deliver the buffer with one `paste-buffer -p -r` transaction, preserving embedded LF bytes and keeping prompt text out of argv/process listings
- delete the buffer on both success and paste failure; fail loud when load/paste fails
- retain the already-merged PR #1493 submission proof: Enter remains separate, idle-gated, and must produce durable advancement before the bridge acknowledges delivery
- document the safe manual long/multiline recovery path

## Root cause
`send-keys -l` is literal but still streams one input event per character. A live Codex redraw interleaved with a 2.2 KiB prompt, corrupting and truncating the composer while tmux returned success. PR #1493 already fixed the independent false-ACK half by requiring submission proof; this PR closes the remaining payload-transport corruption seam.

## Validation
- RED: 3+ KiB multiline regression failed because the old API had no bracketed-buffer path
- real tmux probe preserved a 3,427-character Unicode/multiline buffer byte-for-byte
- `330 passed` across tmux, TUI submission/composer/session, and verified-delivery suites
- `114 passed` across bracketed-paste, exact-target, submission-ACK, and full turn-bridge suites
- audit: `2 passed`
- Ruff, compileall, and `git diff --check`: passed

## Safety
No merge, deployment, restart, or live agent input is included.


ywatanabe1989:
## Coordinator review fix

New exact head: `2d6a2004ce5cad01b76c9c7c48f930a52adbd764`

The prior implementation deleted the named tmux buffer after paste success/failure, but a nonzero `load-buffer` returned before cleanup. If tmux created a partial buffer before returning failure, prompt contents could remain in tmux despite the PR's fail-clean contract.

This head best-effort deletes the named buffer on load failure before raising, and adds an adversarial call-sequence test.

Verification:

- 29 focused tmux tests passed with 8 xdist workers
- load/paste success, paste failure, load failure, exact target and real 3KiB multiline bytes covered
- Ruff and diff checks passed

Prior CI is invalidated. Await current-head checks before merge.

— Agent: scitex-hub-review · Dispatch: coordinator-review-1511

PR #1512, opened 2026-09-17
fix(gui): P0 agents shell renders before inventory; no internal endpoint/token reaches the browser

ywatanabe1989:
## P0: the Agents console blocked its whole render on the control plane, and leaked internals when that failed

Operator-reproduced on mobile: tapping **Agents** waited ~10s, then showed **0 visible / 0 alive** plus raw `http://host.docker.internal:7878` and `SCITEX_AGENT_CONTAINER_API_*` setup instructions.

### Measured before (parent commit, same host)

| | before |
|---|---|
| fleet page render | **60.06s / 60.04s / 60.05s** (operator saw ~10s — worse here) |
| `SCITEX_AGENT_CONTAINER_API_URL` in browser HTML | **1 occurrence** |
| `SCITEX_AGENT_CONTAINER_API_TOKEN` in browser HTML | **1 occurrence** |
| `"bearer token"` prose | **1 occurrence** |
| listener sub-label | `Scoped fleet — http://127.0.0.1:7878` |

**Root cause:** `views.index` ran the entire control-plane read inline, so the browser waited on SAC before it could paint anything — data the **shell** does not depend on. On failure it dumped the endpoint and deployment prose into the page.

### Measured after

| | after |
|---|---|
| shell render (5 server renders) | **0.002s max** (0.001–0.002s) |
| forbidden strings | **{}** — none |
| first-paint state | `loading` (honest, not empty) |

**Browser-verified, both viewports:**

| viewport | click-to-shell | state | leaks |
|---|---|---|---|
| mobile 390×844 | **0.385s** | `loading` | **[]** |
| desktop 1440×900 | **0.269s** | `loading` | **[]** |

First text a mobile user now sees: `Agents | Fleet | 0 VISIBLE | 0 ALIVE | Loading the agent fleet | Reading the fleet…`

### The fix

1. **Non-blocking shell.** `index` no longer reads the control plane. It renders from the identity's short-TTL last-known snapshot if one exists, else empty in a `loading` state, then kicks a read-only background refresh. Click-to-shell never waits on the listener.
2. **No internal leak.** The endpoint, env-var names, token name and setup paragraph are gone; the header sub-label no longer prints the listener either. The browser gets a user-facing state plus **Retry**; diagnostic detail stays server-side.
3. **Real inventory only.** A cold cache renders `loading`, never invented rows. A served snapshot renders as `cached` **with its age**, never as current.

### New module: `_inventory_cache.py`

Identity-scoped, bounded (64 entries), 45s TTL. Three properties enforced **by construction**, which matters because the Hub launcher may prefetch on idle/intent:

- **Read-only** — its only input is a caller-supplied read callable, so a prefetch has no path to a lifecycle verb.
- **Scope-safe** — keyed by identity, so one operator's snapshot can never be served to another.
- **At most one refresh per identity** — a burst of page loads cannot fan out into a burst of control-plane reads.

### Tests

- **14 new** in `test__inventory_cache.py`, **RED first** — 3 failed against the old blocking view.
- Four existing `test_views` tests asserted the old contract (rows on first paint) and were **retargeted to the read path**, where the scope and activity rules still hold. None deleted, none weakened.
- Full suite **110 passed**; `audit-all` **0 findings**; ruff clean.

### Boundary

My package owns the **Agents page and its data visualization**; tile navigation is the Hub's. Nothing here touches the launcher — and the Hub can prefetch `GET /` with no cooperation from me, because it is now cheap and mutation-free by design.

### Evidence

`/home/agent/sac-gui-evidence-20260917/p0/` — `p0-repro-before.json`, `p0-repro-after.json`, `p0-after-mobile.png`, `p0-after-desktop.png`, plus HTML dumps.

No merge, no deploy. No lifecycle mutation.


ywatanabe1989:
## Latest-base integration checkpoint

Current exact head: `ccfd4299803bae0826c875f96c2630fb2488ca66`

Integrated current `develop` and resolved the sole `_remote.py` conflict without reintroducing the old HTTP N+1:

- preserved develop's single enriched `GET /agents` batch projection;
- preserved this PR's immediate shell, identity-scoped bounded cache, truthful loading/cached/unavailable states, endpoint/token leak removal, and 60s background-read ceiling;
- removed the now-obsolete per-agent fan-out constant/path;
- retargeted newer runtime-identity column tests through actual `/api/fleet` → shared partial rendering, because cold first paint is intentionally loading-only.

Verification:

- inventory cache/remote/views/current batch-status slice: **65 passed**;
- audit + delivery-claim security slice: **6 passed**;
- Ruff and `git diff --check`: pass.

All old CI/review is obsolete. Current-head CI and fresh non-author security/product review are required before merge/live rollout.

ywatanabe1989:
## Fresh exact-head HIGH-risk review — FAIL / do not merge

Reviewed exact head: `ccfd4299803bae0826c875f96c2630fb2488ca66`

Exact-head CI is green/CLEAN and the initial server render is non-blocking, but product/security review found six blockers:

1. The shell starts a server-side refresh but has no client poll/request to `/api/fleet`; cold pages stay “loading” and also render “No agents visible,” never transitioning to inventory or unavailable/Retry.
2. Raw listener exception text, including internal URL/transport diagnostics, still reaches `/api/fleet`, detail HTML, and lifecycle 502 responses.
3. Cache key is identity only, not identity + authorization scope. Revoking cross-host scope can still render a cached cross-host row.
4. The 64-entry cap covers completed snapshots only; in-flight identities/threads are unbounded. A probe created 80 simultaneous threads and `_refreshing` entries. `REFRESH_STALE_SECONDS` is unused.
5. Frozen Snapshot retains mutable dict/nested aliases; mutating the fetch result changes cached readers.
6. Tests hide the product failure via permissive/loading assertions and shared async state; a mounted-links test fails in isolation on cold first paint.

Required remediation: bounded global worker/queue and stale-flight cleanup; immutable deep-copied snapshots; cache key/validation bound to current authorization scope; sanitized client error envelopes; actual client poll/state transition including unavailable+Retry; deterministic isolated RED tests. New head requires full CI and fresh non-author review before merge/live rollout.

ywatanabe1989:

